PR #33, opened 2017-07-28
Add CircleCI support

yanokwa:


yanokwa:
This build not work, but we need to add a circleci.yml file first to bootstrap CI.

PR #71, opened 2017-08-21
Overwrite source images when publishing

yanokwa:
I assumed that the existing circle.yml file overwrites the source images. It doesn't, and this command implements the behavior I wanted.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -19,7 +19,7 @@ deployment:
     branch: master
     owner: opendatakit
     commands:
-      - pngquant $HOME/docs/build/_images/*.png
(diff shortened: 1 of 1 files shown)


yanokwa:
This build will likely fail because  `select-search-start.png` is a corrupt PNG but I think we should fix that in another PR.

adammichaelwood:
> select-search-start.png is a corrupt PNG
hmm.. i wonder what caused that... 

> fix that in another PR.
yes.

> overwrite source images

Do you mean just on the circleCI server? (You're not overwrite files in the actual source, right?)

yanokwa:
Sorry I wasn't clear! It will rewrite files in the /build/_images/ folder on the CircleCI server.

PR #72, opened 2017-08-21
Glfs

### LINKED ISSUE #65: Move binary files to GLFS
We currently store images and videos directly inside the docs repo. This causes the size of the repo to grow quickly and makes for slow cloning and fetching. 

I think we should move these files to [Git LFS](https://git-lfs.github.com/) so we can store binary files at a high resolution, but be able to pull and compress them (e.g., https://github.com/opendatakit/docs/issues/64) at publication time.

adammichaelwood:
I had a lot of problems with this, as discussed in #65 
But I *think* i got everything right finally.

Please check on whether GLFS is working properly.
According to the GLFS docs, it works automatically with github -- but I couldn't find anything in the GH UI to indicate it was doing its job. (I had thought I would see a new GLFS section in settings, as some of the tutorials indicated.)

Closes #65


yanokwa:
I agree that this change will allow git LFS to track the file changes, but the files remain in git's history and so the repo will be forever large.

I did some more looking around and I think [git-lfs-migrate](https://github.com/git-lfs/git-lfs/wiki/Tutorial#migrating-existing-repository-data-to-lfs) might be what we are looking for. It is currently the preferred solution over BFG.

To do the migration, we'd do this:
```
mkdir lfs;
git clone --mirror git@github.com:opendatakit/docs.git;
java -jar git-lfs-migrate.jar -s docs.git -d lfs/docs.git -g git@github.com:opendatakit/docs.git "img/**/*" "_static/vid/**/*";
```
I've tried the above steps and I can confirm they make for a much smaller repo. Then all we should have to do are these (totally untested) steps below.
```
cd lfs/docs.git;
git fsck && git push --mirror --force git@github.com:opendatakit/docs.git; 
```
Worst case, we destroy the repo, but that's a relatively easy thing to recover from.

adammichaelwood:
Yes. I've been mulling on this since I went home today and was about to get online to say DON'T MERGE. I hadn't thought all the way through the issue when I had my nose in trying to just get it working.

yanokwa:
Let's think through it together on a call tomorrow. Two brains > one brain!

PR #74, opened 2017-08-22
glfs instructions and add formats to gitattributes

### LINKED ISSUE #65: Move binary files to GLFS

adammichaelwood:
In addition to the work by @yanokwa, this Closes #65


adammichaelwood:
Updated the instructions and organized `.gitattributes` -- added the mentioned formats, and also other common archive formats.

Was unsure of what you meant by:

> Finally, how did you verify that this filtering works?

But I guessed you meant if I checked adding new file formats to the repo and whether they were tracked/included properly. **Yes** -- I tried a PDF and a GIF file in the root dir. Everything worked fine. (I reset the commits after doing so.)

PR #115, opened 2017-09-08
Added more info in contributing file and fixed typing errors

mish24:
Added the following information:
1. about commit messages
2. Python 3 compatibility
3. branches

also fixed typing issues

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 This repo is the source for ODK documentation.
 
-## Current Status — Live! (But still new.)
+## Current Status — Live! (But still new)
 
 Our new documentation website is live at https://docs.opendatakit.org
 
@@ -14,8 +14,8 @@ This is still a new project, and most of the useful documentation is still at th
 
  - Docs are being aggregated, rewritten, edited, and expanded into this repo.
  - The first things we've done are:
-    - a getting started guide
-    - contributors guide
+    - a __Getting Started__ guide
+    - a __Contributors__ guide
  - This repo will be used as a single source for documentation for all ODK components.
     - There are good reasons for separating the docs into individual repos for each component. We may do this in the future. [Please see this ODK Forum thread to discuss whether to keep all the docs together or disaggregate them.](https://forum.opendatakit.org/t/docs-structure-discussion-one-or-many-repos/7080)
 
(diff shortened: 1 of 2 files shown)


adammichaelwood:
These are good fixes and additions. I think the additional information is very helpful.
And **the build passes** - so that's excellent.

Few things to improve:

## Python virtual-env

The problem with `pip install virtualenv` (instead of python3-venv) is if your system python is python 3, instead of python 2. 

I don't know what the Linux situation is, but Macs ship with Python 2 -- I (and others) use the python 2 venv to create python 3 environments.

I don't know if all that context is needed, but it might be helpful to be more clear: *If you get an error like the following (show code block with error) then you need to install the python 3 virtual environment package (show code block with install command.*

Additionally, `apt-get` only works on particular Linux distros. Perhaps make a note of that. Also, Is there a reason you suggested `apt-get` instal of `pip install`?

## Commits in workflow

Thank you for noticing that I forgot to tell people to commit before pushing. If we had non-coding contributors they would have been very confused.

I suggest moving the commit portion to *before* "Once your work on the issue is completed, push your branch to your GitHub repo". It comes first logically, and I think we want to encourage folks to commit as they work.

## Amending commits

I'm concerned about mentioning amending commit messages after you've pushed a commit. This requires the `-f` ("force") flag when re-pushing, which has the potential to cause other problems to an inexperienced committer. I also have a hard time thinking we would ask someone to rename their commits. 

----------

In case you didn't know: you can just add more commits to your branch, and they'll be added to this PR. I *suggest* dealing with each of the above in separate commits.

Thank you so much for your work on this.

mish24:
Hi sir

Regarding the Python 2 and Python 3 virtual environment issue, I will specifically add the error which I got while I was installing the project. Plus, as I was looking, I found a way to create virtual environment in Python 2 and setting Python 3 as the default for it as we initiate it, so I will add that. Plus I will specify that this might happen in Linux. The `python3-venv` installation almost works perfectly with `apt-get` because if we switch to `pip` user might have to specifically mention `pip3` in some cases.

And I will make the required changes in the workflow section! Thank you

mish24:
screenshot from 2017-09-09 12-12-40
@adammichaelwood  I was just trying to recreate the error I got, and if you notice the last two commands, it shows that we require python 3 as the default version, and Linux comes with Python 2. What do you think is the best way to put that only and only `Python 3` is required for installation?

adammichaelwood:
> What do you think is the best way to put that only and only Python 3 is required for installation?

I thought I addressed that already in the doc. Is it not clear?

http://docs.opendatakit.org/contributing/#python-3

You might want to add a note that people may have Python 2 by default.

The instructions right now assume that you have python 2 by default, which is why the py2 version of venv is referenced, *and* why the Python3 interpreter is specified in `venv -p python3 odkenv`

mish24:
@adammichaelwood Done!

PR #116, opened 2017-09-08
Displayed the glossary points in a proper manner

mish24:
Before the PR
screenshot from 2017-09-08 09-28-44

After the PR
screenshot from 2017-09-08 09-26-48

### DIFF
--- a/glossary.rst
+++ b/glossary.rst
@@ -1,92 +1,95 @@
+********
 Glossary
-==============
+********
 
 .. ODK-compatible tools
 
 .. glossary:: :sorted:
 
-  participant
+participant
+***********
 
-    A person being interviewed by a user of :term:`Collect`. (Also sometimes called a "subject".)
+* A person being interviewed by a user of :term:`Collect`. (Also sometimes called a "subject")
 
-  enumerator
+enumerator
+**********
 
-    A person who conducts a survey.
+* A person who conducts a survey.
 
-  Collect
+Collect
+*******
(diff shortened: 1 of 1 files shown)


adammichaelwood:
Thank you for your work.

Unfortunately, this change is not needed or helpful. The glossary uses a specific syntax, not headlines and bullets.

adammichaelwood:
Also - please build the docs locally before submitting PRs.
If you had done so, you would have seen massive numbers or warnings, as this change broke all `:term:` links.

mish24:
@adammichaelwood I got the warnings, but didn't know that is was not required. Sorry!

adammichaelwood:
It's okay. We're all learning.

Maybe you could amend the contributors guide:
**If you submit a pull request that will create build errors, you must include a note mentioning what those errors are, and why they are ok to leave in.**

Should probably just say that we do not accept PRs with build errors or warnings. 
I originally thought we'd allow some warnings (particularly missing cross references, since that might indicate future-planning), but it's actually too complicated to sort out what to accept and not, so we just don't allow any warnings or errors at all.

adammichaelwood:
You could also file an issue and then submit a PR adding info on the glossary syntax, and using the `:term:` cross reference.

http://www.sphinx-doc.org/en/stable/markup/para.html?highlight=glossary

http://www.sphinx-doc.org/en/stable/markup/inline.html#role-term

PR #122, opened 2017-09-09
Display page title on footer buttons (Issue #104)

### LINKED ISSUE #104: Put page title in/on/near next/prev buttons
The Next and Previous buttons at bottom of page would be more helpful if they told you what page was next.

mish24:
screenshot from 2017-09-09 08-05-20
Fixes #104 

@adammichaelwood  Done :)

### DIFF
--- a/_templates/footer.html
+++ b/_templates/footer.html
@@ -1,0 +1,51 @@
+<footer>
+  {% if (theme_prev_next_buttons_location == 'bottom' or theme_prev_next_buttons_location == 'both') and (next or prev) %}
+    <div class="rst-footer-buttons" role="navigation" aria-label="footer navigation">
+      {% if next %}
+        <a href="{{ next.link|e }}" class="btn btn-neutral float-right" title="{{ next.title|striptags|e }}" accesskey="n" rel="next">{{ next.title}} <span class="fa fa-arrow-circle-right"></span></a>
+      {% endif %}
+      {% if prev %}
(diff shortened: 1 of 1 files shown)


adammichaelwood:
This is GREAT!

Unfortunately, there's an unforeseen problem. I'm hoping you can fix it.

When the screen width is small (as on mobile), the buttons collide and then stack poorly.

screen shot 2017-09-09 at 7 43 08 am

I *think* the solution should be that when the screen is small enough they become stacked neatly on top of each other. They would need to be the same size then, and full width.  The Prev button should stack on top of the next button in that case. 

This requires some knowledge of responsive CSS.

If this is something you can do, please add it to the PR.

(Custom CSS is in `/_static/css/custom.css/`.
That file is kind of mess, so just append your changes to the bottom,
and comment above the code to explain what it is for.)

If this is not something you can do -- that's okay. We'll merge in this PR and file a new issue for this problem.

mish24:
@adammichaelwood I don't know much of front end, but I would surely like to do some research and work on this. If this is not too urgent to be fixed, I would like to continue working :)

adammichaelwood:
If you want to take some time to try to figure it out - by all means. This is not an urgent fix.

If you have trouble and can't get it in a day or two, we'll merge this and open the CSS thing as a new issue. We all have different skills, and what's hard for one person will be easy for someone else.

mish24:
@adammichaelwood You can merge this one. I tried to address this, but there are several new things like `@media` queries and CSS properties which I am unaware about. I will open a new issue for the making the footer responsive.

adammichaelwood:
Great. No worries, and thanks for your work.

Issue #114, opened 2017-09-05
Main Menu explanation


adammichaelwood:

lognaturel:
Specifically, clarify what View Sent Form is for - https://forum.opendatakit.org/t/deleted-submission-remains-in-view-sent-form/9669

mish24:
@lognaturel Can I take this up?

adammichaelwood:
(I bet she'll say yes.)

The idea here was just to be real clear on each one. They are mostly self explanatory, but it would be helpful to say that the user will see (for example) Sent forms, but not saved one. Or that they'll see deleted ones but not be able to edit them. Etc.

mish24:
@adammichaelwood It was clear to me when I was reading it for the first time. But I will add extra data :)

mish24:
![screenshot from 2017-09-10 06-17-01](https://user-images.githubusercontent.com/24610197/30246651-b5910284-95ef-11e7-8c70-a1eea8e749a1.png)

@adammichaelwood Please see the highlighted text. I'm a bit confused. The forms (saved and finalized both) are viewable all the time right?

adammichaelwood:
I thought it was clear too (I wrote it!)
But it isn't all in one place.

@ln wanted a short section in the collect guide that just explained each item in the main menu, and what one should expect to see when opening it. (For example, some lists show saved but not sent, some show deleted, while others don't, etc.)

mish24:
@adammichaelwood What about sent forms? They are viewable even after they're deleted right?

PR #123, opened 2017-09-10
Added extra info about lists (Issue #114)

mish24:
@adammichaelwood I have added some info which I thought was missing in the doc. Are these additions correct? Plus, is there a separate doc needed for these explanations?

### DIFF
--- a/collect-forms.rst
+++ b/collect-forms.rst
@@ -118,7 +118,7 @@ If you are connected to :ref:`an ODK Aggregate server <connecting-to-aggregate>`
 
 For local form management, use `ODK Briefcase <https://opendatakit.org/use/briefcase/>`_ to pull :formstate:`finalized` form instances to your local computer.
 
-:formstate:`Sent` forms are no longer editable, but they remain viewable, even after they are deleted. This feature is mainly for for enumerators and/or data managers to confirm which filled out forms were sent even if they were deleted.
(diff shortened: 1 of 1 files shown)


adammichaelwood:
Thanks.

ln 108 -- they are listed, but you can't view deleted forms

ln 134 -- `a directory for each instance.` (not `instances`)

ln 138 `In the case of filled out forms that are not deleted, the data can be reviewed but not edited.`

That is true for Sent, not Deleted, forms, but you have it under the Deleted Forms section.

ln 138, ln 140 `In the case of filled out forms that are...` -- this is too wordy for tech writing. 

 - `Sent Forms can be reviewed but not edited.`
 - `Deleted Forms are listed, but cannot be viewed. They are indicated with the crossed-out eye icon.`

mish24:
@adammichaelwood Actually I was running the emulator today only to check everything on my own. I think all of this is clear to me now. I will make the updates!

adammichaelwood:
Yes -- you should probably never write docs about things you can't actually see or interact with.

mish24:
@adammichaelwood Done. I think if the user is trying out the process as mentioned, this should clear up everything. Is there anything else I can add?

mish24:
screenshot from 2017-09-11 06-39-40

I tried to follow the bullet approach when there is more than a point, like here in the `Tips` section. How do think this should be addressed?

Issue #103, opened 2017-08-31
Switch Prev/Next buttons and file issue placement


adammichaelwood:
Would be better if page nav buttons were above the file an issue.

![screen shot 2017-08-31 at 12 12 52 pm](https://user-images.githubusercontent.com/1776241/29940977-be8f386c-8e45-11e7-9e03-176655b2e19b.png)

troublemagnet:
I'd like to work on it.

adammichaelwood:
Cool. You may want to coordinate with @mish24 working on #104 since that will likely touch the same template files. (It may not, actually - you'll have to dig in a bit.)

I think IDEALLY, text of the "file issue" notice would be pulled out of rst_epilog and put into a conf.py variable, which can then be sucked into the Jinja template directly. We probably want to style it nice to make it clear it isn't a part of the document.

adammichaelwood:
(Also, see my note on #104 about how to work with template files.)

adammichaelwood:
Closed via #130

mish24:
@adammichaelwood Can you close this?

adammichaelwood:
Oh - I meant to. My mistake.

PR #124, opened 2017-09-10
Swaps the navigation buttons with the "file an issue" notice.

troublemagnet:
swapped
The above screenshot reflects the change. The issue has been placed after the horizontal rule so that it doesn't appear to be the part of the document and not styled as the copyright text so that it doesn't look like the footer either.
Closes https://github.com/opendatakit/docs/issues/103 .


adammichaelwood:
Thank you for working on this!

The visual result is exactly what we want, but there's a couple problems now.

 - This conflicts with the recent PR #122 by @mish24, since it does not include those changes for displaying the title in the next/prev buttons. You need to pull in those changes. 
     - This will probably cause a merge conflict on your end, which you will need to resolve.
 
 - You have put the text of the "File an issue" notice into the template file. This violates basic web design principles. Please put the text and link targets into new variables in `conf.py` and pull them into the template.

PR #126, opened 2017-09-10
[WIP] Documenting the usage of QR codes

mish24:
screenshot from 2017-09-10 12-05-46

This is just a basic skeleton of the page that'll have the information. There are several issues like:
1. Representing `Import/Export settings` using :guilabel: tag.
2. There are no screenshots, and the ones I got from https://opendatakit.org/2017/06/configure-collect-on-many-devices-with-qr-codes/ have a very different resolution.
3. The names and titles need to be shortened.

More understanding of generating QR codes through a program (I don't know how that is done). @adammichaelwood A little help for that?

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -42,6 +42,7 @@ For a complete list of our projects, check out `Open Data Kit on Github <https:/
   collect-install
   collect-settings
   collect-forms
+  qr-codes
   
 .. toctree::
   :hidden:
--- a/qr-codes.rst
+++ b/qr-codes.rst
@@ -1,0 +1,61 @@
+***********************************************
+Configure Collect on many devices with QR codes
+***********************************************
+
+This document lists the ways in which QR codes can be used to configure same settings across many devices.
+
+.. _using-qr-codes-through-app:
+
+Using QR codes through the app
+==============================
+
+This section will cover the steps required to set up and use QR codes through the app itself.
+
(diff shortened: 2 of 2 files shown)


adammichaelwood:
this is the right direction. You'll need to provide a complete JSON example.
(You could get a real, live example by generating a QR code in app and then decoding and decompressing it. Be sure to notice, though, that default settings are omited - so you might not get a complete list of settings from a live install)

adammichaelwood:
Do you mean in the app or from a separate program?

mish24:
@adammichaelwood I just got it! I will update this PR soon. I'm really sorry for taking so long

mish24:
@adammichaelwood I am still on this, just that android studio and emulator take a lot of space and it's really really slow. Can I add the rest of the changes as of now? I'll add JSON example once I get it

adammichaelwood:
> emulator take a lot of space and it's really really slow. 

Yes. The docs even say so. :)

The PR isn't that helpful without the JSON format, so we can just hold it.
But it isn't urgent. If you come back around to this in a few days, that will be just fine.

adammichaelwood:
I mean -- you can whatever commits. But I'm not going to merge the PR until is complete.

mish24:
@adammichaelwood I am closing this PR due to conflicts. I'll open a new PR

PR #130, opened 2017-09-10
Swaps the navigate buttons with the file an issue notice.

troublemagnet:
swap

Please find the attached screenshot which reflects the changes. I didn't put the links and text in a single variable though because for some reason the linking wasn't working. However now the links and texts are stored in the conf.py.

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -47,3 +47,10 @@ caption .headerlink::after {
 .gesture {
     font-style: italics;
 }
+
+.footerText {
+    color: black;
+    font-family: Georgia,'Times New Roman',serif;
+    font-size: 16px;
+    font-weight: 300;
+}
--- a/_templates/footer.html
+++ b/_templates/footer.html
@@ -11,6 +11,8 @@
   {% endif %}
 
   <hr/>
+  <p class="footerText"> {{ prevar1 }} <a href="{{ link1 }}"> {{ hyper1 }} </a> </p>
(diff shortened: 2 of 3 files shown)


adammichaelwood:
Thank you for working on this. 

I'm not sure how I feel about the solution you found for dealing with the markup not rendering properly out of conf.py. Did you research other solutions? IS this a common problem? I'm thinking about accepting this, but I need to think on it more.

In the mean time -- Please remove the `style` stuff from the HTML. Add a class to the entire section and style it in one go in `/_static/css/custom.css`

troublemagnet:
Sir from what I could understand, I guess the reStructuredText can only be a part of certain variables. The variables that we define can't have reStructured text. However I shall research more and also fix the styling.

adammichaelwood:
I know rst can't be in the variables. Does plain html markup work?
The current solution creates a weird, tight coupling between the template and `conf.py`, which is not a good practice.

If you fix the inline style problem, I'll pull in the PR and come back and deal with the markup in another issue.

troublemagnet:
I fixed the inline style problem, and I'll try and research about the markup problem.

adammichaelwood:
In the future, if a PR is intended to close an issue, please add "Closes #x" (x = issue number) in the first PR comment. Then the issue will auto-close when the PR is merged.

Please open a new issue to deal with the markup stuff. It is not urgent, but I don't want to forget about it.

troublemagnet:
Really sorry sir. I usually put it the closes label, I forgot this time. And yes about the markup I'll open an issue.

adammichaelwood:
>  I forgot this time
No problem.

> about the markup I'll open an issue.
Thanks

> sir
You are awesome and polite and that is amazing.
You do not need to call me sir. :)

Issue #96, opened 2017-08-29
Line Edits


adammichaelwood:
This could be a perpetual issue, never closed.

Typos sneak in, and branch-based editing makes it inconvenient to go chasing them every time you notice them. So...

IF you are:

 - between issues
 - a little too sleepy in the afternoon to think about high-level stuff
 - a new comer looking for something to do

Start a line-edit branch, fix some typos, and submit a small PR.

## Guidelines for line-edit PRs

 - Typos and basic fixes only; don't rewrite or move sections (You *can* rewrite or move sections, but that needs to be done in individual Issue-Discussion-PR cycles.)
    - Spelling
    - Punctuation
    - Basic grammar
    - Markup corrections
    - Fix links (external and cross references)
 - If your aren't sure, err on the side of not changing something
    - If something needs long task-oriented discussion, file an issue
    - If you need quick clarity on some point of confusion, use the docs-code channel on the [ODK Slack](https://opendatakit.slack.com)
 - Run the build to make sure you didn't introduce errors
 - Don't use line edits to pursue personal grammar pet peeves
 - PR a single document at a time
 - line-edit PRs don't need a long description; the edits should speak for themselves
 - reference this issue number in your PR message, but don't say that it "closes" the issue

mish24:
@adammichaelwood Shall I work on this?

adammichaelwood:
Everyone can work on line edits at any time.
(Don't use the "Closes" syntax in your commit messages.)

nadaa:
Hi,
How can I start with this?
Thanks

adammichaelwood:
@nadaa 

Read the contributor's guide. That provides instruction for setting up your local environment and getting started. Then just pick a file that has spelling or punctuation errors and fix them.

If you need help getting your tooling and authoring environment set up, talk to us on the [developer slack](https://opendatakit.slack.com/), in the #docs-code channel.

nadaa:
Thank you @adammichaelwood.
I am following the contribution guide.  I use windows  (Cygwin) to setup the authoring tools, and I noticed there are no enough instructions in the guide. I will prepare a short doc to describe my approach.

Regards

adammichaelwood:
> I use windows (Cygwin) to setup the authoring tools, and I noticed there are no enough instructions in the guide. 

Yes. We need more details on the Windows experience. Most of us are Mac or Linux people, but we would like to encourage Windows users to contribute.

> I will prepare a short doc to describe my approach.

[...]

PR #135, opened 2017-09-11
Line edits for contributing.rst 

tanvibhakta:
This is with reference to Issue No. #96

These are a bunch of quick fixes to minor issues I found while reading contributions.rst. I figured that this was as good a place to start as any. 

Proposals: I would like to add a line talking about the variant of English used throughout the docs. I would also like to add some missing points in the "submit a PR" section wrt getting changes from upstream.

### DIFF
--- a/contributing.rst
+++ b/contributing.rst
@@ -221,7 +221,7 @@ You can install the SDK Platform tools (including :command:`adb`) as a `standalo
 
 .. _docs-workflow-setup:
 
-Get Ready to Work
+Getting Ready to Work
 -----------------------
 
 .. _fork-the-docs:
@@ -621,7 +621,7 @@ In order to facilitate efficient :ref:`cross-referencing`, sections should be la
 
 The section label is a sluggified version of the section title.
 
-Section titles must be unique throughout the entire documentation set. Therefore, if you write a common title that might appear in more than one document (*Learn More* or *Get Started*, for example), you'll need to include additional words to make the label unique. The best way to do this is to add a meaningful work from the document title.
+Section titles must be unique throughout the entire documentation set. Therefore, if you write a common title that might appear in more than one document (*Learn More* or *Getting Started*, for example), you'll need to include additional words to make the label unique. The best way to do this is to add a meaningful work from the document title.
 
 .. code-block:: rest
 
@@ -1169,7 +1169,7 @@ Image File Names
 Image file names should:
 
 - be short yet descriptive
-- contain only lower case charecters
+- contain only lower case characters
 - have no spaces
 - use hyphens as the separator
 
--- a/contributing.rst
+++ b/contributing.rst
@@ -102,7 +102,7 @@ Check to see if you have virtualenv installed:
 
   $ virtualenv
 
-If you get a help message with information about commands, you have it. If you don't have it, you'll get a ``command not found`` message. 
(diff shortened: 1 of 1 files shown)


contributing.rst
```diff
@@ -221,7 +221,7 @@ You can install the SDK Platform tools (including :command:`adb`) as a `standalo
 
 .. _docs-workflow-setup:
 
-Getting Ready to Work
+Get Ready to Work
```
adammichaelwood: This is an editorial change, not a typo fix. 
Please revert. 

(We should have a discussion about gerunds in titles separately.)

contributing.rst
```diff
@@ -619,9 +619,9 @@ In order to facilitate efficient :ref:`cross-referencing`, sections should be la
 
   Lorem ipsum content of section blah blah.
 
-The section label should usually be a sluggified version of the section title.
+The section label is a sluggified version of the section title.
```
adammichaelwood: But it isn't always.

contributing.rst
```diff
@@ -1166,12 +1166,12 @@ Image File Names
...
 
-Image file names should be:
+Image file names should:
 
-- as short as possible while still being descriptive
-- all lower case
-- no spaces
-- hyphen separators
+- be short yet descriptive
+- contain only lower case charecters
```
adammichaelwood: characters

contributing.rst
```diff
@@ -632,7 +632,7 @@ Section titles must be unique throughout the entire documentation set. Therefore
 
   .. _aggregate-getting-started:
 
-  Getting Started
+  Get Started
```
adammichaelwood: Editorial, not typographical.

contributing.rst
```diff
@@ -619,9 +619,9 @@ In order to facilitate efficient :ref:`cross-referencing`, sections should be la
 
   Lorem ipsum content of section blah blah.
 
-The section label should usually be a sluggified version of the section title.
+The section label is a sluggified version of the section title.
 
-Section titles must be unique throughout the entire documentation set. Therefore, if you write a common title that might appear in more than one document (*Learn More* or *Getting Started*, for example), you'll need to include additional words to make the label unique. The best way to do this is to add a meaningful work from the document title.
+Section titles must be unique throughout the entire documentation set. Therefore, if you write a common title that might appear in more than one document (*Learn More* or *Get Started*, for example), you'll need to include additional words to make the label unique. The best way to do this is to add a meaningful work from the document title.
```
adammichaelwood: Editorial.

contributing.rst
tanvibhakta: If there isn't a reason it shouldn't be then shouldn't we push the narrative that it is? Otherwise it might just creates confusion in the mind of the reader - or at least that was my thinking.

adammichaelwood: I see your point. But here's my thinking:

 - I don't want someone coming through and "helping" by doing line edits to correct existing section labels.
 - Sometimes you want a short, clear label but it makes sense to have a longer, descriptive, human-readable title.
 - Sometimes you change the human readable title. But the label should typically not be changed, as it might have cross references.

You might be right about this, though. 
But a line edit isn't the place to have this discussion.

Add a note to the Style Guide issue, that this is something we should hash out. In the mean time, leave this instruction in place.

tanvibhakta: Thank you for your detailed reply. I will do as you suggested and add these points as a comment to the Style Guide issue.

PR #138, opened 2017-09-11
[WIP] Added information about adb

mish24:
I made a separate document first, but then there was a lot of redundancy in the content of `collect-forms.rst` and using `adb`, so I tried to merge the contents together. I added less information about what it is and where to get it (just gave the link for that). This PR just added these primary informations:

* Location of form instances
* Pull and delete commands in adb
* Screencap feature

(These are the features I've tried as of now. So sure about these)

Things yet to be added:

* What is 'hangups'?
* installing `adb` as standalone tool (this just exists for linux and windows I guess)

### DIFF
--- a/collect-forms.rst
+++ b/collect-forms.rst
@@ -2,6 +2,16 @@
 Form Management in Collect
 *****************************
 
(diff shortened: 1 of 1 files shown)


adammichaelwood:
It needed to be a separate section so that we have a single point of reference when discussing adb throughout the docs.

And putting info about adb at the top of the document doesn't make a lot of sense, as you can use Collect and manage forms without it. It is an optional tool for "power users."

By "hangups" I meant, potential problems one might have while installing or setting up.

mish24:
@adammichaelwood Okay I'll move it as the last section for Collect. What should I name this doc?

adammichaelwood:
"Android Debug Bridge"

mish24:
It's Android Developer Bridge in the current doc. I'll change it

adammichaelwood:
My mistake, probably. 
(Developer Bridge would have been a better name, but that's not what Android calls it.)

https://developer.android.com/studio/command-line/adb.html

PR #146, opened 2017-09-12
Added extra info about Admin settings

mish24:
This adds the basic difference between both the settings and its main purpose, which wasn't clear in the docs before.

### DIFF
--- a/collect-settings.rst
+++ b/collect-settings.rst
@@ -3,6 +3,8 @@ Collect Settings and Security
 
 Settings are accessible from the :guilabel:`⋮` menu.
 
+There are two kings of setting, **General Settings** and **Admin Settings**. As the names suggest, this is really helpful when the person who is collecting the data and the owner of the device are not same. So the admin can set an admin password and choose what all general settings the enumerator (in this case, who is not the owner) can access. This feature can also hide/show the items displayed in the main menu and the options available in the options menu while filling the form. 
+
 .. _general-settings:
 
 General Settings
--- a/collect-settings.rst
+++ b/collect-settings.rst
@@ -3,7 +3,7 @@ Collect Settings and Security
 
 Settings are accessible from the :guilabel:`⋮` menu.
 
-There are two kings of setting, **General Settings** and **Admin Settings**. As the names suggest, this is really helpful when the person who is collecting the data and the owner of the device are not same. So the admin can set an admin password and choose what all general settings the enumerator (in this case, who is not the owner) can access. This feature can also hide/show the items displayed in the main menu and the options available in the options menu while filling the form. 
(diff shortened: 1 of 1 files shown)


collect-settings.rst
```diff
@@ -3,6 +3,8 @@ Collect Settings and Security
 
 Settings are accessible from the :guilabel:`⋮` menu.
 
+There are two kings of setting, **General Settings** and **Admin Settings**. As the names suggest, this is really helpful when the person who is collecting the data and the owner of the device are not same. So the admin can set an admin password and choose what all general settings the enumerator (in this case, who is not the owner) can access. This feature can also hide/show the items displayed in the main menu and the options available in the options menu while filling the form.
```
adammichaelwood: This is a good addition. Can you do a self edit for brevity and clarity.

 - Avoid filler: "As the name suggests"
 - Avoid qualifications and value judgements like "really helpful"
 - Generally, avoid beginning a sentence with "So...."
 - s/"what all"/"which"

Aim for clarity, and brevity in service of clarity.

mish24:
@adammichaelwood I am so sorry sir. I will not repeat such mistakes.

collect-settings.rst
```diff
@@ -3,7 +3,7 @@ Collect Settings and Security
 
 Settings are accessible from the :guilabel:`⋮` menu.
 
-There are two kings of setting, **General Settings** and **Admin Settings**. As the names suggest, this is really helpful when the person who is collecting the data and the owner of the device are not same. So the admin can set an admin password and choose what all general settings the enumerator (in this case, who is not the owner) can access. This feature can also hide/show the items displayed in the main menu and the options available in the options menu while filling the form. 
+There are two kings of setting, **General Settings** and **Admin Settings**. This is used when the person who is collecting the data and the owner of the device are not same. Admin can set an *admin password* and choose what all general settings the enumerator (in this case not the owner) can access. You can also hide/show the items displayed in the main menu and the options available in the options menu while filling the form.
```
adammichaelwood: s/kings/kinds

adammichaelwood: > This is used when the person who is collecting the data and the owner of the device are not same.

I don't think you mean to suggest that admin and general settings are only used when the owner is not the enumerator.

mish24: "generally useful"? or just 'useful' is okay?

adammichaelwood: i don't understand your question. I wouldn't say "useful" at all. Don't tell people something is "useful for x" just say how things work and what they do, and what is possible.

Also...
This information is already in the document. I'm not sure why you feel the need to add it here.

adammichaelwood:
I'm not sure about the need for this PR at all, actually. The info about admin security is already in the document.

mish24:
@adammichaelwood Is it better to move this in the best practices/tips section?

adammichaelwood:
> Is it better to move this in the best practices/tips section?

Just not sure it is needed.

The problem with redundant information in docs is that if something changes, you have to go fix it in more than one place.

Because human-readable content is **not** code, you can't be completely DRY (don't repeat yourself) -- sometimes you have to duplicate info in text. But, if it can be helped, it's best not to do it.

PR #147, opened 2017-09-12
Add a new video directory and update references.

troublemagnet:
Html_extra_path variable contains the files that are copied to the HTML build directory. However when we put ['vid'] in that, what it does is that it copies the content of the vid directory to the build directory. So at the moment after the build it'll transfer the form-widgets directory. Hence the markup has been edited accordingly. I also checked it by adding some temporary directory in the 'vid' directory.


conf.py
```diff
@@ -266,6 +266,6 @@ def setup(app):
 html_context = {'prevar1' : prevar1 , 'prevar2' : prevar2 , 'prevar3' : prevar3 , 'hyper1' : hyper1 , 'hyper2' : hyper2 ,
                 'hyper3' : hyper3 , 'link1' : link1 , 'link2' : link2 , 'link3' : link3}
 
-
+html_extra_path = ['vid']
```
adammichaelwood: Please move this declaration up near the html static path and theme options stuff, and include a comment on what it is.

adammichaelwood:
Excellent work. I made one comment about moving the variable declaration and commenting it. Make that change and I'll merge.

(And I'm glad you realized you could do it inside conf.py instead of having to update the build script.)

Issue #98, opened 2017-08-29
Style generic admonitions


adammichaelwood:
cf. http://docutils.sourceforge.net/docs/ref/rst/directives.html#admonitions

The theme we are using has built-in styling for all specific admonitions (note, tip, warning, etc.), but the generic admonition is does not follow the others.

Example markup and view:

```rst

  .. note:: Forms from Aggregate
  
    Before downloading blank forms from Aggregate to Collect, those forms have to be uploaded to the Aggregate server.
  
  .. admonition:: Forms from Google Drive
  
    If you are going to use Google Drive to manage forms, you won't need to use the XLSForm or Build apps. Just create a Google spreadsheet, following the `XLSForm <http://xlsform.org/>`_ specification.


```

![screen shot 2017-08-29 at 2 56 23 pm](https://user-images.githubusercontent.com/1776241/29846108-494a2fc0-8cca-11e7-9319-95fd05453452.png)

The `admonition` style should match the `note` style.

Minimally, we should implement this in our own `custom.css`.
If someone had the skills and bandwidth to do it in the source and [contribute it back to the theme's repo](https://github.com/rtfd/sphinx_rtd_theme), that would be good citizenship.

adammichaelwood:
(It doesn't matter in connection to this issue, but... I am aware the information about Google sheet in the screenshot above is not right.)

troublemagnet:
I'd like to do it.

troublemagnet:
@opendatakit-bot  claim

adammichaelwood:
Great. 

Styles should go in `custom.css`

Let me know if there's anything unclear here.

troublemagnet:
My question is that we're not using admonition anywhere right now, so should I still work on it?

adammichaelwood:
We're not using any because the style is messed up so I have avoided them.

It is not high priority, but if you think you can do it easily and quickly, it's a good "quick win"

PR #157, opened 2017-09-13
[WIP] Class added to style generic admonition like note.

troublemagnet:
A new class called genericstyle has been added for styling generic admonitions. The styling is same as the styling for note. Please find the screenshots attached which show how an admonition looks like and how we can override the default styling. 
black
heyy

Closes https://github.com/opendatakit/docs/issues/98.

### DIFF
--- a/_static/css/genericstyle.css
+++ b/_static/css/genericstyle.css
@@ -1,0 +1,23 @@
+.genericstyle {
+	background-attachment: scroll;
+	background-clip: border-box;
+	background-color: rgb(231, 242, 250);
+	background-image: none;
+	background-origin: padding-box;
+	background-position: 0% 0%;
+	background-position-x: 0%;
+	background-position-y: 0%;
+	background-repeat: repeat;
+	background-size: auto auto;
+	box-sizing: border-box;
+	color: rgb(64, 64, 64);
+	font-family: Georgia,"Times New Roman",serif;
+	font-size: 16px;
+	font-weight: 400;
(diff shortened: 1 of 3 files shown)


troublemagnet:
@adammichaelwood This issue https://github.com/sphinx-doc/sphinx/issues/3754 has the description of why the build fails. Turns out that switching to sphinx 1.5.6 is the solution.

adammichaelwood:
The CSS work is great.

But this is the wrong approach for adding new styles. We don't need a new CSS file, which means we don't need to update `conf.py` or `header.html`

The new style class should be appended to the existing `_static/css/custom.css` (with a code comment explaining what it is, what it does, and referencing the issue number). No other changes should be required or included in this PR.

adammichaelwood:
Also: 

Is there a way to style the admonition without forcing the user to add a class to the markup? Are there any admonition-specific classes already present in the html output which can be used?

troublemagnet:
@adammichaelwood I shall research more and will try to find out an alternative way to do it. Also I'll move it to custom.css.

adammichaelwood:
Just look at the output in your browser dev tools. There's got to be something already there to hang the styles on.

adammichaelwood:
Are you going to complete this PR?

troublemagnet:
@adammichaelwood yes sir, I'm on it.

troublemagnet:
I'll fix this locally and make a new PR ASAP.

PR #162, opened 2017-09-14
Added initial data for documenting tips

mish24:
Just made a doc for displaying tips/best practices etc.

Things to do -
* Add more relevant data about GPS and mobile encryption.
* Provide more links

@adammichaelwood I thought it's better to start a PR to document what we have already from the issues opened as well. Data can be gathered to elaborate on these points. I'm trying to find some points like these myself too. Plus I already asked on forum, let's see what old users reply!

### DIFF
--- a/collect-tips.rst
+++ b/collect-tips.rst
@@ -2,6 +2,6 @@
 Best Practices and Tips
 **************************
 
-- Specifying the GPS requirements can help users decide the accuracy of location they send and hence help save battery.
+- Specifying the GPS requirements can help users decide the accuracy of location they send and hence help save battery. For more information on enabling GPS services and its relation with battery life visit `this site <https://support.google.com/nexus/answer/3467281?hl=en>`_.
 
-- Rather than using encrypted forms, enabling Android-level data encryption can secure the data more efficiently.
+- Rather than using encrypted forms, enabling Android-level data encryption can secure the data more efficiently. Encrypting the phone stores the data in an unreadable and scrambled form as long as it's locked. Whenever you enter your password, the data becomes readable. To encrypt your device, follow the instructions given `here <https://www.howtogeek.com/141953/how-to-encrypt-your-android-phone-and-why-you-might-want-to/>`_. 
--- a/collect-tips.rst
+++ b/collect-tips.rst
@@ -4,4 +4,4 @@ Best Practices and Tips
 
 - Specifying the GPS requirements can help users decide the accuracy of location they send and hence help save battery. For more information on enabling GPS services and its relation with battery life visit `this site <https://support.google.com/nexus/answer/3467281?hl=en>`_.
 
-- The forms are not encrypted until they are finalized. However, the form data can be encrypted while they're in saved state by turning on Android device encryption. To encrypt your device, follow the instructions given `here <https://www.howtogeek.com/141953/how-to-encrypt-your-android-phone-and-why-you-might-want-to/>`_. 
+- Encrypting the form ensures that data is safe even when sent over an HTTP connection and that it can be stored on a server that is not secure. The forms are not encrypted until they are finalized. However, the form data can be encrypted while they're in saved staten also by turning on Android device encryption. To encrypt your device, follow the instructions given `here <https://www.howtogeek.com/141953/how-to-encrypt-your-android-phone-and-why-you-might-want-to/>`_. 
--- a/collect-tips.rst
+++ b/collect-tips.rst
@@ -1,0 +1,7 @@
+**************************
+Best Practices and Tips
+**************************
+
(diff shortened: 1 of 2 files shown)


adammichaelwood:
Ideally, each one of those bullet points would be a small section explaining how to do the thing (how **does** someone use Android-level encryption?)

I think the sanest organizational approach is to merge this document as is, and then anyone can do PRs adding the individual sections, since they are independent issues.

Do you agree? Do you want to flesh out a section before merging?

mish24:
@adammichaelwood Yes it has to be more thorough than this. I can complete these two points properly before they're merged.

adammichaelwood:
Great. Let me know when it's in shape to merge.

mish24:
@adammichaelwood I found some good sites which can prove really helpful for these topics. I'll add more data subsequently if I find anything. I think this can be merged for a start

collect-tips.rst
```diff
@@ -0,0 +1,7 @@
+**************************
+Best Practices and Tips
+**************************
+
+- Specifying the GPS requirements can help users decide the accuracy of location they send and hence help save battery. For more information on enabling GPS services and its relation with battery life visit `this site <https://support.google.com/nexus/answer/3467281?hl=en>`_.
+
+- Rather than using encrypted forms, enabling Android-level data encryption can secure the data more efficiently. Encrypting the phone stores the data in an unreadable and scrambled form as long as it's locked. Whenever you enter your password, the data becomes readable. To encrypt your device, follow the instructions given `here <https://www.howtogeek.com/141953/how-to-encrypt-your-android-phone-and-why-you-might-want-to/>`_.
```
lognaturel: Using encrypted forms and encrypting the device solve different problems so they should not be presented as alternatives. Encrypting the form ensures that data is safe even sent over an HTTP connection and that it can be stored on a server that is not secure. Encrypting the phone is a good additional precaution because forms are not encrypted until they are finalized.

I would prefer https://support.google.com/nexus/answer/2844831?hl=en as the resource for using device-level encryption.

mish24: @lognaturel Oh, I didn't know this. Guess I'll read more about it and update the contents accordingly. Thank you for pointing this out :)

lognaturel: Clearly it needs to be documented!

mish24:
@adammichaelwood I updated this info.

collect-tips.rst
```diff
@@ -0,0 +1,7 @@
+**************************
+Best Practices and Tips
```
yanokwa: These are more Android tips than they are Collect tips, so I'd frame them as such!

collect-tips.rst
```diff
@@ -0,0 +1,7 @@
...
+**************************
+
+- Specifying the GPS requirements can help users decide the accuracy of location they send and hence help save battery. For more information on enabling GPS services and its relation with battery life visit `this site <https://support.google.com/nexus/answer/3467281?hl=en>`_.
```
yanokwa: This conflates GPS (a location technology) with location. I think this could be a lot more precise. I'd start with "You can configure your device's location mode based on accuracy, speed, and battery use. For more information on location mode and it's relationship to battery life, visit..."

collect-tips.rst
```diff
@@ -0,0 +1,7 @@
...
+
+- Encrypting the form ensures that data is safe even when sent over an HTTP connection and that it can be stored on a server that is not secure. The forms are not encrypted until they are finalized. However, the form data can be encrypted while they're in saved staten also by turning on Android device encryption. To encrypt your device, follow the instructions given `here <https://www.howtogeek.com/141953/how-to-encrypt-your-android-phone-and-why-you-might-want-to/>`_.
```
yanokwa: There are multiple ways that encryption interacts with ODK and I think it's important to spell all three out or it can confuse implementers.

1. You can encrypt the transition of data between device and server. This is what happens when you use HTTPS. It's a good idea to do and besides having to install an SSL certificate on your server (it comes automatically on App Engine) has no real downsides.
2. You can use public/private key encryption to encrypt finalized forms at rest. https://opendatakit.org/help/encrypted-forms spells this out. This has some downsides in that you encrypt only on finalizing and once you do, you can't edit. You also need ODK Briefcase to download and decrypt so you can't view data on the server.
3. You can encrypt everything on your Android device, which is what I think this tip was originally about. If your device supports this, it's definitely a good idea because it secures everything if your device gets lost/stolen.

collect-tips.rst
yanokwa: We should also note that for most deployments, you'll want to use "High accuracy" true GPS.

adammichaelwood:
I'm closing this PR to clean things up.
It can be reopened, or a new one filed.

PR #165, opened 2017-09-15
intent doc fixed 

smartyrad:
Addresses #4

### DIFF
--- a/intents.rst
+++ b/intents.rst
@@ -4,15 +4,15 @@ Collects's Intent
 About Intent
 -------------
 
-*Intent* is a messaging object used to request an action from another app component.
+**Intent** is a messaging object used to request an action from another app component.
 
 Each page pertaining to an android app is called an activity. ODK Collect has several activities and sub-activities. These activities are able to communicate with each other using intents(as mentioned above).
 
 There are two types of intents:
 
-1.Explicit Intent: We use Explicit Intent when we want to communicate between 2 or more activites inside the *same* application.
+1. **Explicit Intent** : We use Explicit Intent when we want to communicate between 2 or more activites inside the *same* application.
 
-2.Implicit intent: We use Explicit Intent when we want to communicate between 2 or more activites inside *different* applications.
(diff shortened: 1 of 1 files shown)


adammichaelwood:
This is coming along. 

Can you please add a paragraph at the top of the document explaining what an intent is in the context of Collect, and what they are used for?

smartyrad:
@adammichaelwood Sure :+1:

intents.rst
```diff
@@ -0,0 +1,136 @@
...
+This will allow user to choose the form and go on filling it .
+
+Similarly for an instance of the form : 
+
+.. code-block:: java
+
+	Intent intent = new Intent(android.intent.action.EDIT);
+	intent.setData(org.odk.collect.android.provider.InstanceProviderAPI.InstanceColumns.CONTENT_URI);
+
+Apart from this , we also need to put in a mode, which will allow to differentiate between editing or viewing form . So we need to put in an extra here , which goes like :
```
lognaturel: This is no longer right. We should not document extras, the code above should only be for launching form edits.

intents.rst
```diff
@@ -0,0 +1,136 @@
+Collects's Intent
```
lognaturel: I think this should be something like Launching ODK Collect From External Apps.

intents.rst
```diff
@@ -0,0 +1,136 @@
...
+==================
+
+About Intent
```
lognaturel: I think this section should focus on why someone might want to launch Collect from their app. Linking out to the Android docs is enough general background on Intents.

lognaturel:
Thanks for looking into this @smartyrad and working with us to figure out what the right content is!

adammichaelwood:
@smartyrad 
Can you address @lognaturel's reviews and update the PR?
Thanks.

adammichaelwood:
@smartyrad Are you still working on this?

adammichaelwood:
Closing due to inactivity.

Issue #109, opened 2017-08-31
guilabel and menuselection style


adammichaelwood:
1. Visual style
2. Some thought about rules of when to use which. (Most cases are obvious, but there's some gray areas.)

troublemagnet:
@adammichaelwood sir what exactly is this issue about?

troublemagnet:
@opendatakit-bot claim

adammichaelwood:
Probably shouldn't claim issues before you understand what they are about.

The roles for `guilabel` and `menuselection` are part of Sphinx. (see the contrib guide and also the sphinx documentation) 

`guilabel` has some styling built into the theme. 
`menuselection` does not.

`menuselection` needs some kind of clear styling, probably similar to `guilabel`

Also, some uses of each are unclear which would be better. (For example, is the home screen of the app a "menu" ?) We need some kind of thought-through clarity about when to use each and why.

I'm certain you can take of the styling issue. So if you'd like to do that you can.
The issue of clarifying usage is something we need to discuss and think about.

troublemagnet:
Apologies for unprofessional behavior sir. I'd work on the styles and will also research about the two options available.

adammichaelwood:
Anyone have an opinion on:

> 2. Some thought about rules of when to use which. (Most cases are obvious, but there's some gray areas.)

troublemagnet:
Sir I don't see how the home screen of app can be "menu" ?

adammichaelwood:
What is a menu other than a listing of items you can select?

mish24:
According to Sphinx guide - wherever there's selection of options, like `Menu A` ---> `Setting B` ----> `Print C`, `menuselection` must be used.

adammichaelwood:
Menu selection is used for any menu. The arrow notation is specifically for nested menus,  but it is not essential to the concept of menu selection.

troublemagnet:
Sir quoting the sphinx guide , 
>  This is used to mark a complete sequence of menu selections, including selecting submenus and choosing a specific operation, or any subsequence of such a sequence. The names of individual selections should be separated by -->.

So sir I think it's usage is to represent a sequence of menu selections.

lognaturel:
@troublemagnet Don't forget that a sequence can have a size of 1! it's not the most interesting sequence but it is a sequence.

mish24:
[...]

PR #173, opened 2017-09-16
Menuselection styling added

troublemagnet:
Menuselection has been styled similar to guilabel. The border of the box has been made black which makes it similar to guilabel but not same.
Addresses https://github.com/opendatakit/docs/issues/109.
menu


adammichaelwood:
I think this is a good solution. 

Let's all keep an eye on it to think about how it does or doesn't improve readability.

PR #178, opened 2017-09-20
Made the styling match for generic admonition and notes

### LINKED ISSUE #98: Style generic admonitions

troublemagnet:
Style generic admonitions same as the note. 
Closes #98

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -113,3 +113,29 @@ caption .headerlink::after {
     -moz-border-right-colors: none;
     -moz-border-top-colors: none;
 }
+
+[class^="admonition-"] {
+      
+               background-attachment: scroll;
+               background-clip: border-box;
+               background-color: rgb(231, 242, 250);
+               background-image: none;
+               background-origin: padding-box;
+               background-position: 0% 0%;
+               background-position-x: 0%;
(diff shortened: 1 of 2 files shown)


troublemagnet:
@adammichaelwood I included it in custom.css.

Issue #66, opened 2017-08-16
Video sphinx directive 


yanokwa:
Currently, images get stored at `img`, but videos are stored at `_static/vid` and it's not clear why. Unless there is a good reason, we should standardize on the location.

yanokwa:
This is likely related to the appearance of a `build/_images` folder when Sphinx is run and thus is related to https://github.com/opendatakit/docs/pull/67. Can you offer an explanation, @adammichaelwood?

adammichaelwood:
The "good reason" is that it is currently the only way to do it.

Sphinx copies images from wherever they are stored to the `/_images` dir as part of the build process. It doesn't copy the entire `/img/` folder -- it copies the image when it finds the `.. image` or `.. figure` directive. 

OTOH - the video files are currently in `_static` -- Sphinx copies over the entire `_static` dir on build (it's typically used for custom CSS and JS files. And I am inserting videos with raw HTML... which is *wrong* (from an aesthetic and consistency standpoint) but the only current way to do it, as Sphinx doesn't  have a `.. video` directive.

The *right* way to do this would be to create a Sphinx plugin to handle video. It would then do precisely what the `.. image` directive does: output proper markup and copy the file over. This would be a benefit because it would be consistent across media types. It would also benefit the larger Sphinx community, as this is something other people need to.

But I don't know that we want to take the time to write and then maintain a Sphinx plugin. I don't *think* it is that big of a deal, and I've been wanting to do it for a while, but we all know how quickly "it's just 100 or so lines of code" becomes a gigantic time sink.

So....  

The real question is: Does anybody *really* mind the current situation, which is:

 - files stored in more than one place in the repo 
 - two different markup constructions for images and video

And, is it worth a day or so of work to fix those two problems?

yanokwa:
Thanks for explaining this, Adam. 

I don't think it's a good use of time to change the current behavior. Or rather, maybe this is the kind of thing a contributor might want to take on, but it's not a priority for the core team.

jumperskirt:
is there a python library equivalent to PIL for video?

adammichaelwood:
Probably, but I'm not sure how that would be relevant to this issue.

Could you clarify what you are considering?

jumperskirt:
[...]

PR #187, opened 2017-09-22
WIP: Video directive for Sphinx

jumperskirt:
video directive written in video.py
added node and directive to conf.py
issue #66
error when i attempted to build:

> Exception occurred:
  File "conf.py", line 206, in setup
    from . import video
KeyError: "'__name__' not in globals"

was not sure how to configure conf.py and ss.py

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -19,7 +19,10 @@
 #
 import os
 import sys
+
 sys.path.append(os.path.abspath('docs'))
+
+extensions = ['video']
 # sys.path.insert(0, os.path.abspath('.'))
 
 
--- a/conf.py
+++ b/conf.py
@@ -17,12 +17,8 @@
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 #
-import os
-import sys
-
-sys.path.append(os.path.abspath('docs'))
-
-extensions = ['video']
+# import os
+# import sys
 # sys.path.insert(0, os.path.abspath('.'))
 
 
@@ -206,6 +202,15 @@ intersphinx_mapping = {'https://docs.python.org/': None}
 
 def setup(app):
     app.add_stylesheet('css/custom.css')
+
+    from . import video
+
+    app.add_node(video,
+                 html=(visit_video_node, depart_video_node),
+                 latex=(visit_video_node, depart_todo_node),
+                 text=(visit_video_node, depart_video_node))
+
+    app.add_directive('video', VideoDirective)
 
 
 # At top of every document
--- a/conf.py
+++ b/conf.py
@@ -17,8 +17,9 @@
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 #
-# import os
-# import sys
+import os
+import sys
+sys.path.append(os.path.abspath('docs'))
 # sys.path.insert(0, os.path.abspath('.'))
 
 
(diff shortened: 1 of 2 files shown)


adammichaelwood:
I can't really do anything with this. Debugging an error is one thing, but it isn't clear what you think this is supposed to even accomplish.

jumperskirt:
thank you for your feedback

lognaturel:
@jumperskirt Could you please say a little bit more about what you think this should do? It looks like it comes from [here](https://github.com/ryfeus/lambda-packs/blob/master/Selenium_PhantomJS/source/docutils/parsers/rst/directives/images.py) (it should be credited!) and is meant for figures and simple images -- did you see a reference that suggested it could be used with videos?

jumperskirt:
hello! first, i surely didn't mean to not credit; where does one credit? in
the commit message? as a comment in the file itself? second, yes, the
video.py i added is just the image directive with 'image' swapped out for
'video', and then i added video as a custom extension in the config.py,
along with the desired output formats

i did this because after going through the spinx tutorial, and the
reStructured docs and adam's input that processing the video was likely
outside the scope of the extension, i thought the video directive would
look and act like the image directive (and therefore copy over the videos
and put them in their own vid directory upon build, like it does currently
for img);

i knew i hadn't gotten it right, but i was trying to run the program to see
if the error messages could help me figure out what i needed to be
rethinking. i was having trouble understanding the aforementioned error
message and submitted the pr hoping for guidance in that area. is there a
better way to ask for code review besides gists or pull requests?

at any rate, it seems i'm not on the right track, so i'll go back over the
tutorial and welcome any other suggestions you might have

adammichaelwood:
Thanks for this.

A PR is the right venue for a code review.
I just need a little more info on what your thinking is and what you are trying to accomplish.

Building off an existing image directive makes some sense, as there are definitely similarities.

Here's what I recommend:

 - start with the unmodified code, and see if you can get things to work
     - this probably involves diving into how to load extensions in Sphinx
 - then modify and see how that updates the output

Changing one thing at a time makes it more clear where the problem is.

jumperskirt:
will do, thank you

nadaa:
@jumperskirt, please check [this](https://groups.google.com/forum/#!topic/sphinx-users/_z00m3zoRAY).

jumperskirt:
@nadaa hello! thanks for sharing; i too ran into that thread during my internet peregrinations; it seems to be explaining the way everyone (odk included) is currently doing the video embedding (without the video directive i am attempting to write)—did i miss something?

@adammichaelwood i did manage to get rid of the errors i was getting before by just starting over and trying one thing a a time as you suggested; i now have a video directive written  and get a message that the build succeeded, but with a [different] error and with no video output. i will submit a [WIP] PR, and if you have any suggestions while i continue to try different things, i'd greatly appreciate it

adammichaelwood:
you can commit on the same branch and keep everything in this PR

adammichaelwood:
> and with no video output.

restructured text has an internal representation of each node in a document. each builder has a specific way of rendering those nodes.

Since images are a standard part of the document model, every builder has a representation for it. But if you add a new node type ("video"), the html builder has no understanding of that node, so it doesn't have any way of rendering it.

So a new rst directive that introduces a new document node type has to include details on how the builder(s) should render it.

If you look at the sphinx contrib plugins, i'm sure at least a few of them cover a similar use case and would provide some insight into how to provide rendering help to the builder through the plugin.

jumperskirt:
yes, thank you for the feedback! i have been taking a look at the youtube
plugin but will check out others too

adammichaelwood:
Closing due to inactivity.

PR #198, opened 2017-09-23
Added differences between Briefcase and Aggregate

### LINKED ISSUE #17: Differences between Aggregate and Briefcase
The practical difference between Aggregate and Briefcase is often clear — for example, many features are available in one but not the other — but I've struggled with the conceptual difference. Where does Aggregate end and Briefcase begin? For example, when do you export data from Aggregate vs. export from Briefcase?

I discussed with Yaw and Hélène, who described the following differences and suggested I create an issue in the docs repo:
* Briefcase is good for offline use, for when there's no need for a server. Briefcase is often enough for smaller projects. However, note that Aggregate can also be used offline and locally.
* Briefcase can be used to back up all forms and submissions on Aggregate. In particular, Briefcase's command line interface makes this easier.
* Briefcase is needed to transfer data between Aggregate installs.
* Briefcase is required to decrypt any encrypted submissions.

ankita240796:
Closes #17

### DIFF
--- a/briefcase-vs-aggregate.rst
+++ b/briefcase-vs-aggregate.rst
@@ -1,0 +1,51 @@
+**********************************
+Briefcase vs Aggregate
+**********************************
+
+.. _briefcase-points:
+
+Briefcase
+-------------
+
+- Briefcase provides a mechanism to pull blank and finalized forms from ODK Collect or ODK Aggregate or any custom ODK directory into a local briefcase storage location.
+- It can push blank forms and finalized forms from briefcase storage location to an Aggregate 1.0 instance.
+- Export the finalized form to a CSV file for processing by other applications.
+- It can be used to transfer data across incompatible upgrades of ODK Aggregate as well as it is required to decrypt any encrypted submissions.
+- Briefcase is good for offline use when there is no need for a server. It is often enough for smaller projects.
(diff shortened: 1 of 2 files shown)


briefcase-vs-aggregate.rst
```diff
@@ -0,0 +1,51 @@
+**********************************
+Briefcase vs Aggregate
+**********************************
```
adammichaelwood: Please change this to the underline style:

```
Title
=======
```

The over-under style is for larger docs that *might* be broken up in the future. 
When the over-under style is used, the first subsection headline with the `===` underline.

briefcase-vs-aggregate.rst
```diff
@@ -0,0 +1,51 @@
...
+
+.. _briefcase-points:
+
+Briefcase
+-------------
+
+- Briefcase provides a mechanism to pull blank and finalized forms from ODK Collect or ODK Aggregate or any custom ODK directory into a local briefcase storage location.
```
adammichaelwood: capitalize `Briefcase`
(here, and throughout)

index.rst
```diff
@@ -52,6 +52,7 @@ For a complete list of our projects, check out `Open Data Kit on Github <https:/
    briefcase-guide
    briefcase-install
    briefcase-forms
+   briefcase-vs-aggregate.rst
```
adammichaelwood: remove file extension in TOC directive

ankita240796:
I made the requested changes.

PR #221, opened 2017-09-27
edit pr template

### LINKED ISSUE #214: PR Template Question unclear
I wrote:

> ## What new work does this PR create?

What I meant was: 

 - What will you or someone else now have to do that wasn't needed before?
 - What future work are we committing to because of this PR?

Apparently that was not clear.

Is this a bad question? Is there a better way to say it?

adammichaelwood:
closes #214 

I changed the unclear question to:
What new issues will need to be opened because of this PR?

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -9,7 +9,7 @@ addresses #
 
 
 <!-- Answer any that apply and delete the others. -->
-## What new issues will need to be opened because of this PR??
+## What new issues will need to be opened because of this PR?
 
 ## What is left to be done in the addressed issue?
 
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -9,7 +9,7 @@ addresses #
 
 
 <!-- Answer any that apply and delete the others. -->
-## What new work does this PR create?
+## What new issues will need to be opened because of this PR?
```
mish24: `What new issues does this PR create?` conveys the same message. Isn't it better?

adammichaelwood: hrm...

That depend on how you are using the word "issue".
I think for precision, we should only use "issue" to mean "numbered tasks tracked by github".

lognaturel: Let's try this one and keep iterating if needed!

PR #234, opened 2017-09-29
Add run instructions

lognaturel:
Simple steps for building and viewing the docs once one's environment is set up. The contribution guide is fantastic for getting started but too detailed to use as a reference. This addition means those familiar with Sphinx can quickly get up and running and those who have gone through the contribution guide can get a quick reminder.


mish24:
@lognaturel We can add this link too -- https://docs.opendatakit.org/contributing/#work-on-the-docs.

Issue #226, opened 2017-09-28
Add "Create Issue" template


mish24:
Here I want to add a similar template as there is for PR now. This would have:
* Guide for title format (bug fix, suggestion)
* Description
* Steps to recreate (if applicable)
* People to notify
* Resources or links for reference which the contributor can provide us.

adammichaelwood:
I have mixed feelings about this, but I'm willing to give it a try.

I think we need to keep in mind that most of what we have for issues in docs are not bugs, but rather new documents, new sections, or things like that. So we don't want a lot of things people have to delete every time.

But we can try it out, iterate on it, and either use it or get rid of it if it doesn't seem helpful.

mish24:
@adammichaelwood Yes. That's true. I opened one PR which just has everything has comments for reference.

adammichaelwood:
I've merged. I'm keeping the issue open as a place to discuss whether this works and is helpful.

troublemagnet:
I had a doubt. Why is one of the titles as BUG FIX? Shouldn't it only be Bug? I mean there're cases when contributors find bugs and they create an issue so they could fix it, in that case BUG FIX is fine I guess. but what if a non tech person wishes to report a bug so that the developers can fix it. In that case BUG FIX doesn't make sense.

adammichaelwood:
also, i'm not sure how often a docs issue is going to be a bug that needs fixing.

adammichaelwood:
I find myself deleting the issue template every time I file one.
Any one else?

ankita240796:
@adammichaelwood  I find myself doing the same and no new issue has used the template. I feel we should remove this template.

lognaturel:
I agree. It was a good thing to try but since docs are inherently more subjective and varied it doesn't work as well as with code.

PR #236, opened 2017-09-29
Added Issue template

mish24:
addresses #226

A template for creating issues


mish24:
@adammichaelwood Everything here is just a comment. What do you think?

adammichaelwood:
I still have mixed feelings about it, but I don't think it hurts to try it out.

PR #243, opened 2017-09-29
Added information and edited lines for the environment set-up

Hbler:
addresses #96 

I little more detailed information for the local environment set-up, and a couple of line edits.

There may be more details to be added to other sections of the page, but these ones are ones that I had some trouble with while setting up my local environment.

code lines with missing code, and not mentioned details in some steps.


contributing.rst
```diff
@@ -306,30 +306,30 @@ The rest of the documentation assumes you are in the directory for the repo (the
...
       - docs/
-	
+
 .. _upstream-the-docs:
 
 Set the Upstream Remote
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-When you clone down a repo, the local copy calls your GitHub copy ``origin``. You should also set ``upstream`` as the name of the original, main GitHub repo.
+When you clone down a repo, the local copy calls your GitHub copy ``origin``. You should also set ``upstream`` as the name of the original master branch of the main GitHub repo.
```
adammichaelwood: That change is not accurate.
 
'upstream` refers to the whole repo, not specifically the master branch

contributing.rst
```diff
@@ -306,30 +306,30 @@ The rest of the documentation assumes you are in the directory for the repo (the
...
 
 
 .. code-block:: console
 
-  $ git remote add --track upstream https://github.com/opendatakit/docs.git  
+  $ git remote add --track master upstream https://github.com/opendatakit/docs.git
```
adammichaelwood: I don't think this change is correct. It looks like you are adding a remote named master, not a remote named upstream

I do think we need to remove `--track` from this. 
I can't remember why it is here, and it isn't part of the git documentation...

contributing.rst
```diff
@@ -345,7 +345,7 @@ Run ``git remote -v`` to check the status, you should see something like this:
 Install Dependencies
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-The first time you clone down the repo, you'll need to install the dependencies. Make sure you have your Python 3 virtual environment set up and activated and then:
+The first time you clone down the repo, you'll need to install the dependencies. Make sure you have your Python 3 virtual environment set up and activated from the docs directory and then:
```
adammichaelwood: That is not actually what we recommend.

contributing.rst
Hbler: Alright, thanks for the correction.
This one was actually my last edition to the file to make it more consistent with the alterations that I've made further in the code.

contributing.rst
Hbler: Yes, about this one, while I was setting up my local environment the first time I pasted this line in my terminal console I've got the following error:

usage: git remote add [<options>] <name> <url>

    -f, --fetch           fetch the remote branches
    --tags                import all tags and associated objects when fetching
                          or do not fetch any tag at all (--no-tags)
    -t, --track <branch>  branch(es) to track
    -m, --master <branch>
                          master branch
    --mirror[=<push|fetch>]
                          set up remote as a mirror to push to or fetch from

and I saw that the before the name of the remote, when using the `--track` option one must define which branch(es) should be tracked. At least that what I understood from this message.

I solved it in my local environment by substituting the `--track` for the `--fetch`, then I was able to get the entire repo.

But before altering this part of the documment I also tested with way (`--track` option + branch name before the upstream name) and it worked. 

So I opted for the edition that would change less of the documment that was adding the branch name after the `--track` instead of changing the in code line the `--track` to `--fetch`).

But if we need to remove the `--track` what shoud we replace it with?

adammichaelwood: I think if the command was just `git add remote upstream`, everything would work just fine.
Specifying  a tracking branch doesn't really seem to matter, and it doesn't change anything else we ask people to do.

contributing.rst
Hbler: This one I've changed also because of tests I've performed while setting up my local environment. 

From reading the documentation I understood that I should activate the virtual environment where it was created so I left the recently cloned *docs* directory and went to the previously created *odkenv* directory and after I activated the virtual environment and ran the command to install the dependencies I got the error message that the file requirements.txt didn't exist.

So I left the *odkenv* directory to the *odk* folder that has both the *odkenv* and the *docs* directories and tried again, got the same message.

My last try was: activate the virtual environment from within the *docs* directory with the code `$ source ../odkenv/bin/activate` and then run the command to install the dependencies, and that worked, no error messages. 

That is why I felt that this information should be in the text.

adammichaelwood: It doesn't matter where you activate the environment (thought that obviously affects the specific path you have to type).
The issue is that you have to be in the docs repo to `pip install` from the requirements file

contributing.rst
Hbler: Alright, I'll edit just the line of code to the one you suggested.

contributing.rst
Hbler: So maybe change the phrasing of what I added to the text to:
> 
...virtual environment set up and activated in the docs repo and then:

Would that be more accurate?

Also (sorry, git is new to me), do the cloned _**docs**_ repo resides just inside the _**docs**_ directory in my local file system? Or should I be able to access it from the _**odk**_ directory or even from the _**odkenv**_ one?

Issue #254, opened 2017-09-29
Validate 


adammichaelwood:
https://opendatakit.org/use/validate/

This page needs to be checked, edited, and pulled in.
The person doing this should actually install and use the tool.

Also, the existing page is poorly organized (it includes usage instruction under the heading "Installation"), so it could stand a little re-writing for clarity.

This will be a new doc in the Form Builder TOC.

nadaa:
I will work on this @adammichaelwood.

nadaa:
@opendatakit-bot  claim

nadaa:
@opendatakit-bot label  "in progress"

ankita240796:
@adammichaelwood This issue has to be closed.

adammichaelwood:
closed via #266

PR #266, opened 2017-10-04
Validate odk

nadaa:
addresses #254

Validating ODK forms taken from the old documentation.

I guess we can add more examples to show various validation errors and ways to solve them.

There is not much info on the old site about the tool.


ankita240796:
@adammichaelwood This issue has to be closed.

PR #268, opened 2017-10-04
Minor fixes

ManognaMutyala:
closes #

addresses #96 

Edited incorrect spellings

## What new issues will need to be opened because of this PR?
Nothing


adammichaelwood:
the legit typos you found were also fixed in another PR.
The added spaces did not need to be added.

adammichaelwood:
Thank you for your effort.

To have more successful PRs in the future:

 - Don't make assumptions about non-standard spacing or spelling when a doc references something in an application. Look into the application.
 - For line edits, only change one file at a time.

ManognaMutyala:
Thanks for the info,I will take care of those instructions from now on.

PR #273, opened 2017-10-06
Use s3_website to enable easy redirects

### LINKED ISSUE #272: 301 Redirects in case of page renaming
Came up in Slack, at suggestion og @lognaturel 

A page's URI in Sphinx is determined by its filename (not it's place in the TOC).
There may be cause from time to time, to rename a document's file.
We should probably have a way of mapping old names--> new names.

We could watch PRs for file name changes (git can surface these), and then append a line to some redirect-config file, and then deploy that to the server...(?)

This is mostly going to be on @yanokwa to figure out.

yanokwa:
Closes #272 

This PR swaps out the old AWS CLI tool with [s3_website](https://github.com/laurilehmijoki/s3_website). s3_website makes uploading static sites to S3 easier than the AWS CLI, and for this particular issue, it enables users to add redirects to a single file. I've used s3_website for years and I use it on all my static sites. The alternative is a series of `aws cp` commands for each redirect we need, but that feels hacky and unsustainable. 

I verified that this worked by testing on a master build and running the commands manually. There aren't a ton of risks to deploying this, but it's important to check the website immediately after the build to make sure nothing is broken.

Please squash before merging.


s3_website.yml
```diff
@@ -0,0 +1,33 @@
...
+  default_cache_behavior:
+    min_ttl: <%= 60 * 60 * 24 %>
+  aliases:
+    quantity: 1
+    items:
+      - docs.opendatakit.org
+
+cloudfront_wildcard_invalidation: true
+
+# redirects:
```
yanokwa: @adammichaelwood Here is where redirects would go.

s3_website.yml
```diff
@@ -0,0 +1,33 @@
+s3_bucket: docs.opendatakit.org
```
yanokwa: s3 secrets and keys are already stored in the CircleCI build environment.

PR #287, opened 2017-10-10
Style guide

### LINKED ISSUE #121: Style Guide
We're going to need one.

adammichaelwood:
closes #121 

A new doc, with rules and advice on writing.

As time goes on, I'm sure we'll add to this guide. But I think it will be a good start for now.


adammichaelwood:
This needs an official maintainer review from @yanokwa or @lognaturel 

If any of our Outreachy applicants are interested in doing a peer-review, feel free.

style-guide.rst
```diff
@@ -0,0 +1,625 @@
...
+
+Node
+~~~~~
+
+The word *node* is often used interchangeably with *element*.
+
+For clarity, we make the following distinction:
+
+- An *element* is a piece of XML or HTML code.
+- A *node* is an element rendered into a DOM tree or other dynamic representation.
```
Divya063: I think this is quite confusing as there are different types of nodes like element node, document node, etc., and a node is a part of DOM tree, this should be the other way around.I guess **An element is a particular kind of node** would be accurate.

adammichaelwood: This is a good point.

The thing I was trying to get across is that some code in a ` *.xml` file is **not** a node. A node is something in a "live" DOM tree. 
Look at my next commit and see if it makes more sense then.

style-guide.rst
```diff
@@ -0,0 +1,628 @@
...
+  <element>  <!-- The opening tag of this element. -->
+    Some content.
+  </element> <!-- The closing tag. -->
+  
+The word *tag* has often been used to refer to the entire element. For clarity, we will avoid that here.
+
+
+.. _xml-node:
+
+Node
```
adammichaelwood: @Divya063 Is this section better?

Divya063: @adammichaelwood Yes, this looks fine 😃

style-guide.rst
```diff
@@ -0,0 +1,628 @@
...
+
+  Apples, oranges and pears.
+  
+.. rubric:: Right
+
+.. code-block:: rst
+
+  Apples, oranges, and pears.
+  
+A bulleted list is often more clear than an inline list.
```
ankita240796: I feel we should also have a tip on when to use a bulleted list and when to use a numbered list i.e a note like
> A numbered list is preferred over a bulleted list when you have a series of related points.

adammichaelwood: This is a good point.

style-guide.rst
```diff
@@ -0,0 +1,628 @@
...
+-----------------------
+
+.. _adverbs:
+
+Adverbs
+~~~~~~~~~~~
+
+Adverbs often contribute nothing.
+
+Common offenders include:
```
yanokwa: I'd remove the previous whitespace and bump this up as a second sentence.

style-guide.rst
```diff
@@ -0,0 +1,628 @@
...
+American spelling and grammar
+-----------------------------
+
+Whenever U.S. English and British (or other) English spelling or usage disagree, standard U.S. spelling and usage is preferred.
+
+.. rubric:: Wrong
+
+.. code-block:: rst
+
+  The colour of the button is grey
```
yanokwa: You need a period after this sentence.

style-guide.rst
```diff
@@ -0,0 +1,628 @@
...
+
+Semicolons are used to separate two independent clauses which could stand as individual sentences but which the writer feels would benefit by close proximity.
+
+Semicolons can almost always be replaced with periods (full stops). This rarely diminishes correctness and often improves readability.
+
+.. rubric:: Correct
+
+.. code-block:: rst
+
+  These "canned phrases" are pervasive in technical writing; remove them whenever they occur.
```
yanokwa: On a small screen, these code blocks don't wrap nicely. It'd be good if they did.

style-guide.rst
```diff
@@ -0,0 +1,628 @@
...
+
+  ODK Collect is an Android app. The same can be used for...
+  
+.. rubric:: Right
+
+.. code-block:: rst
+
+  ODK Collect is an Android app. It can be used for...
+
+  .. OR THIS:
```
yanokwa: Maybe have another "Right" section?

style-guide.rst
```diff
@@ -0,0 +1,628 @@
...
+Document titles should be in ``Title Case`` -- that is, all meaningful words are to be capitalized.
+
+Section titles should use ``Sentence case`` -- that is, only the first word should be capitalized, along with any proper nouns or other words usually capitalized in a sentence.
+
+.. _title-verb-forms:
+
+Verb forms
+-----------
+
+If a document or section describes a procedure that someone might do, use a gerund. Do not use the "How to..." construction.
```
yanokwa: I had to look up gerund. Maybe add a link or an example?

yanokwa: Also, why not "How to". I kinda like the phrase. And I think people might search for "How to"

style-guide.rst
```diff
@@ -0,0 +1,628 @@
...
+
+.. rubric:: Wrong
```
yanokwa: Add a link because I don't know what it is.

style-guide.rst
```diff
@@ -0,0 +1,628 @@
...
+=============
+
+.. _writing-about-odk:
+
+ODK and ODK Docs
+-------------------
+
+.. rubric:: Wrong
+
+- Odk
```
yanokwa: People often also prepend "The" and in my humble opinion, that's terrible. It's "Open Data Kit" not "The Open Data Kit".

style-guide.rst
```diff
@@ -0,0 +1,628 @@
...
+
+A few projects should *always* use the **ODK** prefix:
+
+- ODK XForm
+- ODK Javarosa
+- ODK Docs
+
+.. _xform-xlsform:
+
+XForm and XLSForm
```
yanokwa: It's XForms (with an s). An individual form is called an XForm. 

XLSForm, confusingly, doesn't have an s. It probably should, but that ship has sailed...

style-guide.rst
```diff
@@ -0,0 +1,628 @@
...
+.. _semicolons:
+
+Semicolons
+-------------
+
+Semicolons are used to separate two independent clauses which could stand as individual sentences but which the writer feels would benefit by close proximity.
+
+Semicolons can almost always be replaced with periods (full stops). This rarely diminishes correctness and often improves readability.
+
+.. rubric:: Correct
```
yanokwa: Sometimes you use Right, other times you use Correct. Please standardize!

style-guide.rst
```diff
@@ -0,0 +1,628 @@
...
+  def myFunction(foo):
+
+    for bar in foo:
+       bar[foo] = foo[spam] + spam[foo]
+
+    return foobar
+
+.. _xml-html-style-guide:
+
+XML and HTML
```
yanokwa: We don't really use HTML in ODK, so I don't think the HTML section is relevant. XML, is fine.

style-guide.rst
```diff
@@ -0,0 +1,628 @@
+*****************
+Style Guide
+*****************
+
+.. _spelling-and-grammar:
```
yanokwa: It'd be good to add a section on acronyms. That is, avoid if possible.

yanokwa:
This is looking great, Adam. Was this Style Guide inspired by something else? I ask because if we didn't, it might be good to look at a few big OSS projects and see if we can learn from their work.

style-guide.rst
adammichaelwood: Agree.

style-guide.rst
adammichaelwood: Definitely.

style-guide.rst
adammichaelwood: ok

style-guide.rst
adammichaelwood: LAWD what was I doing?!?!

style-guide.rst
adammichaelwood: yes, but that can't be solved here. I have opened #305 to address it

style-guide.rst
adammichaelwood: good point

style-guide.rst
adammichaelwood: Maybe it would be better to not use the word "gerund" since that is sort of unhelpful jargon.

> Why not "How to"

This is a good question. My ready answers are:
 - it feels right
 - it has become the standard in the TW community
 - "how to" is redundant -- most of documentation is how to

there are probably better answers

style-guide.rst
adammichaelwood: ok

style-guide.rst
adammichaelwood: I agree with you in theory, but-
I can't think of any examples of *bad* acronyms or initialisms.

We use, all the time:

 - ODK
 - XML
 - adb
 
We should avoid Latin, though:

 - e.g.
 - i.e.
 - etc.

style-guide.rst
adammichaelwood: OH! I REMEMBER WHAT I WAS DOING

I used "Right" vs. "Wrong"
but...
"(Technically) Correct" vs. "Improved"

style-guide.rst
adammichaelwood: I asked around for why *not "how to"* has become standard. The most compelling reason is that it makes a document or table of contents harder to scan, because of the redundancy.

style-guide.rst
adammichaelwood: well...

There's almost nothing here that is HTML specific (except the one short section on null tags), so I don't see the harm in leaving it in. Also, it might come up in docs about docs templating.

adammichaelwood:
> Was this Style Guide inspired by something else?

Mostly I was just trying to address specific issues that have already come up. I do read a lot of style guides, though. They're a favorite genre for me.

adammichaelwood:
I think this is ready to go. @yanokwa

style-guide.rst
```diff
@@ -0,0 +1,717 @@
...
+
+Therefore, avoid the use of personal pronouns whenever possible. They are often out of place in technical writing anyway. Rewriting passages to avoid personal pronouns often makes the writing more clear.
+
+.. rubric:: Correct
+
+.. code-block:: rst
+
+  When using Collect, first the enumerator opens the app on their device. Then they complete the survey.
+  
+.. rubric:: Improved
```
yanokwa: You used Improved here, but Better everywhere else. Was that your intention?

adammichaelwood: No. I'm just losing my mind is all.

PR #292, opened 2017-10-11
Line edit for Aggregate use guide and oauth2-service

ankita240796:
addresses #96 

Minor corrections in aggregate-use.rst and oauth2-service.rst.

### DIFF
--- a/aggregate-use.rst
+++ b/aggregate-use.rst
@@ -153,7 +153,7 @@ If you selected `Columns` specify the columns you wish to display or hide in the
 .. image:: /img/aggregate-use/column-filter.*
    :alt: Image showing column selection.
 
-- Click on :guilabel:`Save` to save the filter or filter group for future use. Clicking on :guilabel: `Save As` allows you to give a name to the filter or filter group.
+- Click on :guilabel:`Save` to save the filter or filter group for future use. Clicking on :guilabel:`Save As` allows you to give a name to the filter or filter group.
 - Click on :guilabel:`Delete` to delete a filter or filter group.
 - You can check the :guilabel:`Display Metadata` checkbox to display or hide metadata.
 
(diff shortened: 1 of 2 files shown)


aggregate-use.rst
```diff
@@ -302,20 +302,20 @@ Preferences
...
-   - ODK Tables Synchronization Functionality - check this to enable ODK Tables functionality i.e.  able to download, upload and alter data records within ODK Tables as restricted by table-access privileges granted to the user.
+   - `ODK Tables Synchronization Functionality` - check this to enable ODK Tables functionality i.e.  able to download, upload and alter data records within ODK Tables as restricted by table-access privileges granted to the user.
 
-- Aggregate features: These settings affect the operations of the server.
+- **Aggregate features**: These settings affect the operations of the server.
    
-   - Disable faster background actions - check this to reduce AppEngine quota usage.
-   - Skip malformed submissions - check this to ignore corrupted submissions.
+   - `Disable faster background actions` - check this to reduce AppEngine quota usage.
+   - `Skip malformed submissions` - check this to ignore corrupted submissions.
```
adammichaelwood: I'm not sure what the purpose of single backticks are...

ankita240796: @adammichaelwood I added them just to make the checkbox options look different from the explanation. Should I remove them?

adammichaelwood: Single back-ticks have no semantic meaning in RST or Sphinx. (Unlike MArkdown, where they wrap literal inline blocks.)

 - If the test is verbatim UI text from the app, use the `:guilabel:` role.
 - If they should be interpreted as `literal inline text`, use double-backticks
 - If they need to be semantically set-off, but no particular markup is appropriate, you can use *single asterisks*.

PR #297, opened 2017-10-13
Added upgrade guide for Aggregate

### LINKED ISSUE #228: Add documentation on upgrading Aggregate
We should have a new doc on how to upgrade Aggregate with details of various versions. @adammichaelwood  What do you think?

ankita240796:
closes #228 

- Added need for upgrading Aggregate
- Added upgrade steps and release notes for Aggregate
- Added upgrade admonition for Upgrade notes

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -139,3 +139,37 @@ caption .headerlink::after {
                padding-top: 12px;
           
 }
+
+.upgrade {
+   background-attachment: scroll;
+   background-clip: border-box;
+   background-color: lemonchiffon;
+   background-image: none;
+   background-origin: padding-box;
+   background-position: 0% 0%;
+   background-position-x: 0%;
+   background-position-y: 0%;
+   background-repeat: repeat;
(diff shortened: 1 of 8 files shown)


adammichaelwood:
This is good work.

I'm very curious about why you created a Upgrade Notes admonition, instead of using the existing Notes or Warning. I'm not opposed to the idea, but I don't really understand it.

ankita240796:
@adammichaelwood I felt that since each version of Aggregate requires upgrade notes and they are different from general notes, tips. Further this admonition can also be used when we add upgrade docs in any other section. Also I think this improves readability. 
What do you think ?

adammichaelwood:
So what is the difference between the upgrade notes, and the other notes also in the upgrade document?

ankita240796:
Normal notes have general information but here we have the information specific to upgrade process.

adammichaelwood:
I'm merging because the information is good.
I'm going to open an issue to discuss the Upgrade Note thing -- we might refactor it out.

Issue #63, opened 2017-08-08
Automate the form widget guide from a structured XLSForm


adammichaelwood:
It would be great if someone with Android/Java skills could work with me to create a script (using monkeyrunner, or similar) that would parse an XLSForm and run through the form on a device, taking screenshots and outputting the images and other info into a doc -- (replacing form-widgets.rst)

anamika2910:
@adammichaelwood Can I work on this?

adammichaelwood:
That would be AWESOME.

The relevant form is:
https://docs.google.com/spreadsheets/d/1af_Sl8A_L8_EULbhRLHVl8OclCfco09Hq2tqb9CslwQ/edit?usp=drive_web

The end result doesn't need to be *exactly* like form-widgets.rst, but it should be pretty close. And the whole point is to have a robust enough solution that we can edit the all-widgets form and generate a new form-widgets.rst at any time.

I suggest making a copy of all-widgets sheet and adding the relevant text descriptions and image alt-text, and then using that as a starting point.

The script should be Python 3 if at all possible, so it can be run in the same environment as the rest of the docs tools.

If you have questions, ideas, suggestions, or need more discussion -- I'm available here. Or you can just go dive in. I'm very excited to have you working on it.

ngumbau:
Just wondering if you have considered using [screengrab](https://docs.fastlane.tools/getting-started/android/screenshots/) from fastlane.

adammichaelwood:
That looks like a really good alternative to monkeyrunner. Thanks.

The fact that it is in Ruby presents an added complication to our stack. However, that might not be a problem. I don't actually expect that every docs contributor will need to run this screenshot utility. It is just something that would need to get run once per Collect release.

Whoever ends up actually building this feature will decide which tools are best suited (or, probably, easiest to get working...)

ngumbau:
[15:09 - 20:11 Automation with fastlane](https://www.youtube.com/watch?v=Nr5RHaQa3Uc). It actually works with expresso tests. No Ruby required except for installing command line tools.

PR #299, opened 2017-10-15
WIP: Add PDF output

### LINKED ISSUE #271: Add PDF output
We *could* build a PDF as part of the build process, and make it available at some URL
This came up in the forum.

I don't know if it is actually a good idea:

 - The docs change enough that I don't know whether a PDF would be valuable
 - Some amount of work would be required to get the PDF template just right, so the value has to be weighed against that work

@yanokwa would have to add pdf deploy and build steps to circleCI, 
I, or someone else with some LaTeX experience would have to futz with the template.

troublemagnet:
addresses #271
 
This PR builds a PDF as part of the build process. To produce the pdf, you can simply run sphinx-build -b pdf . build. 
Please find the pdf  [here.](https://drive.google.com/file/d/0B5Qi6LfWkjN-SFRNUUMtSVVCdW8/view?usp=sharing)
rst2pdf does not work with python3, hence leaves this PR for discussion.

### DIFF
--- a/b.sh
+++ b/b.sh
@@ -2,4 +2,6 @@
 
 rm -rf build/*
 sphinx-build -b dirhtml . build
+sphinx-build -b pdf . build       
+
 


adammichaelwood:
This is good work, but obviously I can't merge it in at this point.

Can you look into what would be required to apply custom css style (or similar) so we can fix the formatting.

Also, we need to figure how to use this in our Python 3 environment. Options (in order of my preference):
 - there are some tools for executing py2 in a py3 environment. we could look into that
 - we could run pdf-build in a separate py2 environment, and not require contributors to build and test pdf locally
 - we could switch to py2 (I hate this option the most)

troublemagnet:
@adammichaelwood true, we shouldn't go for the third one. I'll try and fix it ASAP.

adammichaelwood:
Look into the py2 inside py3 possibility... 
If we could make that work, it would solve some other problems as well
(There's a py2-only replacement for Monkeyrunner, which would help with #63.)

Also, I just want to know if styling is possible -- you don't have to fix/finish it. We just need to know a way forward.

PR #310, opened 2017-10-19
Fix overflow lines in code-blocks

### LINKED ISSUE #305: Code samples require side scrolling
If a line in a code sample is too long, you have to scroll sideways.

I don't think this is ideal... but maybe it it (?).

At any rate, I think the fix is just some CSS to wrap overflow lines in code blocks.

If someone could do it, and submit a PR -- at least we could see if it is an improvement.
(It might not be.)

Note: I can imagine it being the case that some languages wrap better than others. For example, `rst` blocks will probably read much better when wrapped, but XML might not. Looking into how css classes are applied based on language, and making it something we could choose on a per-language basis would be extra cool.

ankita240796:
closes #305 

- Added wrap and unwrap class.
- Added default wrap for rst code-blocks.
- Added note on using wrap and unwrap classes.

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -173,3 +173,15 @@ caption .headerlink::after {
     padding: 6px 12px;
     margin-bottom: 12px;
 }
+
+div[class^='highlight-rst'] pre {
+    white-space: pre-wrap;
+}
+
+div[class^='wrap'] pre {
+    white-space: pre-wrap;
+}
+
+div[class^='unwrap'] pre {
+    white-space: pre;
+}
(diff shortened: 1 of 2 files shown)


ankita240796:
selection_004
The fix works fine but space is left blank inside a block in some cases.
selection_003
Box-sizing can be changed in this case but I am not sure whether I should do this or not since box-sizing is same for all the blocks.

adammichaelwood:
I like the idea of having wrapping on an as-needed basis. 
But I don't like so much clutter in the source files. There's little call for that.

Here's my thought:

 - a `wrap` class, as you have it

ankita240796:
Thank you for the review @adammichaelwood. I will make the required changes.
I will take care about spacing. There was some indentation error due to which I had to this.

ankita240796:
@adammichaelwood I have made the requested changes.

PR #318, opened 2017-10-23
Describe jump button. 

### LINKED ISSUE #186: Jumping between questions while filling a form
While filling a form, an enumerator can jump between questions or skip to the beginning or end by tapping on the arrow icon at the top of the screen. The icon is not currently all that intuitive so I think including a screenshot would be worthwhile.

Describe what shows up in the jump menu, especially when repeats are involved. This is not obvious as the conversation in https://forum.opendatakit.org/t/looped-group-broken/9946 demonstrates.

I think this should go in the [Filling out forms](http://docs.opendatakit.org/collect-forms/#filling-out-forms) section as a subsection.

troublemagnet:
closes #186 

Added information about jump button in the "filling a form" section.

### DIFF
--- a/collect-forms.rst
+++ b/collect-forms.rst
@@ -60,14 +60,22 @@ Once you have at least one form :ref:`loaded into ODK Collect <loading-forms-int
 - Typically, you will :gesture:`swipe left` to advance forward through the question, and :gesture:`swipe right` to back up.
 
   - You can switch from :gesture:`Swipe` navigation to Forward/Backward buttons in :menuselection:`⋮ -> General Settings -> User Interface`
-  - The |arrow| icon on the top right corner opens the jump menu. From the jump menu, you can go to any question or go to the beginning/ending of the form.
-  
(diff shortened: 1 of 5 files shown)


adammichaelwood:
This is good.

A couple things would make it better

### Image spacing

There's a problem with the margin on the screen picture. I'm not sure what is causing it.
Typically, images have a bottom margin of 24px. This one does not, and there is no clear reason why. (I suspect something about being inside a list, which is a little weird.

If you can track that down and correct it, that would be great. 

### Show jump icon in place

You show a picture of the jump menu, but not the screen that actually has the icon on it. I think showing that screen would be helpful.

adammichaelwood:
Oh -- also --

Can you file an issue about inline images? What you did to show that icon in the text is a useful thing that other people might want to do.

troublemagnet:
Yes sir, found it here - http://www.sphinx-doc.org/en/stable/rest.html.

Issue #323, opened 2017-10-24
Build failing because of bad png?


yanokwa:
The command that is preventing master from shipping releases is
`pngquant $HOME/docs/build/_images/jump*.png --force --ext .png`

It fails with on remote server (Ubuntu 14.04)
```
jumpmenu.png:
  error: Not a PNG file
  error: rwpng_read_image() error 25
jumpscreen.png:
  error: Not a PNG file
  error: rwpng_read_image() error 25
```
This command works fine on my local machine (macOS 10.13). 

* I've confirmed that the command fails on remote source image
`pngquant ~/docs/img/collect-forms/jump*.png --verbose`
* I've confirmed that the command fails on remote build image 
`pngquant ~/docs/build/_images/jump*.png --verbose`
* I've confirmed that the MD5 hashes between local/remote are the same.
```
MD5 (jumpmenu.png) = 1a4caec3e2c4bc858b6ea8c1d1b2611c
MD5 (jumpscreen.png) = 700448df36766436573aa0f6cd7e739a
```

* Remote pngquant is 2.0.1 and local pngquant is 2.9.1, but I've built from source on remote (v2.10.2) and that fails too.
```
/home/ubuntu/docs/img/collect-forms/jumpmenu.png:
  error: Not a PNG file (libpng failed)
  error: cannot decode image jumpmenu.png
/home/ubuntu/docs/img/collect-forms/jumpscreen.png:
  error: Not a PNG file (libpng failed)
  error: cannot decode image jumpscreen.png
```

https://github.com/opendatakit/docs/pull/318/files is the pull request that is the root of this problem. @adammichaelwood just to fix this, can you send in a PR with an updated image, so we can fix the build?

adammichaelwood:
These images were created by @troublemagnet

adammichaelwood:
These images were created by @troublemagnet 

I can reproduce them... but is there any way to find out why they are failing?
The images don't seem corrupt otherwise (they are viewable in GH and locally when I build)

yanokwa:
Yeah, I can't repro it either and I've spent way too much time on it. I'd propose you take new screenshots because we know your process works and then merge that in to fix the build.

Then I have a PR (https://github.com/opendatakit/docs/pull/324) that will check "bad" PNGs in the future and we can try to repro it then.

adammichaelwood:
Rather than recreate the images, I thought it would be faster to pull them into gimp and re-export them as new pngs.

[...]

PR #324, opened 2017-10-24
Move pngquant as part of test. Move s3_website as part of deploy.

yanokwa:
Bad pngs can cause pngquant to fail before a build goes to production (See https://github.com/opendatakit/docs/issues/323). We should catch this issue when the initial PR is sent in. For this reason, we should run pngquant as part of the test. This makes for slower builds when a PR is sent in, but so be it. I'd rather we catch this before we merge.

We only need to use s3_website when deploying to production, so rather than install it on each commit, we only install it when deploying to production.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,7 +1,6 @@
 dependencies:
     pre:
       - sudo apt-get install pngquant
-      - gem install s3_website && s3_website install
 
 test:
   pre:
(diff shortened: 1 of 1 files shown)


yanokwa:
Please squash and merge!

yanokwa:
One random note so I don't forget. One alternative way to speed up PR builds is to use this command on the test: `pngquant ... --speed 11` and then use this command on deploy: `pngquant ...`

I decided against that approach because:
1. There might be something different in the speed implementations (doubt it)
1. Philosophically, we should just use deploy to deploy and not run different commands.
1. The time difference of the various flags at the time of this message is ~ 4 minutes. I think people can wait.

PR #327, opened 2017-10-26
Added documentation on projecting ODK Collect on to another screen

### LINKED ISSUE #306: Projecting ODK Collect onto another screen
For users who need to train others on how to use ODK Collect, it would be very nice to have instructions on how to project ODK Collect on a screen.

According to https://forum.opendatakit.org/t/showing-odk-collect-form-with-projector/10063, a number of people use and like Vysor, so we should link to that and provide high-level instructions. It's great for users who have an Android device and USB debugging turned on. The upside is that it's easy to get started. The downside seems to be ads every 30 minutes (or paying $40 for it, which is a fair price!)

For more advanced users, we should point to instructions on how to install the Android emulator and get ODK Collect installed. The upside of the emulator is that it's free, you don't need a device, and you can install different versions of Android. The downside is that it's a bit big and slow and a pain to install. We want to reduce the pain of that install and unless there is a great reason, we don't want to have people use Android Studio.

I'm imaging that this document would be part of some deployment tips guide.

Oh, and whomever takes on this issue needs to *actually use Vysor* and *actually use the emulator* and make sure it all works nicely, preferably on Windows.

Divya063:
closes #306 

While I'm having all the images files in my working branch, in some image files there is this message `Git LFS file not shown`.

### DIFF
--- a/img/project-collect/acceleration.png
+++ b/img/project-collect/acceleration.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:968a0514f1abb1b778077d3cd144c7cbe80310d09d19fe71b2b0f03052d8c262
+size 44824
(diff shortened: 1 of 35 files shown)


adammichaelwood:
This is good information and I'm looking forward to merging it in.

Can you do an editing pass on it?
There's a number of typos, and several places where the space after a period is missing.

Issue #332, opened 2017-10-27
Re-organize Form Building and Developer-facing


adammichaelwood:
Started with the work I'm doing related to #239, but also move JavaRosa out to its own section and probably put JavaRosa and OpenRosa together in a dev-docs section...

Issue #239, opened 2017-09-29
Form Design Page *and Children* from old site (XForm Docs discussion)


adammichaelwood:
@mish24 has recently worked on pulling in the [Form Design Page](https://opendatakit.org/help/form-design/) -- #217 

But that page has a bunch of child pages dealing with specific XForm issues:

- Binding
- Body
- Guidelines
- Data Preloading
- Cascading Selects
- External Itemsets
- External Apps
- OpenStreetMap
- iText
- Examples
- XLSForm
- Gotchas and Workarounds

Do we think all this should be folded into docs?  
Or should it be deprecated altogether?   
Moved into the XForm specification repo?

mish24:
we'll have XForm specific repo for documentation? The points mentioned here are important for understanding forms in general. Some of these things are important for *form breakdown*, and others for *form creation*, and there two parts make up the entire content of Form Design.

adammichaelwood:
> we'll have XForm specific repo for documentation? 

Exists already:

https://github.com/opendatakit/xforms-spec

It's really for the specification only. In theory, usage tutorials and human-readable explanations could (should?) belong here in docs. But it's tough to delineate precisely where the line is.

mish24:
@adammichaelwood  Exactly. It's difficult to draw a line here since everything is inter-dependent and nothing can be left out if someone is actually following these docs for understanding.

adammichaelwood:
Will do first pass on collating/combining into something usable

adammichaelwood:
Everything from this list either:

- has been migrated
- will not be migrated
- has another issue with more details 

I am therefore closing this issue.

PR #333, opened 2017-10-27
Form design intro

adammichaelwood:
addresses #239 
addresses #332

A new doc at the top of the Form Building TOC that provides an overview of how forms work in ODK ecosystem. 

This is roughly analogous to https://opendatakit.org/help/form-design/

The [original page](https://opendatakit.org/help/form-design/) has a number of sub pages, only some of which will be pulled in.

I'm also going to move some pages around and do some editing, but since this doc is useful immediately, I thought it was more sensible to do a standalone PR for it.

### DIFF
--- a/form-design-intro.rst
+++ b/form-design-intro.rst
@@ -1,5 +1,37 @@
-Intro to Form Design
+Intro to Forms in ODK
 ========================
 
-Forms used in the ODK ecosystem are XML documents following the `ODK XForms specification <https://opendatakit.github.io/xforms-spec/>`_, a subset of the `W3C XForms specification <https://www.w3.org/TR/xforms/>`_.
+Forms used in the ODK ecosystem are XML documents following the `ODK XForms specification <https://opendatakit.github.io/xforms-spec/>`_, a subset of the `W3C XForms specification <https://www.w3.org/TR/xforms/>`_. Most ODK tools use the `ODK Javarosa <https://github.com/opendatakit/javarosa>`_ library to render forms. Form transfer between ODK components is governed by the :doc:`openrosa` API.
+
+Because of the complexity of the XForms format, we do not recommend "hand-coding" XForms. The recommended process is:
+
+- Begin with one of the :ref:`compliant-build-tools`.
+- :ref:`Edit the XML only if neccessary <>`.
+- Use :doc:`validate` to check that form is well-formed and fully compliant.
+
+
+.. _compliant-build-tools:
+
+JavaRosa-compliant build tools
+---------------------------------
+
+Most ODK users design their forms in Excel, following the `XLSForm <http://xlsform.org/>`_ specification. To convert XLSForms to XForms, you can use:
+
+- `ODK's online XLSForm converter <http://opendatakit.org/xiframe/>`_
+- `XLSForm Offline for Mac or Windows <https://gumroad.com/l/xlsform-offline>`_
+- - `XLSForm for Windows <https://opendatakit.org/downloads/download-info/xlsform-for-windows/>`_
+- `Pyxform, a Python XLSForm conversion library <https://github.com/uw-ictd/pyxform>`_
+
+For simple forms, `ODK Build <https://opendatakit.org/use/build/>`_ is an online drag-and-drop form designer.
+
+.. _other-xform-build-tools:
+
+Other XForm build tools
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+These tools product XForms that *may not* strictly comply with the JavaRosa library. 
+
+- `Kobo <http://www.kobotoolbox.org/>`_
+- `Enketo <https://enketo.org/>`_
+- `purcforms <https://code.google.com/archive/p/purcforms/>`_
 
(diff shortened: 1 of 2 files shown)


form-design-intro.rst
```diff
@@ -0,0 +1,49 @@
+Intro to Forms in ODK
+========================
+
+Forms used in the ODK ecosystem are XML documents following the `ODK XForms specification <https://opendatakit.github.io/xforms-spec/>`_, a subset of the `W3C XForms specification <https://www.w3.org/TR/xforms/>`_. Most ODK tools use the `ODK Javarosa <https://github.com/opendatakit/javarosa>`_ library to render forms. Form transfer between ODK components is governed by the :doc:`openrosa` API.
```
ankita240796: Maybe we should link some [example forms](https://github.com/opendatakit/sample-forms) here to give an initial insight to a new user.
> To get an initial insight about forms in ODK ecosystem, you may see these examples.

form-design-intro.rst
```diff
@@ -0,0 +1,49 @@
...
+
+Because of the complexity of the XForms format, we do not recommend "hand-coding" XForms. The recommended process is:
```
ankita240796: We should avoid the use of quote marks as specified in the [style guide](http://docs.opendatakit.org/style-guide/#quote-marks).

form-design-intro.rst
adammichaelwood: That's an interesting idea, but I wonder if it would make things even more confusing...
I'll think about how that might be done without overwhelming someone.

form-design-intro.rst
adammichaelwood: This is not that.

cf. https://en.wikipedia.org/wiki/Scare_quotes

form-design-intro.rst
ankita240796: We can add a note further which will be linked to a detailed form design section which can be added later.
> These examples are just to provide a basic idea. Details are discussed here.

form-design-intro.rst
```diff
@@ -0,0 +1,49 @@
+Intro to Forms in ODK
```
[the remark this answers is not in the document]
adammichaelwood: What about  just `Forms in ODK`?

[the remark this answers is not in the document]
troublemagnet: I guess at this point since there are more things that are left to be added, it's a good idea to keep it `Introduction to Forms in ODK`.

form-design-intro.rst
lognaturel: How about the simpler "writing XForms directly by hand"?

lognaturel: (It may not be clear what "hand-coding" refers to)

form-design-intro.rst
```diff
@@ -0,0 +1,49 @@
...
+
+JavaRosa-compliant build tools
+---------------------------------
+
+Most ODK users design their forms in Excel, following the `XLSForm <http://xlsform.org/>`_ specification. To convert XLSForms to XForms, you can use:
+
+- `ODK's online XLSForm converter <http://opendatakit.org/xiframe/>`_
+- `XLSForm Offline for Mac or Windows <https://gumroad.com/l/xlsform-offline>`_
+- `XLSForm for Windows <https://opendatakit.org/downloads/download-info/xlsform-for-windows/>`_
+- :doc:`pyxform`, a Python XLSForm converter with a command-line tool
```
lognaturel: This should link somewhere. I think there's a section or page in the docs somewhere already.

form-design-intro.rst
```diff
@@ -0,0 +1,49 @@
...
+
+.. _other-xform-build-tools:
+
+Other XForm build tools
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+These tools produce XForms that *may not* strictly comply with the JavaRosa library. 
+
+- `Kobo <http://www.kobotoolbox.org/>`_
+- `Enketo <https://enketo.org/>`_
```
lognaturel: Enketo isn't a form builder, it's a form renderer so it shouldn't be here. Purcforms doesn't exist as an independent project (it's part of CommCare).

I'm not sure this section is necessary. As far as I know, the Kobo form builder is built on top of pyxform so it should be compliant. It can either be omitted because it's not a core tool or included in the Non-XLSForm section.

form-design-intro.rst
adammichaelwood: `:doc:` is a Sphinx document reference -- a link to that page.

form-design-intro.rst
adammichaelwood: I'll cut the section. I retained it because it was in older docs, but I'd rather keep this sort of thing out of the docs, with the possible exception of a single page that points to Third Party tools.

form-design-intro.rst
adammichaelwood: right.

form-design-intro.rst
lognaturel: Hehe, 🎶 going too quickly 🎶 . Sorry 'bout that.

adammichaelwood:
@lognaturel 
I thought I made the requested changes already, and was waiting for approval.
That was... not accurate. 😬

This should be good to go now.

PR #340, opened 2017-10-31
Added script to convert the color profile

### LINKED ISSUE #330: Deal with color profile problem locally before PR
We recently had a problem (#323) caused by an image having color profile that was incompatible with with [pngquant](https://pngquant.org/) (a lossy image compression tool we use on deployment).

There was no way for the contributor to know that this would be a problem. It appears that *some* Android devices use the problematic color profile. (Who knows which ones?) 

The solution -- which is easy -- is for screenshots with the bad color profile to be converted.

But if you don't run pngquant locally, there's no way to know this is going to happen. (Unless it has happened to you.)

## Possible solutions

 1. require contributors to install `pngquant` and add it to `b.sh` to catch errors
 2. add [pillow](https://python-pillow.org/) to `requirements.txt` and include color profile checking/conversion in `ss.py` (along with instructions for running it independently)
 3. add color profile conversion via pillow to circleCI and not make contributors think about it

I think I like the second option. I'm also good with combining 1 and 2.

The last option is definitely an option, but I think is probably the least good one.

Doing nothing is also an option, and should not be overlooked.

troublemagnet:
closes #330 

I've added a function that checks and converts the color profile of bad pngs. 
I've added a temp.py for testing purposes and jumpscreen.png(a bad png) in img/collect-settings. The script can be tested on this bad png(it successfully converts the color profile) and on any other png.
I'll remove the files once the script has been finalized, however it's ready for review now.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,4 +13,4 @@ Sphinx==1.6.1
 sphinx-rtd-theme==0.2.4
 sphinxcontrib-websupport==1.0.1
 typing==3.6.1
-update==0.4.4
+Pillow==4.3.0
--- a/ss.py
+++ b/ss.py
@@ -1,9 +1,11 @@
 #!/usr/bin/env python
 
-
 import os
 import subprocess
 import sys
+import PIL.Image
+import PIL.ImageCms
+import tempfile
 
 def ensure_dir(file_path):
     directory = os.path.dirname(file_path)
@@ -17,11 +19,22 @@ def save_screencap(cap, filename):
     with open(filename, 'wb') as f:
         f.write(cap)
 
+def convert_profile(filename):
+    img = PIL.Image.open(filename)
+    icc = tempfile.mkstemp(suffix='.icc')[1]
+    if 'icc_profile' in img.info:
+        with open(icc, 'wb') as f:
+            f.write(img.info['icc_profile'])
+        srgb = PIL.ImageCms.createProfile('sRGB')
+        img = PIL.ImageCms.profileToProfile(img, icc, srgb)
+    img.save(filename)
+
 if __name__ == "__main__":
     short_filename = sys.argv[1]
     filename = 'img/' + short_filename + '.png'
     ensure_dir(filename)
     save_screencap(screencap(), filename)
+    convert_profile(filename)
     if sys.argv[2] != '-s':
         print("Insert image into doc with:")
         print(".. image:: /img/" + short_filename + ".*")


adammichaelwood:
The code added looks reasonable, so I'm not sure where the problem is exactly.

I am getting an error (two, actually) when I run:

```
python temp.py collect-settings/jumpscreen
```

```
Traceback (most recent call last):
  File "/Users/adamwood/odk/odkenv/lib/python3.5/site-packages/PIL/ImageCms.py", line 341, in profile\
ToProfile
    inputProfile = ImageCmsProfile(inputProfile)
  File "/Users/adamwood/odk/odkenv/lib/python3.5/site-packages/PIL/ImageCms.py", line 162, in __init_\
_
    self._set(core.profile_open(profile), profile)
OSError: cannot open profile file

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "temp.py", line 15, in <module>
    img = PIL.ImageCms.profileToProfile(img, icc, srgb)
  File "/Users/adamwood/odk/odkenv/lib/python3.5/site-packages/PIL/ImageCms.py", line 354, in profile\
ToProfile
    raise PyCMSError(v)
```


Notes:

 - I did run `pip install -r requirements.txt`, so I have Pillow.
 - the first time I ran it, I got no error -- but the image was not converted (confirmed by opening in GIMP and checking color profile)
 - I got the errors above after adding a one line `print()` statement (just to see if the `if` clause responsible for conversion was running... it is.)
 - unrelated to the above, but... Python style is indenting on four spaces - no tabs. I got some errors from that, too.

troublemagnet:
Sir can you confirm if Pillow's version is 4.3.0 because when I was writing the script while using lower version I kept getting a similar error?

adammichaelwood:
Pillow 4.3.0 confirmed

I'll try again with the changes you made.

troublemagnet:
@adammichaelwood also requirements.txt needs to be updated. The installation of `update` version==0.4.4 gives an error.

adammichaelwood:
can you make sure requirement.txt on this PR is correct?

adammichaelwood:
I am now having a problem... I cannot get the build to FAIL locally prior to conversion.
If I can't duplicate a failing scenario, it's hard to know if this solves the problem.

I have added the bad image to a doc and will submit a DO NOT MERGE PR to see if I can get it to fail in circle.ci...

Divya063:
@troublemagnet Have you checked with Adobe RGB color profile?There was an issue regarding this https://github.com/python-pillow/Pillow/issues/1662, the script looks same, so maybe the script can be tested on the images having this color profile to see if it is working or if some modifications are required.

troublemagnet:
@Divya063 The script seems to work fine with me. It does perform the transformation from the odd color profile to built in sRGB!

adammichaelwood:
Can you please add a commit that includes the "bad" image in a document, so that we can test failure? I can't replicate a failing condition locally. I'm also having some trouble with missing refs, which is making it hard for me to push a copy of this branch.

troublemagnet:
@adammichaelwood Build failed

adammichaelwood:
okay... let's move this forward

Can you clean up the branch and commit a final version that just has the changes to `ss.py` and `requirements.txt`?

adammichaelwood:
build passes. Can you update this or issue a new PR with just the needed changes?

troublemagnet:
@adammichaelwood  done sir.

PR #354, opened 2017-11-14
reference tomcat 8 (not 6) in aws install

danbjoseph:
addresses opendatakit/opendatakit#1360

The current AMI version is: Amazon Linux AMI 2017.09.1 (HVM), SSD Volume Type - ami-32d8124a and no longer includes the tomcat6 package currently referenced in the docs.

Also fixes where the command for **Install MySQL** was noted as `sudo yum install tomcat6` and not `sudo yum install mysql mysql-server`.


lognaturel:
💫 Thanks so much, @danbjoseph!

Issue #356, opened 2017-11-15
Reorganize Collect pages


adammichaelwood:
I'm currently working on a general re-org. (See: https://forum.opendatakit.org/t/re-organizing-the-docs/10527), but that is too much for a single issue or PR.

---------

This issue is for the Collect section.

My current plan is:

 - top-level TOC (per @yanokwa suggestion):
     - Intro  -- the current `collect-guide.rst` with some expansion, renamed collect-intro.rst
     - Install and Setup -- the current `collect-install.rst` 
     - Using Collect -- the current `collect-forms.rst`, with an added TOC for the other pages:
          - settings
          - adb
          - external apps
      - FAQ / Best Practices -- will combine and re-org projecting, best practices, and some new material 
      - Limitations / Problems / Known Issues (this page needs a good name)

PR #357, opened 2017-11-15
WIP:  re-org of collect docs -- combine pages option

adammichaelwood:
addresses #356 

Big Question:

**Should the several pages of collect user-guide stuff be combined onto the `collect-using` page, creating one mega page of user guide?** 

This PR is an example of what it would look like if that was done.

I like the way this organizes the information if you are looking for something from the TOC. I'm not sure how I feel about the multitude of subjects covered on individual pages.


adammichaelwood:
@yanokwa @lognaturel 
This is ready for an in-progress view... or you can wait for the next PR to see another option and compare.

adammichaelwood:
After doing #358, I think it's clear that this PR is not the right one to move forward with.
I'll be closing it unless anyone objects.

PR #363, opened 2017-11-29
Add CircleCI 2 support

yanokwa:
I tested this by running commands on my branch and running the commands individually. There is a risk that this can't deploy to master, but the only way to test is to merge and try.


adammichaelwood:
approved, but i'll let you merge

.circleci/config.yml
```diff
@@ -0,0 +1,67 @@
+version: 2
+jobs:
+  build:
+    working_directory: ~/code
+    docker:
+      - image: circleci/python:latest
+    steps:
+      - checkout
+      - run: 
+          name: Install checkout requirements
```
lognaturel: Indentation is inconsistent.

.circleci/config.yml
```diff
@@ -0,0 +1,67 @@
...
+    working_directory: ~/code
+    docker:
+      - image: circleci/ruby:latest
+    steps:
+      - attach_workspace:
+          at: ~/code
+      - run:
+          name: Install deploy requirements
+          command: |
+              if [[ "$CIRCLE_PROJECT_USERNAME" == "opendatakit" && "$CIRCLE_BRANCH" == "master" ]]; then \
```
lognaturel: It's not immediately clear why this test is needed given the workflow filter.

PR #370, opened 2017-12-06
Added video directive

### LINKED ISSUE #66: Video sphinx directive

ankita240796:
closes #66 

- Added video directive for sphinx
- Modified form-widgets.rst to use this directive

I am not sure whether the options like width, height, etc are required with this directive. I will add the options as required.

There are very few tutorials for creating a sphinx directive and they don't really clear the basics. I referred some of the tutorials and other extensions to create the video directive.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 build/
+__pycache__/
 .DS_Store
 \#*.*\#
 .\#*
--- a/conf.py
+++ b/conf.py
@@ -17,10 +17,10 @@
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 #
-# import os
-# import sys
-# sys.path.insert(0, os.path.abspath('.'))
-
+import os
+import sys
+sys.path.insert(0, os.path.abspath('.'))
+import video
 
 # -- General configuration ------------------------------------------------
 
@@ -36,7 +36,8 @@ extensions = ['sphinx.ext.autodoc',
     'sphinx.ext.intersphinx',
     'sphinx.ext.todo',
     'sphinx.ext.coverage',
-    'sphinx.ext.mathjax']
+    'sphinx.ext.mathjax',
+    'video']
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ['_templates']
@@ -115,8 +116,8 @@ html_favicon = "_static/img/odk-favicon.ico"
 html_static_path = ['_static']
 
 # Add paths that contain extra files which are not directly related to the 
-# documentation and which are are copied to the output directory.
-html_extra_path = ['vid']
+# documentation and which are copied to the output directory.
+# html_extra_path = []
 
 # -- Options for HTMLHelp output ------------------------------------------
 
--- a/form-widgets.rst
+++ b/form-widgets.rst
@@ -15,7 +15,6 @@ Basic Form Widgets
 
 This section shows examples of all the form widgets types, with no additional options displayed.
 
-
 .. _string-input:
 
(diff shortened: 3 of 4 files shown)


adammichaelwood:
This is really great. 

I think the alt content could be more robust, which would help for print and also web accessibility (not everyone can view video online).

As you know, HTML video has the following structure:

```
<video src="path/to/video" attr="value" other_attr="other_value">
  This content here functions as alt. It is displayed only if the video is not displayed.
  And... this content could be anything... even a picture.
  <img src="path/to/img">
</video>
```

And we might want to be able to provide long text and images as alternative to video. So, I'm wondering if you use the content node of the video directive, which could then be interpreted as restructured text:

```
.. video:: path/to/video

  This content here functions as alt. It is displayed only if the video is not displayed.
  And... this content could be anything... even a picture.
  
  .. image:: path/to/image
``

adammichaelwood:
Oh, also -- 
The video directive should support the `:class:` attribute.

adammichaelwood:
> The alt content is displayed only if the browser does not support HTML5 video element. It is not displayed if the video cannot be played due to a slow connection or the file is missing.

Maybe the `poster` attribute? This could at least provide a placeholder image.
Anyway I think we can live with that for this iteration.

Thinking about the `poster` attribute reminds me that you should also make sure the directive handles the other html attributes we might use: controls, muted, etc. They can have sensible defaults (yes controls, no autoplay, yes muted, no loop) but we should ideally be able to specify that stuff.

ankita240796:
Yes, using the poster attribute is a good solution. For including any alt text, I tried adding some javascript but it made the things unnecessarily complicated and did not produce a proper output. So, it is better to use the poster attribute and just include an image. 

I will update the directive to support the html attributes.

ankita240796:
@adammichaelwood I added support for html attributes as well as a robust alt content.

As of now the image in the poster attribute is copied over to images directory on build. I was not sure about creating a new poster directory for this. 
Also If poster attribute is not included, the first frame of the video will be used instead. So, I am not sure about using a poster attribute. What do you think?

adammichaelwood:
> As of now the image in the poster attribute is copied over to images directory on build. 
Where would it go in the source currently? 

I think it would be fine to put poster images in the video directory, with some convention like `[same-file-name-as-video]-poster.jpg`

ankita240796:
We can put them anywhere in the source. It gets copied from the source address to the images directory on build. 

Yes, that is a nice idea. I will change the code to copy over the images to the video directory. We can document the naming for the poster in the contributors guide later.

adammichaelwood:
I meant they should go in the vid directory in the source. 
It doesn't really matter to me where they go in the output -- i guess the vid directory makes sense. I don't think it matters.

ankita240796:
Oh alright, I got that. :sweat_smile: 
Yes, we can keep them in the vid directory in the source. I changed the location to videos on build as well as.

adammichaelwood:
This looks really good, and seems to me to be ready to merge. Is there anything else you want to do with it at this point?

I tested adding different options, a poster image, "alt" content -- seems to work well.

It would be a good idea, at some point, to pull this out as a standalone Sphinx plugin so that other projects can take advantage of it.

ankita240796:
No, there is nothing else to do. We need to use this directive in some other docs as well as provide alt content for all videos but that can be done in a separate PR. As of now, this is ready to merge!

Yes, it would be a good idea to pull this out as a standalone Sphinx plugin. I would need your guidance to proceed with that.

PR #374, opened 2017-12-07
Added form update guide

### LINKED ISSUE #369: Document form update possibilities and limitations
https://forum.opendatakit.org/t/updating-form-question-without-losing-previous-data/10851/3 is a start.

ankita240796:
closes #369 

Added form update guide which describes:
- Cases where version update is needed
- Cases where a new form is needed


lognaturel:
Fantastic, thanks for taking this one on, @ankita240796! One thing to add to the list of allowed changes is changing the order of questions. It's also possible to add translations. I initially thought that was implied by changing translations but it might be better to make it explicit. It's also possible to update media like CSVs or images.

It would also be good to mention the behavior on the Collect side. Collect will treat different versions of the same form completely independently. It won't notify the users of the existence of new versions. That behavior is about to change thanks to some features @grzesiek2010 is working on but we should start with documenting the current behavior.

@grzesiek2010 anything else you'd like to see added?

Including full form examples is a new thing. I'm not sure how I feel about it! On one hand, it is nice to be able to try things out right away and your examples are cool! On the other it's one more thing to maintain and manage. Targeted, inline examples usually feel like enough to me.

It would be good for you and @adammichaelwood to discuss options for where this document should go.

ankita240796:
Thank you @lognaturel!
I added the information about the order of questions, translations, updating media and Collect behavior with updated forms.

I feel full form examples are good for clear understanding. I added them to form-interaction as well. Since they are not really large examples, we can maintain them with little or no issue. What do you think?

form-update.rst
```diff
@@ -0,0 +1,140 @@
...
+
+.. note::
+
+  ODK Collect treats different versions of the same form completely independently. It won't explicitly notify the users of the existence of new versions. When a user tries to get new blank forms, a form with updated version will be selected by default in the list but there will be no explicit notification unless a user tries to get new blank forms.
+
+  .. image:: /img/form-update/get-new-version.png
+   :alt: Image showing new version 2017120708 selected in the list of forms to be downloaded.
+   :class: device-screen-vertical
+
+  Both versions of the form exist in the device of the user and the user will be allowed to fill a older version and submit the form to the Aggregate server.
```
grzesiek2010: an older*

grzesiek2010: an older*

grzesiek2010: an older*

form-update.rst
```diff
@@ -0,0 +1,140 @@
...
+.. _change-existing:
+
+Making changes to existing form
+---------------------------------
+
+Certain changes which don't involve adding or removing a question can be made without needing to replace the existing forms. The kind of changes are:
+
+- Changing the text or translation of a label
+- Changing validations, calculations, relevants
+- Changing options for a select_one or select_multiple
```
grzesiek2010: Is that about external csv files too or only about options that we store in a form? Because we can edit csv files too like media files.

grzesiek2010:
@ankita240796 @lognaturel maybe it would be good o mention that we don't need to use (or update) version number if we need to change only media files or csv files. I mean if your changes are not in a form you don't need to change the version number.

form-update.rst
ankita240796: I checked this and we can update the options in external csv files too but in that case, we don't need to update the version number.

ankita240796:
Thanks for the review @grzesiek2010! I made the requested changes.

form-update.rst
```diff
@@ -0,0 +1,145 @@
...
+- Changing the order of questions
+- Adding translations
+- Update media for your form
```
lognaturel: To match the rest in the list this should start with "Updating" rather than "Update". I think you could make this something like "Updating media including CSVs"

form-update.rst
```diff
@@ -0,0 +1,145 @@
...
+
+For such changes you can update the version and re-upload the form. 
+
+.. note::
+
+  - If you update media for your form, you don't need to change the version.
```
lognaturel: "only update"?

ankita240796: I made this statement clear in the recent commit.

adammichaelwood:
> adding full form examples 

#386 will cover that when we get to it, and this is informative now, so I think it's good to merge

> and on where the page should go

The form building TOC is good for now.

Issue #380, opened 2017-12-09
Pull in docs from ODK wiki


ankita240796:
- [ ] Collect
     - [Collect troubleshooting](https://github.com/opendatakit/opendatakit/wiki/Collect-Troubleshooting)

- [ ] Aggregate
     - [Aggregate 0.9.x to 1.x](https://github.com/opendatakit/opendatakit/wiki/Aggregate-0.9.x-to-1.x-Migration): Added in Aggregate upgrade guide
     - [AppEngine Master Slave migration](https://github.com/opendatakit/opendatakit/wiki/Aggregate-AppEngine-Master-Slave-Migration)
     - [AppEngine troubleshooting](https://github.com/opendatakit/opendatakit/wiki/Aggregate-AppEngine-Troubleshooting)
     - [AWS Install](https://github.com/opendatakit/opendatakit/wiki/Aggregate-AWS-Install): Some information added in Aggregate install guide. Issue #355 
     - [Aggregate Publishers details](https://github.com/opendatakit/opendatakit/wiki/Aggregate-Publishers-Implementation-Details): can be added in [Aggregate use guide](http://docs.opendatakit.org/aggregate-use/#publishing)
     - [Security note](https://github.com/opendatakit/opendatakit/wiki/Aggregate-Security-Note)
     - [Tomcat troubleshoot](https://github.com/opendatakit/opendatakit/wiki/Aggregate-Tomcat-Troubleshooting)
     - [Uninstall instructions](https://github.com/opendatakit/opendatakit/wiki/Aggregate-Uninstall-Instructions)

- [ ] Contributors guide
     - [Repository migration](https://github.com/opendatakit/opendatakit/wiki/New-Repository-Migration-Guide)

- [ ] Developing with ODK
     - [Aggregate Eclipse Setup](https://github.com/opendatakit/opendatakit/wiki/Aggregate-Eclipse-Setup)
     - [Aggregate database structure](https://github.com/opendatakit/opendatakit/wiki/Aggregate-Database-Structure)
     - [Aggregate deployment configuration](https://github.com/opendatakit/opendatakit/wiki/Aggregate-Deployment-Configuration)
     - [Aggregate source code overview](https://github.com/opendatakit/opendatakit/wiki/Aggregate-Source-Code-Overview)
     - [Briefcase Aggregate API](https://github.com/opendatakit/opendatakit/wiki/Briefcase-Aggregate-API): Issue #375 
[...]

PR #382, opened 2017-12-11
Added Briefcase API doc

### LINKED ISSUE #375: Move Briefcase API docs
https://github.com/opendatakit/opendatakit/wiki/Briefcase-Aggregate-API

It should go to the DEVELOPING WITH ODK section.

ankita240796:
closes #375 
addresses #380 

Pulled in Briefcase API doc.

### DIFF
--- a/briefcase-api.rst
+++ b/briefcase-api.rst
@@ -1,0 +1,262 @@
+Briefcase Aggregate API
+=========================
+
+.. _introduction:
+
+Introduction
+----------------
+
+This describes the interfaces ODK Briefcase uses when interacting with ODK Aggregate v1.0.
+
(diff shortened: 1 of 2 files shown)


lognaturel:
Content and formatting look good. 👍

I think this should perhaps be under one more level of indirection from the main TOC. I'm thinking there could be a section called either APIs or non-OpenROSA APIs. Other things that would go in this category are some of the publisher APIs: https://github.com/opendatakit/opendatakit/wiki/Aggregate-Publishers-Implementation-Details. @adammichaelwood?

adammichaelwood:
I think @ln's suggestion makes sense, but not until some of those other ones are present. So I would put this page in the TOC as this PR has it, and then once we have at least one more documented, we can pull them into a secondary TOC labeled *Other APIs*.

lognaturel:
Makes sense. And as long as links aren't broken, things can certainly move around in TOCs and on the page!

PR #395, opened 2017-12-17
Added info on external settings in Collect

### LINKED ISSUE #5: Describe how loading external settings works
opendatakit/collect#456

ankita240796:
closes #5 

Added info on saving and loading external settings

### DIFF
--- a/collect-settings.rst
+++ b/collect-settings.rst
@@ -214,7 +214,11 @@ You can go to :menuselection:`⋮ --> Save settings to disk` to save the QR code
 
 .. note::
 
-  Settings are saved to :file:`/sdcard/odk/settings/collect.settings` and loaded from :file:`/sdcard/odk/collect.settings`. This is because we don't want to load and delete the saved settings on next app launch. When settings are saved, they are not necessarily saved for the same device but for some other device or mass configuration. So, saving and loading settings is done from separate locations.
+  Settings are exported and saved so that they can be loaded to other devices. When you save external settings, they are saved to :file:`/sdcard/odk/settings/collect.settings`. You can load external settings from this location to other devices as well.
+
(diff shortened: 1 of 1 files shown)


collect-settings.rst
```diff
@@ -210,7 +210,11 @@ You can click on the |share| icon to share the QR code as an image. When you cli
 Saving QR code locally
 """"""""""""""""""""""""
 
-You can go to :menuselection:`⋮ --> Save settings to disk` to save the QR code.  
+You can go to :menuselection:`⋮ --> Save settings to disk` to save the QR code.
+
+.. note::
+
+  Settings are saved to :file:`/sdcard/odk/settings/collect.settings` and loaded from :file:`/sdcard/odk/collect.settings`. This is because we don't want to load and delete the saved settings on next app launch. When settings are saved, they are not necessarily saved for the same device but for some other device or mass configuration. So, saving and loading settings is done from separate locations.
```
adammichaelwood: The explanation here has a weird voice. 

 - Who is "we"?
 - on next app launch -> the next time the app launches
 - "When settings are saved, they are not necessarily saved for the same device but for some other device or mass configuration." -- it's hard to know what is meant by this.

I think you are trying to say:
 - Here is where they are saved and loaded from
 - The reason is to avoid deleting saved settings.
 - Which makes more sense if you have some context for why settings are saved.
 - And so, I repeat, they are saved and loaded form here and here.

But...

If all of that makes more sense with the context of *why* one would save some settings, then start with. -- In fact, start with that *before* you say how to do it.

Briefly:

 - Settings can be *exported and saved*
 - This is done so that they can be loaded onto other devices
 - You do it like this ....
 - Note: At launch, settings are loaded from this other location, which tracks the state of your settings at all times.

ankita240796: Thank you for such a detailed review. This really helps me to learn a lot :smile:

Issue #213, opened 2017-09-26
Describe when calculates, constraints and relevancy are recomputed


lognaturel:
This will require a fairly deep understanding of Collect and JavaRosa and/or experimentation.

lognaturel:
I think thinks will go in two places. One narrative for users in a section on form design and one narrative for developers about the JR innards.

For users I'm thinking of info like:
- When a value changes, all dependent nodes are recomputed
- When a form is finalized, all values are recalculated from start to finish
- The `once` function can be used to only evaluate an expression if the node's value is empty (e.g. `once(random())`)

For devs:
- `Triggerable` represents an expression, its context and the nodes that should be updated when the value of the expression changes...
- A directed acyclic graph of triggerables is kept and manages the recomputation processes...

@mdudzinski since you've been spending time in this space recently, it would be great to get an notes you have as comments on this issue.

mdudzinski:
@lognaturel Sure! I will focus on the dev stuff first.

mdudzinski:
@opendatakit-bot claim

lognaturel:
#413 is awaiting a decision on where it will actually go.

adammichaelwood:
This is covered now in Form Logic:

https://docs.opendatakit.org/form-logic/#when-expressions-are-evaluated

sebmercier:
Hello,

This is my first contribution here, from a freshly created GitHub account, so I hope this the right place (and procedure) to post my comment. Please let me know if that's not the case or if I should create a new issue for this.

I'm precisely coming from the link posted by @adammichaelwood ( https://docs.opendatakit.org/form-logic/#when-expressions-are-evaluated ), because I realised that the `now()` function doesn't behave (anymore?) as described there. Used on its own (as in the "datetime_last_saved" example), it is evaluated only once when the survey is opened. I've read several discussions here about this subject (and related ones) over the past few weeks and I'm not sure where the debate is at today. But because the ODK documentation is an important source of information for XLSForm coders such as me, I thought I should point out this problem. (BTW, at the end of this same documentation section, the calculation formula should read `if(${age} = '', '', once(now()))` instead of `if(age = '', '', once(now()))`. I wasted more time than I care to admit because of this, back in December, but really, that's on me. :-) )

[...]

Issue #46, opened 2017-08-02
Add and document the javasphinx java domain


adammichaelwood:
(Also, make sure it even works.)

https://bronto.github.io/javasphinx/

adammichaelwood:
this is only needed if we pull dev docs into this repo. Until then, I'm closing this.

PR #413, opened 2017-12-20
[Work in Progress] Dev documentation for DAG calculations management

mdudzinski:
addresses #213 

The developer documentation describing how calculations are managed in the JavaRosa codebase.

## What new issues will need to be opened because of this PR?
None for now but refactoring issues - like opendatakit/javarosa#232 - may affect this document as it references the JavaRosa codebase a lot.

This doc was intended to be comprehensive description of the DAG magic. This approach has some drawbacks. For example, it may be very sensitive to JavaRosa refactoring around the evaluation logic.
Some parts - like evaluation for the 'complex question' - are not properly described. Yet.

There are some problems with formatting in line code snippets formatted with double backquote char. For long method signatures perhaps another approach should be taken. In some places, a multiline code block is used but that wouldn't work when referencing a method in the middle of a sentence.

It is a first document that describes logic in the code and there are no standards yet. Once they (the standards) are established, this document may need to be re-organized.

### DIFF
--- a/javarosa-dag.rst
+++ b/javarosa-dag.rst
@@ -1,0 +1,304 @@
+Evaluation Logic Architecture
+##########################################
+
+JavaRosa uses a Direct Acyclic Graph (DAG) to manage triggerable XPath expressions and their dependencies.
+Those expressions can be put in the following attributes of the `bind <https://opendatakit.github.io/xforms-spec/#bindings>`_ section 
+in a form:
+
+* ``calculate``
+* ``relevant``
+* ``required``
+* ``readonly``
+
+The ``constraint`` attribute also takes an XPath expression but it is not managed by DAG.
+It is represented by ``Constraint`` class and evaluated before an answer is committed into the data model
+in the ``FormEntryController.answerQuestion(FormIndex index, IAnswerData data, boolean midSurvey)`` method.
+
+Classes
+-------
+
+- ``Triggerable`` --- An abstract class represents such expression, its context and targets which should be updated by this expression. It has two implementations:
+
+    * ``Condition`` --- Represents an expression for ``relevant``, ``required`` and ``readonly`` attributes which always evaluates to ``true`` or ``false``.
+    * ``Recalculate``--- Represents an expression for the ``calculate`` attribute and can evaluate to any supported data type.
+- ``QuickTriggerable`` --- A ``Triggerable`` class wrapper which provides a shallow comparison of ``Triggerble`` implementations. Once the DAG structure is built, no new ``Triggerable`` instances are created and therefore there is no need for the deep comparison. For the sake of readability, this document will always reference to ``Triggerable`` class even if ``QuickTriggerable`` is actually used in the code. 
+- ``IDag`` --- An abstract class represents an interface for the DAG management and triggerable evaluation logic.
+- ``LatestDagBase`` --- An abstract implementation of ``IDag`` class which serves as a base for all actual implementation of ``IDag``
+- ``Safe2014DagImpl`` --- The only supported ``IDag`` implementation.
+- ``FormDef`` -- A form representation along with the data model. This class communicates with ``IDag``.
+
+High level description
+-----------------------
+
+Each triggerable expression is evaluated at form load and save. Additionally, if the XPath expression references other node(s),
+it is evaluated if any of the referenced node is updated. Moreover, it may happen that a referenced node also has a triggerable
+expression which references other nodes etc. 
+So triggerables may depend on each other. 
+
+Lets check an example:
+Assume there are nodes ``nodeA``, ``nodeB`` and ``nodeC`` and:
+
+* ``nodeA`` has a ``calculate`` expression ``tA`` which references the ``nodeB`` value
+* ``nodeB`` has a ``calculate`` expression ``tB`` which references the ``nodeC`` value
+* ``nodeC`` has no triggerable expressions
+
+So the dependency is ``tC`` > ``tB`` > ``tA``.
+When a form is loaded or saved, those triggerables should be evaluated in the following order: ``tC``, ``tB``, ``tA``.
+When a value of ``nodeC`` is changed then ``tB`` and ``tA`` must be evaluated. In this order.
+
+Lets change the above so the ``nodeC`` node has a ``relevant`` expression which returns ``true`` if ``nodeA`` has certain value.
+This gives the following dependency ``tA`` > ``tC`` > ``tB`` > ``tA``. Since ``tA`` depends on ``tB`` and ``tB`` depends on ``tC``
+it results in an infinity loop. A paradox which cannot be handled.
+
+Direct Acyclic Graph ensures that triggerables are in the topological order. If there are cycles then DAG cannot be constructed
+and an exception is thrown.
+
+``Triggerable``
+----------------
+
+This class stores the following information:
+
+* The expression to be evaluated.
+* A list of target nodes which should be updated with the expression evaluation result.
+* The context reference.
+* A list of ``Triggerable`` instances - named ``immediateCascades`` - that depend on ``this`` evaluation result and should be evaluated after ``this`` is evaluated.
+
+Note: terms 'triggerable', 'calculation' or 'condition' mean the same - an instance of ``Triggerable`` class - and are used interchangeably in this document.
+
+``FormDef`` and ``IDag`` communication
+---------------------------------------
+
+The ``FormDef`` class talks to ``IDag`` through the following methods:
+
+.. code-block:: java
+
+    // Initialization methods:
+    addTriggerable(Triggerable t);
+    finalizeTriggerables(FormInstance mainInstance, EvaluationContext evalContext);
+
+    // Triggering evaluation methods
+
+    initializeTriggerables(FormInstance mainInstance, EvaluationContext evalContext, TreeReference rootRef, boolean midSurvey);
+    deleteRepeatGroup(FormInstance mainInstance, EvaluationContext evalContext, TreeReference ref, TreeElement parentElement, TreeElement deletedElement);
+    createRepeatGroup(FormInstance mainInstance, EvaluationContext evalContext, TreeReference ref, TreeElement parentElement, TreeElement createdElement);
+    triggerTriggerables(FormInstance mainInstance, EvaluationContext evalContext, TreeReference ref, boolean midSurvey);
+    copyItemsetAnswer(FormInstance mainInstance, EvaluationContext evalContext, TreeReference copyRef, TreeElement copyToElement, boolean midSurvey);
+
+    // Validation
+
+    reportDependencyCycles (XFormParserReporter reporter);
+    shouldTrustPreviouslyCommittedAnswer();
+    validate(FormEntryController formEntryControllerToBeValidated, boolean markCompleted);
+
+    // Serialization methods
+
+    getConditions();
+    getRecalculates();
+
+    // Misc
+
+    getEvalBehavior();
+    getTriggerableForRepeatGroup(TreeReference ref);
+    printTriggerables(String path);
+
+    // Seems to be used only by a FormDef method which isn't used in JavaRosa
+
+    getConditionExpressionForTrueAction(FormInstance mainInstance, TreeElement instanceNode, int action);
+
+Initialization
+-----------------
+
+Collecting
+~~~~~~~~~~
+
+Triggerables are constructed during the `bind <https://opendatakit.github.io/xforms-spec/#bindings>`_ parsing.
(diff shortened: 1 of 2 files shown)


adammichaelwood:
Is this intended for developers of JavaRosa, or developers of third-party apps that incorporate JavaRosa?

mdudzinski:
@adammichaelwood Hmm a good question. I was writing it from a JavaRosa developer perspective and focused on the internal implementation. Third-party apps developers are usually mostly interested in the public interface/API of a library.

@lognaturel What do you think?

lognaturel:
I think this should be mostly for ODK developers. Primarily for folks working directly on JR though I think the high-level information is useful for anyone working in the ecosystem.

adammichaelwood:
> I was writing it from a JavaRosa developer perspective and focused on the internal implementation. 

That's what I thought, but I didn't want to make assumptions.

Currently, it is not our intention for this docs repo to house internal developer docs. Rather, developer docs for each project are housed with the project.

But.... @ln ... Is this of enough general interest to developers of other ODK projects that it should be here?

(And... should we circle back on the question of developer docs?)

lognaturel:
Oh, I thought we said dev docs would be here! That's why we had started http://docs.opendatakit.org/javarosa/ Do you think it expands the scope too much @adammichaelwood? Did you have thoughts on where they would go for each tool if not here? Another docs example I've come across recently that I really like is https://ec.haxx.se/http-multipart.html and they have everything in one place. Where should we continue this higher-level discussion?

adammichaelwood:
> Oh, I thought we said dev docs would be here! That's why we had started http://docs.opendatakit.org/javarosa/ 

I was under the impression that this documented the JavaRosa API for the benefit of third party developers... 

We should probably continue this convo. in the forum.

javarosa-dag.rst
```diff
@@ -0,0 +1,304 @@
...
+in the ``FormEntryController.answerQuestion(FormIndex index, IAnswerData data, boolean midSurvey)`` method.
+
+Classes
+-------
+
+- ``Triggerable`` --- An abstract class represents such expression, its context and targets which should be updated by this expression. It has two implementations:
+
+    * ``Condition`` --- Represents an expression for ``relevant``, ``required`` and ``readonly`` attributes which always evaluates to ``true`` or ``false``.
+    * ``Recalculate``--- Represents an expression for the ``calculate`` attribute and can evaluate to any supported data type.
+- ``QuickTriggerable`` --- A ``Triggerable`` class wrapper which provides a shallow comparison of ``Triggerble`` implementations. Once the DAG structure is built, no new ``Triggerable`` instances are created and therefore there is no need for the deep comparison. For the sake of readability, this document will always reference to ``Triggerable`` class even if ``QuickTriggerable`` is actually used in the code.
```
dcbriccetti: Triggerble

javarosa-dag.rst
```diff
@@ -0,0 +1,304 @@
...
+* ``readonly``
+
+The ``constraint`` attribute also takes an XPath expression but it is not managed by DAG.
+It is represented by ``Constraint`` class and evaluated before an answer is committed into the data model
+in the ``FormEntryController.answerQuestion(FormIndex index, IAnswerData data, boolean midSurvey)`` method.
+
+Classes
+-------
+
+- ``Triggerable`` --- An abstract class represents such expression, its context and targets which should be updated by this expression. It has two implementations:
```
dcbriccetti: Perhaps there’s something like HTML’s <dl> that would format this section better than this approach with the three hyphens. Also, when I see multiple hyphens, I wonder if an em dash (—) would be better. Maybe @adammichaelwood can offer something.

javarosa-dag.rst
```diff
@@ -0,0 +1,304 @@
...
+- ``IDag`` --- An abstract class represents an interface for the DAG management and triggerable evaluation logic.
+- ``LatestDagBase`` --- An abstract implementation of ``IDag`` class which serves as a base for all actual implementation of ``IDag``
+- ``Safe2014DagImpl`` --- The only supported ``IDag`` implementation.
+- ``FormDef`` -- A form representation along with the data model. This class communicates with ``IDag``.
+
+High level description
+-----------------------
+
+Each triggerable expression is evaluated at form load and save. Additionally, if the XPath expression references other node(s),
+it is evaluated if any of the referenced node is updated. Moreover, it may happen that a referenced node also has a triggerable
```
dcbriccetti: >any of the referenced node is

nodes are?

javarosa-dag.rst
```diff
@@ -0,0 +1,304 @@
...
+
+* ``nodeA`` has a ``calculate`` expression ``tA`` which references the ``nodeB`` value
+* ``nodeB`` has a ``calculate`` expression ``tB`` which references the ``nodeC`` value
+* ``nodeC`` has no triggerable expressions
+
+So the dependency is ``tC`` > ``tB`` > ``tA``.
+When a form is loaded or saved, those triggerables should be evaluated in the following order: ``tC``, ``tB``, ``tA``.
+When a value of ``nodeC`` is changed then ``tB`` and ``tA`` must be evaluated. In this order.
+
+Lets change the above so the ``nodeC`` node has a ``relevant`` expression which returns ``true`` if ``nodeA`` has certain value.
```
dcbriccetti: Sorry for all the copy editing here. Perhaps I’ll also have something substantive to add.

>has certain

has a certain

javarosa-dag.rst
```diff
@@ -0,0 +1,304 @@
...
+This gives the following dependency ``tA`` > ``tC`` > ``tB`` > ``tA``. Since ``tA`` depends on ``tB`` and ``tB`` depends on ``tC``
+it results in an infinity loop. A paradox which cannot be handled.
```
dcbriccetti: infinite loop

javarosa-dag.rst
```diff
@@ -0,0 +1,304 @@
...
+----------------
+
+This class stores the following information:
+
+* The expression to be evaluated.
+* A list of target nodes which should be updated with the expression evaluation result.
+* The context reference.
+* A list of ``Triggerable`` instances - named ``immediateCascades`` - that depend on ``this`` evaluation result and should be evaluated after ``this`` is evaluated.
+
+Note: terms 'triggerable', 'calculation' or 'condition' mean the same - an instance of ``Triggerable`` class - and are used interchangeably in this document.
```
dcbriccetti: em dashes are great fun! Tough to make on a typewriter but not today.

mean the same—an instance … class—and are

javarosa-dag.rst
```diff
@@ -0,0 +1,304 @@
...
+* Both have the same triggers. This is implied by the second bullet as triggers are the nodes referenced in the expressions.
+
+If such triggerable already exists, ``IDag`` does not add a new instance but updates the existing``Triggerable`` one's context so it points to the 
+highest common root. For the above example, the highest common root for ``nodeA`` and ``nodeB`` calculations is ``/data``.
+The passed instance ``t`` is ignored and a previously added ``Triggerable`` one is returned and should be used for the further processing.
+The context is used later when constructing a DAG structure.
+
+If such triggerable does not exist, it is added to the ``unorderedTriggerables`` list and the ``triggerIndex`` map.
+The passed triggerable ``t`` instance is returned.
+The list is later used to construct the topoligically sorted list ``triggerablesDAG``.
```
dcbriccetti: > topoligically

I won’t flag any more spelling errors that should be caught by a spelling checker.

javarosa-dag.rst
```diff
@@ -0,0 +1,304 @@
...
+.. ``Triggerable.addTarget(TreeReference target)`` because  ``IDag.addTriggerable(Triggerable t)` may return a previously added
+.. instance and there fore ``t`` is ignored.
+
+The ``targets`` list population is postponed and takes place during the form instance parsing, 
+in the ``FormInstanceParser.applyInstanceProperties(FormInstance instance)`` method.
+
+Finalization and constructing the DAG
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Once all ``bind`` nodes are parsed and all ``Triggerable`` collected, the ``IDag`` can be finalized and
```
dcbriccetti: and all ``Triggerable``s are collected

javarosa-dag.rst
```diff
@@ -0,0 +1,304 @@
...
+
+Triggering
+----------
+
+Evaluation of triggerables is triggered when:
+
+* A form is loaded
+* A value of a node changes (an answer is committed)
+* A repeat group is added
+* A repeat group is deleted
```
dcbriccetti: A repeat group is added or deleted

What do you think? I personally have a mild preference for the refactored form.

javarosa-dag.rst
```diff
@@ -0,0 +1,304 @@
...
+
+When this method is called after the form is loaded,
+the root reference is passed and therefore all triggerables in the form should be evaluated. 
+
+A value of a node changes
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When an answer is commited, 
+the ``triggerTriggerables(FormInstance mainInstance, EvaluationContext evalContext, TreeReference ref, boolean midSurvey);``.
+method is called. It finds triggerables that are triggered by the ``ref`` in the ``triggerIndex`` map.
```
dcbriccetti: I think there’s an extra period.

javarosa-dag.rst
adammichaelwood: I agree that a definition list is preferable here. BUT...

These are Java classes, yes? Sphinx supports semantic authoring of software documentation (that's why we use it), so this stuff should ideally be documented using the Java domain. 

cf.:
 - Domains in Sphinx: http://www.sphinx-doc.org/en/stable/domains.html
 - Java domain in Sphinx: http://bronto-javasphinx.readthedocs.io/en/latest/

Java Domain in Sphinx is a plugin which we haven't added yet (see issue #46)
This has not yet been done because I didn't expect to be housing this kind of developer documentation. But... we could prioritize it.


------

Side note... I truly think three hyphens in source should automatically render to em-dash. I might implement that on this repo, because I hate:

 - trying to remember which key combo is em dash, especially since I have remapped some of my keys for emacs
 - trying to guess which type of dash I'm looking at in fixed-width fonts

dcbriccetti: --- mapping to —   : great!
- – and — in fixed-width font: true!

javarosa-dag.rst
mdudzinski: @dcbriccetti Yeah that's right. I should've used a spelling checker. Will fix all of them. Thanks for all the errors you've caught so far!

javarosa-dag.rst
mdudzinski: Won't work. Will require a space between `Triggerable` and 's'

lognaturel: How about “Triggerable objects” or something like that, then?

mdudzinski: @lognaturel Sounds good, will do!

javarosa-dag.rst
mdudzinski: @dcbriccetti I think I fixed all simple spelling errors and apologize I hadn't done it earlier.
I also fixed other errors like missing articles or wrong words (e.g. infinite vs infinity) you've caught but there may be more typos like them.

mdudzinski:
> Excellent work! This is valuable documentation. I wish I’d had this—and more like it—when I first started working on JavaRosa.

@dcbriccetti  I really appreciate these words. If this is helpful then I can finish (in 2018 😄 ) this doc once we establish the standards (dashes, formatting etc.)

adammichaelwood:
So... three dashes in the source *already* makes em dashes, and two dashes become en dashes.
(I discovered this recently...)

As far as styling the class references, that depends on adding the sphinx java domain. 
(See #46)
You could add this as part of this commit, and that would be great. 

See: http://bronto-javasphinx.readthedocs.io/en/latest/

You'll need to:
- `pip install` the extension
- do some configuration (see the docs)
- `pip freeze > requirements.txt` so that we can all update our environments.

Also, some merge conflicts have crept in...

adammichaelwood:
As to the other conversation about whether docs.opendatakit.org is the right place...
Let's just say it is. If it becomes a problem, we can change it later.

dcbriccetti:
>So... three dashes in the source already makes em dashes, and two dashes become en dashes

Adam, surely you aren’t using “dashes” to mean “hyphens”, especially in this context where the distinction is essential. 😀

adammichaelwood:
I did, and...

dont-call-me-shirley

lognaturel:
Most of the content here is now reflected at https://github.com/getodk/javarosa/blob/master/dag.md including some refactor suggestions. I've filed https://github.com/getodk/javarosa/issues/591 for content that is missing.

Phew!

PR #424, opened 2018-01-04
Updated changeref script to add redirects and added info in contribuotrs guide

### LINKED ISSUE #400: Flesh out and document changeref script
While doing the Collect reorg, I wrote a script (`util/changeref.py`) that searches `.rst` files and updates internal references, for use if you have renamed or moved a section or page.

This needs an additional feature: adding a redirect to `s3_website.yml`
And it needs to be documented.

ankita240796:
closes #400 

- Added feature to add redirects
- Added info on changeref script in contributors guide

### DIFF
--- a/util/changeref.py
+++ b/util/changeref.py
@@ -26,8 +26,14 @@ if new_ref_type not in ["ref", "any", "doc"]:
 
 redirect = "  %s: %s\n" %(old_path, new_path)
 
-with open('s3_website.yml', 'a') as file:
-    file.write(redirect)
+with open("s3_website.yml", "r") as in_file:
+    buf = in_file.readlines()
+
+with open("s3_website.yml", "w") as out_file:
+    for line in buf:
+        if line == "redirects:\n":
+            line = line + "%s" %redirect
+        out_file.write(line)
 
 ref_pattern = re.compile(r':((ref)|(doc)|(any)):`([^<`\n]*)<?([^>`\n]*)>?`')
     # group 1 : ref, doc, any (reference type)
--- a/util/changeref.py
+++ b/util/changeref.py
@@ -9,8 +9,10 @@ old_ref_type = sys.argv[1]
 old_ref = sys.argv[2]
 new_ref_type = sys.argv[3]
 new_ref = sys.argv[4]
+old_path = sys.argv[5]
+new_path =sys.argv[6]
 
-if not all((old_ref_type, old_ref, new_ref_type, new_ref)):
+if not all((old_ref_type, old_ref, new_ref_type, new_ref, old_path, new_path)):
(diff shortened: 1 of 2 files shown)


util/changeref.py
```diff
@@ -20,7 +22,12 @@
 
 if new_ref_type not in ["ref", "any", "doc"]:
     print("Third arg not a valid ref type string.")
-    sys.exit()    
+    sys.exit()  
+
+redirect = "  %s: %s\n" %(old_path, new_path)
+
+with open('s3_website.yml', 'a') as file:
```
adammichaelwood: This approach requires that the `s3_website` file never changes structure. What would happen if the redirects listing wasn't the last item in the file?

I *wonder* if we could split out the redirects to their own file. Or some other solution that minimizes this kind of risk.  (My first thought was to parse in the yaml, update it internally, and then write the whole file back out. That seems like overkill, maybe.)

util/changeref.py
```diff
@@ -20,7 +22,12 @@
 
 if new_ref_type not in ["ref", "any", "doc"]:
     print("Third arg not a valid ref type string.")
-    sys.exit()    
+    sys.exit()  
+
+redirect = "  %s: %s\n" %(old_path, new_path)
```
adammichaelwood: We might call this a naive implementation.

It depends on the writer's ability (and willingness) to construct the entire redirect string, which might be from a page to a section, or from a section to a page, and include `#reference-targets` that have a different name than the section labels in the code. 

We just need to think about this a little more.

util/changeref.py
ankita240796: I have fixed this issue. Now the redirects line is searched for and a redirect is added below the redirects line when found. So, ``s3_website`` file can change structure without any issue.

util/changeref.py
ankita240796: I felt the same here but the issue here is how do we construct the redirect string internally and what inputs do we need from the writer. We cannot construct the string without the section labels, doc info. So, we would require all of this as input or the complete redirect string.

adammichaelwood:
> the redirects line is searched for and a redirect is added below the redirects line

Clever. Have you tested it to make sure it behaves as expected?

ankita240796:
Yes @adammichaelwood, I have tested this. It works correctly.

adammichaelwood:
This probably needs to be rethought in light of #433

ankita240796:
@adammichaelwood The instructions are updated in contributors guide but I would need your suggestions on how should we take input from the user to add the redirects.

adammichaelwood:
How we do redirects has changed a bit -- we can't actually do section-to-section redirects, it turns out.
So I'm closing this PR.

I'll update the originating issue with the current requirements. It's low priority, though.

PR #427, opened 2018-01-07
Style Guide Testing

### LINKED ISSUE #94: Style Guide testing
- Find and implement appropriate testing tool (something like https://github.com/amperser/proselint)
- implement and document project-specific style guide (which we can edit with relative ease)
- incorporate into build script (#93)

ankita240796:
closes #94 

This is the first try to incorporate Proselint for style guide testing. I have included all the pre-defined tests and added a simple test spelling-odk.py and added a test script style-guide-test.py. The script can be run as : ``python style-guide-test.py``. I  have also added a new function errors_formatted in tools.py to display the warnings clearly. I will keep on adding tests according to other sections in style guide.


adammichaelwood:
Very exciting.

Is it possible to (like...) `pip install` this so that we don't need to include over 100 proselint files in our source?

ankita240796:
We want to add new checks specific to the style guide. I am not sure if this possible after ``pip install``. I will try exploring this after I am done with adding all the checks.

adammichaelwood:
>We want to add new checks specific to the style guide. I am not sure if this possible after pip install proselint. It is also not a good idea to edit an installed package. 

There's got to be a way to extend it without editing the core. I will look at this a bit and see if I have any suggestions for how to do that...

wonderchook:
@ankita240796 is there a reason you can't create a subclass and then add the additional checks? When you do figure out how to extend it this would be a great blog post, since there doesn't seem to be that much info out there on extending proselint

lognaturel:
I shared some thoughts on Slack at https://opendatakit.slack.com/archives/C5KCRU1L1/p1515703435000068 but for completeness let me also summarize here.

`proselint` hardcodes a couple of things that make extensions difficult:
- Checks have to live in the projects' `check` module: https://github.com/amperser/proselint/blob/master/proselint/tools.py#L127
- `.proselintrc` is only searched for in specific places: https://github.com/amperser/proselint/blob/master/proselint/tools.py#L143

I think a cool thing to do (after reading about proselint contribution) would be to find out if two modifications would be appreciated:
- Provide a setting to define a custom checks directory in `proselintrc`
- Add a command-line argument to specify a particular `proselintrc` file (so that projects like ours can check it into their repos)

If they're not, we could maintain a fork with those two changes. Both are small so it should be easy to keep that in sync with core.

ankita240796:
@adammichaelwood @wonderchook @ln This is something different I thought to do. I added a script which copies over our custom checks to proselint's check directory as well as reads and updates the proselintrc file: https://github.com/opendatakit/docs/pull/427/commits/4508f4a1c258417c0f0d4b8b713eeddbf1cdc5d9
If this looks fine, I will proceed with this approach.

adammichaelwood:
> I added a script which copies over our custom checks to proselint's check directory as well as reads and updates the proselintrc file: 

Fascinating. When does this occur? At testing time?

ankita240796:
Yes, I will work on the "literate" approach once I am done with all the checks!

style-test.py
```diff
@@ -0,0 +1,148 @@
...
+            severity = "severity: %s, " %e[7]
+            replace = "replacements: %s " %e[8]
+
+            # add errors to list
+            err_str = check + msg + line + column + start + end + extent + severity + replace
+            err_list.append(err_str)  
+
+        # display filename
+        print(filename)
+
```
adammichaelwood: I think this would be more user friendly if each error message included the file name and (if possible) the matched text, + a few words on either side of it. (Not the whole line, as that would end up being an entire paragraph.)

I'm thinking something like this:

```
sandwich-making.rst | ln 101 | col 23
...peanut butter, jelly and bread...
Use serial comma after penultimate item.
style-guide.serial-comma | warning
```

(I'm not sure what `start`, `stop`, and `extent` do, but I think they can be left out because everything the user needs to fix the problem is present there.)

Color would also help writers:

```
sandwich-making.rst | ln 101 | col 23      // blue
...peanut butter, jelly and bread...           // default color 
Use serial comma after penultimate item.    // yellow, orange, or red depending on severity
style-guide.serial-comma | warning    //  maybe gray
```

(Note: Feel free to use something else you like better, but -- I've used the [blessings](https://pypi.python.org/pypi/blessings/) for this in the past and found it very easy to work with.)

If the other (or some other) format for output is needed for circleCI, there can be like a `-verbose` flag or something.

ankita240796: File name is displayed before the error list. Should the filename be displayed with each error?
 I will try to add the other features.

ankita240796: start, stop and extent also tell the exact position of the error string counting word by word but that can be skipped safely.

ankita240796: I will work on adding the features.

ankita240796: @adammichaelwood Done!!
screenshot from 2018-01-18 00-42-19
These colors make me happy. This was a great idea indeed!!!

ankita240796:
A description of all what is done and what is still to be done:

## Tasks done
- Almost all the checks are added. The check for section labels and indentation is left. Proselint does not seem to check the quotes and label formats from the regex. I will need to include the complete code for checking instead of calling the inbuilt functions in proselint. It fails to check curly quotes and straight quotes even in the checks which are pre-defined in proselint. Some of the lines and columns are correctly matched and others are ignored. So, I will add complete check instead of using inbuilt functions. Other checks are working fine.
- To run the tests:
   - Navigate to docs and run: ``pip install -r requirements.txt``.
   - To run checks for all files: ``python style-test.py`` 
   - To run tests for specific files: ``python style-test.py -- [file1] [file2] ...``
- I have separated a [list of checks](https://github.com/ankita240796/docs/blob/a2aff9b7f71119d2d6378f57e447c55ab6892381/style-test.py#L107), failure of which will lead to build failure. Apart from these, all other tests cannot be error-free always. So, they are just used as warning. I did not use the severity defined already as it is fixed for the inbuilt checks in proselint. So, preparing a list seemed better.
- Output is generated in the format specified [here](https://github.com/opendatakit/docs/pull/427#discussion_r162126526).
- User can ignore lines to avoid testing on them.
```
.. startignore
Any text
.. endignore
```
The text between the comments ``.. startignore`` and ``.. endignore`` will not be checked.

## Tasks remaining
- Add testing to build script.
- Automatic fixing of errors which are fixable.
- Adding checks for section labels and indentation and fixing checks for quotes.

style-test.py
```diff
@@ -0,0 +1,148 @@
...
+            check = "check: %s, " %e[0] 
+            msg = "message: %s, " %e[1]
+            line = "line: %d, " %(1 + e[2])
+            column = "column: %d, " %(1 + e[3])
+            start = "start: %d, " %(1 + e[4])
+            end = "end: %d, " %(1 + e[5])
+            extent = "extent: %d, " %e[6]
+            severity = "severity: %s, " %e[7]
+            replace = "replacements: %s " %e[8]
+
```
adammichaelwood: In general, I don't think it is a good practice to compose strings in several places.
Here you are pulling unnamed strings out of `e`, and concat-ing then with their name, and then again later concat-ing them together. This requires the next person who has to work on this code to figure out what is going in a few different places. For a simple script like this, on our small team, that's not a huge deal. But in a larger production environment, this is a habit that leads to technical debt.

The other thing is that pulling `e` into named variables (probably a single dict) but without the additional text is that it makes it easier to change what the whole output string looks like. -- (Very relevant, since I'm suggesting you do exactly that in my comment below.)

```
for e in errors:

    check = e[0]
    msg = e[1]
    line = e[2] + 1   # just a matter of convention, this is better than  1 + e[2]

```

This just makes it clearer what you are doing.

Anyway... 
All this error-message code would be cleaner if we upgraded to python 3.6 and used f-strings.

But I don't think we should force that on people. 

I wonder if there is a simple templating package that would replicate the function of f-strings so that you could build the whole error msg in one place:

```
err_msg = """
 {{file_name_}} | ln {{linum}} | col {{colnum}} \n
{{text_excerpt}} \n
{{message}} \n
{{check}} | {{severity}}
"""

(See my other comment for why that format...)

This would make the code more readable and maintainable.

**Comment by ankita240796** [comment]
@adammichaelwood Ready for review!

File: style-guide/check-howto.py [hunk 3]
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+"""How to construct."""
```
adammichaelwood: Is it possible to only check for this in headlines and titles?

style-guide/check-same.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
```
adammichaelwood: There are places where this word is perfectly fine. I wonder if there's a way to narrow it down to likely problems...

At any rate, it doesn't always crop up as `The same`.

style-guide/check-space.py
```diff
@@ -0,0 +1,15 @@
+# -*- coding: utf-8 -*-
```
adammichaelwood: Does this exclude inline literal blocks, which might have dot notation? Like: `object.method()` or `example.com`?

style-guide/extra.py
```diff
@@ -0,0 +1,85 @@
+# -*- coding: utf-8 -*-
```
adammichaelwood: Wondering if this should be broken up into individual files like the others.

(Maybe it doesn't matter if we're going to move to having the source be the style-guide file...)

style-guide/latin-abbr.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
```
adammichaelwood: Probably want to include poorly-punctuated versions of these, such as `eg.`, `ie.` (with and without trailing periods, surrounded by spaces or surrounded by asterisks (italics):

```
Some people, *ie* people who don't read the style guide, might put Latin abbreviations in italics. *Eg* like this.
```

style-guide/order-list.py
```diff
@@ -0,0 +1,14 @@
...
+"""ordered list check."""
+
+from proselint.tools import memoize, existence_check
+
+@memoize
+def check_orderlist(text):
+    """Use unordered list if order of steps does not matter."""
+    err = "style-guide.order-list"
+    msg = "Use unordered list if order of steps does not matter."
+    regex = "([0-9]*\.\s[(\x00-\x09)|(\x0B-\x7F)]*\n)+"
```
adammichaelwood: I'm curious about your regex here.

style-guide/personal-pronoun.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
```
adammichaelwood: If he/him or she/her, the message to the user should recommend use of "they" (as per the style guide) if a personal pronoun is absolutely needed.

style-guide/unorder-list.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
```
adammichaelwood: We should think about if there is some way to intelligently guess whether a list looks like procedural steps or if it looks like a collection of things. I'm not sure... 
Off the top of my head, I would think looking for verbs and verb cases. But that might be too hard to actually do.

adammichaelwood:
Great work so far.

I'd like to remove the corrected content files from the PR, if that isn't too difficult, so we can check in a "clean" new feature.

I think this is usable (or very close to usable) as a utility for writers, and I'd like to get it merged in this state, when you feel like its ready to merge. (We'll need some explanation for how to use it in the contributor guide, but I'm in the middle of a re-organization to those pages, so it can wait another PR cycle.

Then adding it to automated checking can be another PR (unless you did that already here and I missed it), and moving to the literate style can be another PR.

style-guide/extra.py
ankita240796: I actually created one file since all these checks are run directly without proselint. Proselint did not produce correct output with quoted text, so I had to create these separately. I kept them in a single file so that only one file is imported while checking. If we have separate files, we will need to import all of them in the style-test script.

style-guide/check-howto.py
ankita240796: Yes, it is! Fixed this!

style-guide/check-same.py
ankita240796: As of now, I cannot find a way to crop up as ``The same``. Maybe we could avoid ignoring case to crop to some cases or just remove this test? What do you suggest?

style-guide/order-list.py
ankita240796: An ordered list will be of format: ``1. some characters\n``
``[0-9]*`` checks any number of digits from 0-9, then a literal ``.`` and ``\s`` checks a space.
In the character set which is checked then: ``[(\x00-\x09)|(\x0B-\x7F)]*`` checks all characters except a newline. If a newline is also matched in this character set, any pargaraph or statement which is after a list will also be matched. So, I used this character set and then checked a ``\n``.

style-guide/check-same.py
ankita240796: I made some changes to the test. Now the check only matches ``The same`` after a period and space and it is case-sensitive as well.

style-guide/unorder-list.py
ankita240796: I searched for this but I was not able to find a suitable way to do this. Maybe we should remove the checks for lists because as of now we have all lists in the proper format and for new PRs we can manually check the lists. It would be better to this rather than generating a list of warnings for all lists.

ankita240796:
> Great work so far.
Thank you @adammichaelwood.
> I'd like to remove the corrected content files from the PR
I have made a separate line edit PR #442. Once that is merged, I will remove the corrected file content.

ankita240796:
@yanokwa @ln I would need your review on this PR  to make some more improvements!
Also, this PR adds style guide testing to CircleCI Build. That needs a check as well!

ankita240796:
Guidelines on adding checks
- All the checks reside in the folder style-guide in the docs repo.
- Any new check can be added to this folder.
- New check should follow the coding format in the existing checks.
- The pre-defined functions used from proselint are ``memoize`` , ``existence_check``, ``preferred_forms_check``. 
- In a new check, specify ``@memoize`` above the function definition to cache the check.
- The function ``existence check`` has following parameters:
      - text: text to check for
      - list: regex list
      - err: check name
      - msg: message to be displayed when match found
      - ignore_case: True/ False (re.IGNORECASE)
      - str: True/False (re.UNICODE)
      - max_errors: maximum number of errors to be displayed
      - require_padding: True/False 
      - dotall: True/False (re.DOTALL)
      The parameters offest and excluded_topics, join are not required in any case.
      A typical usage would need only the text, list, err and msg parameters as in various other checks. Other parameters are optional. This functions checks for the existence of a regex pattern(s) in the text.
- The function ``preferred_forms_check`` has following parameters:    
      - text: text to check for
      - list: list of comparisons: [correct form , incorrect form]
      - err: check name
      - msg: message to be displayed when match found
      - ignore_case: True/ False (re.IGNORECASE)
      - max_errors:  maximum number of errors to be displayed
      The parameter offset won't be required.
- You can also use the function ``consistency check`` as a consistency checker for the given word_pairs.
   The parameters are:
     - text: text to check for
     -  word_pairs
     - err: check name
     - msg: message to be displayed when match found
    The parameter offset won't be required.
- The code for all these functions can be found here: https://github.com/amperser/proselint/blob/master/proselint/tools.py
- Any check which is to be run without proselint can be defined in the file ``extra.py`` which resides in the style-guide folder. You can define a new function which returns the following format of errors: `` errors = [(err, msg, row, col, start, end, extent, "warning", "None")] ``. To find row and col, use the function ``line_and_column`` defined in the same file. Then call the function in the ``check function`` in the same file to run the check.

ankita240796:
Guidelines on style-guide testing:
- Run the script ``style-test.py`` to execute the tests on all docs.
    - Run ``python style-test.py`` to run check on all files.
    - Run ``python style-test.py [file1] [file2] ...`` to run checks on the specified files. For example, ``python style-test.py oauth2-service.rst`` will perform the tests only on the ``oauth2-service.rst`` file.
- All warnings are displayed in yellow and errors in red. File name, row and column number, check information and error message is displayed as well. A count of all errors and warnings is displayed in end. If errors are found, you need to fix them to avoid build failure. Go through the warnings as well to make the required changes manually.
- To fix the errors automatically, run the script ``style-test.py`` with the fix option.
    - Run ``python style-test.py --fix`` to fix all errors.
    - Run ``python style-test.py --fix [file1] [file2] ...`` to fix errors in specified files. For example, ``python style-test.py --fix oauth2-service.rst`` will fix the errors in ``oauth2-service.rst`` file.
- After fixing the errors, run ``git diff`` to find the changes. There might be case issues which you can fix manually.

ankita240796:
@adammichaelwood I think this is ready for a final review and if everything is fine, it is ready for merge!

adammichaelwood:
This is great. The individual error messages are much easier to read now.
I'm going to merge shortly after some additional review.

Here's a few new issues to file, to improve on this...

### Prune grammar check

One big problem at the moment is when I run it, I see over 1000 warnings. 

Some of these will get fixed, of course. And we might use the `.. ignore` syntax on some of them. But it's still way too many to be helpful.

So we're going to need to figure out how to cut that down to something more reasonable.
I suspect some of these grammar issues we're checking aren't really checkable practically, and we're getting a lot of false positives. I'm looking through what gets returned so we can prune a bit.

### Feedback to user

Another thing we could improve in another iteration is feedback during run time. It takes several minutes to return anything, which will lead to some users assuming it isn't running.

The simplest solution is to start by printing a message like, "Running style guides tests. This will take several minutes..."

A better solution would be to print a message for each file, and output the errors for each file, as they complete.

```

Testing file-name.rst....

xlsform.rst | line: 32 | col: 105
...hey are used. They can be manu...
Avoid using third-person personal pronouns like "he", "she". In case of absolute need, prefer using "they".
style-guide.personal-pronoun | warning


xlsform.rst | line: 32 | col: 174
...troduction>`. Simply place the...
Avoid using unneeded adverbs like "just", "simply".
style-guide.unneed-adverb | warning


Testing another-file.rst....


```

If this comes in a continuous stream, rather than all at once after finishing, it will be a better user experience.

### Output to another format / to file

ideally, one could:

 - from the terminal, output the errors to a file as csv records
 - from another python script, call style-test and get a collection of errors (probably a list of dicts or namedtuples)

(This is probably something more suited toward incorporating back into proselint core than maintaining locally.)


### Testing

We need a unit test for style-test.py. 
As (I think) we've discussed, this would be best incorporated as part of the literate programming approach. (That is, style-guide.rst contains the grammar check definitions and the tests for them, along with prose explanations of what we are testing for.)

adammichaelwood:
## Some issues to resolve before merging

As I mentioned, the sheer volume of warnings makes this less than helpful at the moment.
I've looked through the output, and there's a few things we can fix without too much trouble.


### Pronoun check -- remove they

"they" and its variants (them, their, they're) are often used legitimately, and should be removed.
We should still check for "he" and "she" and their variants.

### overlap between style-guide and existing proselint checks

several style rules and checks are duplicated. Most of these are in "unneeded words" and "Filler phrases"

### list comments are unhelpful

it currently generates a note on every list. This check should be turned off altogether, unless and until we can make it more discerning.

(It is also responding to section-title underlines as if they are unordered lists, which is unhelpful.)

### Ignore inline literals

If possible we should ignore text inside double-backticks.

Also (this might be harder...)
we should ignore text inside one of the rst roles that contains code 
(`:command:`, `:option:`, `:py:`, `:js:`, etc...)

### semantic dots mistaken for periods without spaces

We are getting the spaces-after-periods warning on:

 - filenames
 - URIs
 - version numbers

Some of that would probably be solved by ignoring inline literals and certain rst roles.

### ODK spelling issues

the "an XForm" rule is supposed to enforce a preference for "an XForm" over "a XForm"
But currently it catching when any word other than "an" precedes "XForm"
It should only catch on "a XForm"

(This goes for "an XLSForm" as well.)

### counter examples in style guide creating warnings

I guess we should probably just use the `..ignore` block for these.

### Remove the ellipses warning

the typography checker built in to proselint checks for "..."
This test should be excluded.
(We should examine all the typography rules because there are probably others that don't apply due to our RST syntax.)

ankita240796:
@adammichaelwood  In the recent commits, I have made the following fixes:
- Text between double or single backticks is ignored.
- Removed checks for lists since as of now we don't have a strong test for it.
- Add ignore blocks in ``style-guide.rst``.
- Add a list of checks to be excluded as the typography checks. We can add new checks to the list as needed.
- a/an check added to XLSForm and XForm.
- they, them, their is removed from pronoun check.
- Better UI -- printing error after each file is checked.

adammichaelwood:
i like the way [Pandoc](https://pandoc.org/) handles output formatting. 

They have an `-o` (for output) flag which is followed by a file path.
Then the format is determined by the extension on the path.

So `-o output.csv` would generate a csv output,
while `-o output.txt` would generate something that looks like you have already.
Later on, if we find we want to,
we could add support for `-o output.json` or whatever, etc..

And if no `-o` flag is present, it just does what it does currently.

The only (to some people) unintuitive thing is the output file then is listed in the command before the input files (if they are specified):

```
$ python style-test.py -o output.csv input-file.rst
```

This is only confusing, though, if it isn't well documented.

ankita240796:
@adammichaelwood I have added support for csv output generation. It can be produced by running:
`` python style-test.py -o filename.csv [input files] ``

adammichaelwood:
@ankita240796 this is creating an error

ankita240796:
@adammichaelwood I forgot to specify the required parameter :sweat_smile: . Now, this will work fine!

adammichaelwood:
i love the csv output -- will make it easier to investigate which tests are producing too many false positives, etc..

I'm going to merge this because:
 - it is highly useful right now
 - i want to use it for some editing work I'm currently doing

Can you add planned enhancements we've discussed as individual issues?
(Even ones we might not actually be able to solve -- like OL/UL testing, etc. -- so we have a place to discuss possible approaches.)

ankita240796:
Yes, I will the add planned enhancements as individual issues!

PR #428, opened 2018-01-16
Added script for generating form-widgets guide

Divya063:
This PR includes a python script which generates relevant text descriptions, image alt-texts, and directives by parsing all-widgets sheet and Xform. I have submitted a PR opendatakit/collect#1788 for the setup of screengrab tool. I would add instructions in the docs on how to generate the screenshots. The script can be run as ``python3 automate_screenshots.py``. The output of the script is available[ here](https://drive.google.com/file/d/1fwDloij4N8Tk7fucDG4rAfRWd30lNI-6/view?usp=sharing).


adammichaelwood:
Neat!

automate.py
```diff
@@ -0,0 +1,113 @@
+
```
adammichaelwood: this script needs a more descriptive name than `automate`
Perhaps `automate_widgets_screenshots.py` or just `automate_screenshots.py`

Also, I think we should move this into `util/` directory.
I'd like to move all our utility scripts there.

automate.py
```diff
@@ -0,0 +1,113 @@
...
+
+# source directory - generated screenshots are stored here
+
+dir_src = os.path.expanduser("~/collect/fastlane/metadata/android/en/images/phoneScreenshots/")
+
+# destination directory
+
+dir_dst = os.path.expanduser("~/odk/odkenv/docs/img/form-widgets")
+
```
adammichaelwood: these file paths concern me, because they make assumptions about how the user has their files organized. (This is bad practice generally, BTW. For example -- this isn't how *I* have my files organized.)

The should be relative to the directory the user is in when they run the script, which should should be the root of the odk-docs repo.

automate.py
```diff
@@ -0,0 +1,113 @@
...
+# To remove unwanted numbers appended to screenshots' names
+for file in files:
+    if "_" in file:
+        f=file.split("_")
+        os.rename(os.path.join(path, file), os.path.join(path, f[0]+'.png'))
+
+# path to form-widgets guide
+
+file1 = open("path/odk/odkenv/docs/form-widgets.rst", "w")
```
adammichaelwood: i don't understand this file path either

automate.py
```diff
@@ -0,0 +1,113 @@
...
+
+
```
adammichaelwood: i'm confused about what these text definitions are doing here...

adammichaelwood:
I'm a little confused about the overall logic here. 
Could you provide a bit of a narrative about how this will be used?

Divya063:
@adammichaelwood The current approach overwrites the form-widgets guide and replaces everything with image captions, directives and alt-texts. The problem with this approach is section labels, csv tables, headlines and xml code blocks are getting removed. Another approach can be appending only new widgets to the guide using the script, thus leaving the description of already added widgets as it is, and changing the image file names manually as it would change only once.

adammichaelwood:
A followup to our slack convo:


> The problem with this approach is section labels, csv tables, headlines and xml code blocks are getting removed.

The xlsform, and the xform it generates, should be the source of this. 

 - The alt text and narrative prose should be added as columns to all-widgets (though those columns should not be included in the xlsform example snippets, obviously).

 - I still don't understand why content is in the python file.

 - The script should pull the images off the device and deposit them in the appropriate folders (over-writing if needed)

 - This can be in a further iteration, but --- we should consider how to deal with input. I am fine prompting the script runner to input things. Or we could add further columns on all-wdigets defining input behavior, which would then be consumed by the script.

Divya063:
Thanks, @adammichaelwood for clearing my doubts! The copy of all-widgets sheet which I'm using is available[ here](https://drive.google.com/file/d/1WAqY19nKJgfg2Y8eUm0xRnaoWOD-kAs6/view?usp=sharing). The script, all-widgets sheet, and [xform](https://drive.google.com/file/d/1paqinyFoFrqfo4C6w2OaMoxLDcxbzITx/view?usp=sharing) should be in the same directory while running the script. The generated output is [here](https://drive.google.com/file/d/1fwDloij4N8Tk7fucDG4rAfRWd30lNI-6/view?usp=sharing), some editing is needed in the sheet for adding other widgets.

adammichaelwood:
I discussed this with @yanokwa a bit and I think the path forward on this aspect of the automation is probably too time consuming in relation to the added value at this time. So I'm going to close this PR. The screenshot script, though, will get some use and we should get the instructional PR merged in.

However --- I still think that at some point this is a worthwhile thing to do. The work here will provide some guidance on how to do that, when we get to it.

yanokwa:
Thanks for your work on this, @Divya063! And sorry we couldn't find a way forward.

PR #433, opened 2018-01-17
Remove hash redirects because they are not supported on S3

### LINKED ISSUE #414: Bug: Broken link on Collect page
## Description

The Collect page at https://opendatakit.org/use/collect/ notes that the page is deprecated and suggests that the user visit http://docs.opendatakit.org/collect-guide/. However, that URL gives a 404 error.

Let me know if this is the wrong repository for this issue... Apologies if it's not!

### LINKED ISSUE #425: Redirects not working for sub-sections in a page
When we rename a page and add a redirect, the redirect is not added for the sub-sections in the page. For example, if the page collect-guide is changed to collect-intro, https://docs.opendatakit.org/collect-guide is successfully redirected to https://docs.opendatakit.org/collect-intro/ but https://docs.opendatakit.org/collect-guide/#learn-more-about-odk-collect is not redirected.

screenshot from 2018-01-04 16-36-50

### LINKED ISSUE #426: Redirects not directing to the correct page
Some of the other redirects are not redirecting to the correct page.
https://docs.opendatakit.org/collect-install/#connecting-to-a-server is not redirecting to https://docs.opendatakit.org/collect-connect/
  
screenshot from 2018-01-04 16-38-13

yanokwa:
Fixes #425 
Fixes #426 
Fixes #414

In [my research](https://stackoverflow.com/a/15133106/152938), I have learned that URL fragments (the stuff after the hash/bang) does not get sent to the server. Because of this, S3 (and probably any other service that would rewrite URLs) cannot redirect based on the hash. 

You could do the [redirect with JavaScript](https://stackoverflow.com/a/24377443/152938), but that feels hacky.

I confirmed the above by testing all the various combinations of urls (e.g., `collect-install/#connecting-to-a-server`, `collect-install#connecting-to-a-server`) to see if I could get a redirect working. I could not.

I verified this change by testing on my own S3-based static website and confirming that the url/index.html syntax redirect when I type `url`, `url/`, `url/index.html` and hash passthroughs.

As far as implications for design, I guess we should be really careful about using anchors.

### DIFF
--- a/contributing.rst
+++ b/contributing.rst
@@ -524,14 +524,14 @@ Complete the pull request. The maintainers will review it as quickly as possible
 
    If you happen to rename any document file(:file:`*.rst`), then be sure that you add the redirect in your PR.
 
-   To add the redirect go to :file:`s3_website.yml` and uncomment the **redirects:** line. Add a mapping from the old file name to the new file name below the **redirects:** line, one mapping per line. Several examples of how to format these are shown in the comments.
+   To add the redirect go to :file:`s3_website.yml`. Add a mapping from the old file name to the new file name below the **redirects:** line, one mapping per line. Several examples of how to format these are shown in the file.
 
    For example you rename a file to :file:`newcheck.rst` from :file:`oldcheck.rst`, then to add the redirect:
 
    .. code-block:: yaml
 
      redirects:
-      /oldcheck: /newcheck
+      oldcheck/index.html: /newcheck
 
 
 .. _keep-working-the-docs:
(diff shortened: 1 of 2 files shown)


adammichaelwood:
Thanks for figuring this out.

> As far as implications for design, I guess we should be really careful about using anchors.

eesh.

Sphinx generates title anchors by default, and uses them for reference links.
So we'll always have them.

We can't stop other people from linking to anchors, 
but this means we should probably avoid doing so ourselves,
as it makes future re-orgs break links.

Also -- we should try to avoid future re-orgs.

yanokwa:
@adammichaelwood Agreed. And now that we know what we know, we should be even more careful about shipping changes to master because once those are live, it's hard to remap.

PR #435, opened 2018-01-18
Fixes filename

adammichaelwood:
The file `aggregate-increase-aws-size` is not about aws, but Google App Engine. 
The file name was a mistake, which this PR corrects.

Also, "App Engine" has been added to the title to make this more clear.


lognaturel:
Looks like the source was https://opendatakit.org/use/aggregate/#Changing_Server_Size. It did say "(this also applies to Tomcat deployments but Tomcat server size is configured differently and is not discussed here)"

Issue #398, opened 2017-12-18
Aggregate Reorg


adammichaelwood:
Essentially, the same work I did on #361, for Aggregate

PR #452, opened 2018-01-26
Aggregate fixes

### LINKED ISSUE #440: Add warning about columns in MySql to tomcat install guide
See conversation here: 
https://forum.opendatakit.org/t/request-for-review/11491/5

MySQL has a problem as an Aggregate datastore if the tables have too many columns (that is... if the forms have too many questions). There are some mitigation strategies, but the easiest solution is to just use PostgreSQL instead.

adammichaelwood:
closes #440
addresses #398 

 - Moved the TOCs around a little to better match what we did with collect
 - Updated titles to used gerund verbs wherever appropriate
 - added a mention of the MySQL table row width problem

It might be worthwhile to expand the section I added about choosing a database into a larger document about pros/cons etc. of different RDBMSes. But... it might not be. The information I wrote there covered the problem identified in #440, but it doesn't really mention any other concerns.

### Titling... verbs vs. nouns; different verb cases

I don't like any solution to the question of consistent titling. I prefer gerunds for a lot of reasons, and that's what we've settled on for this repo. But then sometimes they seem like the weirdest possible way to write a title (like `Administering Aggregate` rather than `Aggregate Admin`, for example). 

A different verb form doesn't solve the problem. Whatever the advantages or disadvantages of imperative ("Install the thing") or infinitive ("How to install") over our chosen gerun ("Installing"), sometimes a verb just doesn't seem quite right -- like a page about "Form Uploader", for example. This is fine at the point of titling a single page, but then sometimes you end up with TOCs that have a list of titles with wildly divergent grammar forms.

I have done my best to end up at a place that seems as reasonable as possible. But I'm open to specific and concrete suggestions.

### DIFF
--- a/aggregate-tomcat.rst
+++ b/aggregate-tomcat.rst
@@ -49,6 +49,25 @@ Encrypted forms can be used in conjuction with either of the first two suggestio
 
 If you are not using encrypted forms and are handling sensitive data, a computer security specialist should review your system and your security procedures. 
 
+Database Systems
+~~~~~~~~~~~~~~~~~~
+
+ODK Aggregate works with any of these database servers:
+
+- MySQL
+- PostgreSQL
+- Microsoft SQL Server
+- Azure SQL Server
+
+Most deployments will use either MySQL or PostgreSQL, as they are the two most prevalent open source relational database servers.
+
+MySQL is the most popular, so you will likely find more qualified professionals to install, administer, and debug it. However, you should consider PostgreSQL if you plan to:
+
+- collect geographic data
+- use forms with a very high number of questions (over 200)
+
+PostgreSQL has better built-in support for geographic data, and MySQL's table have a row size limit that will affect perfromance for very large forms.  
+
 .. note::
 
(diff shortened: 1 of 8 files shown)


aggregate-tomcat.rst
```diff
@@ -49,6 +49,25 @@ Encrypted forms can be used in conjuction with either of the first two suggestio
...
+- Azure SQL Server
+
+Most deployments will use either MySQL or PostgreSQL, as they are the two most prevalent open source relational database servers.
+
+MySQL is the most popular, so you will likely find more qualified professionals to install, administer, and debug it. However, you should consider PostgreSQL if you plan to:
+
+- collect geographic data
+- use forms with a very high number of questions (over 200)
+
+PostgreSQL has better built-in support for geographic data, and MySQL's table have a row size limit that will affect perfromance for very large forms.
```
yanokwa: s/perfromance/performance

adammichaelwood: one day I will figure out how to turn spell check on automatically...

PR #453, opened 2018-01-26
fixes show details labels

### LINKED ISSUE #448: show screenshot vs show details

adammichaelwood:
closes #448

If a details block only contains a screenshot, the show/hide label says "Screenshot" instead of "Details"

## What new work does this PR create?

Now that so many labels on some pages say "show screenshot," it isn't clear why the other screenshots are displayed by default. This will take some thought as I work through editing for style.


adammichaelwood:
What you correctly describe as expected behavior is precisely what I see locally.


screen shot 2018-01-29 at 4 52 52 pm

yanokwa:
I see these changes in CI, but not on my local machine. The problem is likely on my end, so I'll merge and check the live site.

Issue #436, opened 2018-01-18
Build and maintain inventory of migrated content


lognaturel:
A lot of content has been moved over from the ODK website, [developer wiki](https://github.com/opendatakit/opendatakit/wiki) and other sources into the docs here. It's hard to tell which content should be considered the "source of truth."

I have started a document at https://docs.google.com/spreadsheets/d/1IfZydsmSWvxEzBE5yO7-2FZVJt6UY4dDGc4GzCjlnmc/edit#gid=0 to keep track of the migration status. Each resource from the website and wiki should eventually be listed there. It might be useful to use some kind of crawler or site mapper to generate the initial list of resources and then search can be used on the docs site to find the corresponding pages. I have given write access to @adammichaelwood @yanokwa @ankita240796 @Divya063. Everyone has commenting access and I can add more people if interested.

This will help make sure that all redirects are properly addressed in the [website migration](https://forum.opendatakit.org/c/contributions/website) and reduce confusion caused by duplicate content.

adammichaelwood:
I've added a couple sheets to the document. The original shee tnow has a list of all the old-site pgaes that are specifically relevant to ODK [1.0]. There's now also a sheet for blog posts (which might contain some useful content) and the ODK 2.0 pages (which I figured might be useful for other ppl). I'm sifting through them as I walk through on an editing pass through the whole doc repo.

adammichaelwood:
The main sheet (**ODK Pages**) has been completely updated with the current status of everything from the old site that is directly related to ODK.

Along with that, the pages that have been migrated have all been updated on the old site with a deprecation notice and grayed out text. Barring some reason I don't know about, it seems like those pages (marked TODO: Redirect) could all be `301`ed immediately.

Several pages are marked TODO:?  -- @ln and/or @yanokwa can probably provide some guidance here. This is mostly content that is in-theory appropriate to docs, but might no longer be relevant. (PLus a few other random things.)

A whole section of pages are related to editing XForm XML directly. There is already a conversation about those pages in issue #239.

I'll add the wiki pages next and we can sift through that. Is the plan to completely deprecate the github wiki?

adammichaelwood:
I believe all of this content that is doc-related has been accounted for,
either moved over or at least issues made of them, with the text copied in.

IT would be good to get additional review, though,
before the old site disappears.

What I'm most concerned about before the new site launches is redirects.
[...]

Issue #401, opened 2017-12-18
Clean up repo


adammichaelwood:
At very least, drop all the content into a [source directory](http://www.sphinx-doc.org/en/stable/glossary.html#term-source-directory).

This would mean also updating build scripts, etc.

I still [don't think it's a good idea to subdivide pages into a series of folders](https://github.com/opendatakit/docs/issues/35), as that needlessly complicates page organization. However, moving all the source files out of the top level would separate them from all the other things: build config files, utility scripts, settings, etc.

adammichaelwood:
From Slack:

> Hi all. I've started work on https://github.com/opendatakit/docs/issues/401 .
>
> I moved all the rst files into a src directory and fixed all the image paths to be absolute. This works, but the problem is that Sphinx reflects the directory structure in the path. So http://docs.opendatakit.org/collect-intro becomes http://docs.opendatakit.org/src/collect-intro.
>
> I am having trouble finding a way in restructuredText to control paths. Does anyone have any documentation they can point me at to help with this problem?

My understanding is that if you put `conf.py` in the `src` directory, and update the build script to specify it as the source directory, sphinx-build treats that as the document root. (See the [sphinx-build docs](http://www.sphinx-doc.org/en/stable/man/sphinx-build.html)...)

(There's also [something about putting `conf.py` in it's own directory](http://www.sphinx-doc.org/en/stable/glossary.html#term-configuration-directory), but I don't understand the point of that.

I am unclear whether `img`, `vid` or `static` need to be under `src` in this case. (I think they do.)

Ideally, we'd have a `src` dir that is everything content related, a `util` dir that has all our custom scripts and utilities.

adammichaelwood:
To clarify, I *think* that if once you move everything to `src`, then:

```
sphinx-build -b dirhtml . build
```

will need to be changed to 

```
sphinx-build -b dirhtml src build
```

and then everything will work as it is supposed to.

That will need to be updated in the build utility (`b.sh`), the documentation, and the CircleCI config.

jbeorse:
Ah, thanks for the update. This does work and I had gathered as much about the build command change. I had forgotten the CircleCI config though. I will update that and issue a pull request.

jbeorse:
Actually, I have only moved source files into the `src` directory and done nothing to move custom scripts and utilities into `util`. Which files would need to be moved? `video.py`? 'style-test.py`?

ankita240796:
[...]

PR #480, opened 2018-02-01
Port ODK 2.0 documentation into Sphinx

jbeorse:
addresses #436

It also addresses the steady stream of forum posts expressing confusion at the status of the ODK 2.0 tools and their documentation. Here is the most recent one: https://forum.opendatakit.org/t/about-odk-2-0-tools/11588/2

I have moved the entirety of the ODK 2.0 documentation into this repository. I have ported the docs into reStructuredText and applied the ODK style guidelines. [This site](http://opendatakit-dev.cs.washington.edu/2_0_tools) was my source, but I also did some minor updating of the text to fix errors as I found them, remove release designations (the suite is now fully released), and break up the pages into a similar structure as the rest of the docs. 

I have added a fairly unobtrusive link to the ODK 2.0 docs to the home page. But I imagine this will be a point of discussion, that this language is particular, and I intended this only as a suggestion. I will update this to whatever structure the community agrees on.

The purpose of this PR is to migrate the ODK 2.0 docs into the main documentation website. However after this merge I will continue improving and adding to the ODK 2.0 docs. In particular there are community made tutorials that will be integrated.

This PR also does not complete the migration process from the old website. The Google Doc linked in issue #436 has a list of more documentation that needs to be addressed. 

1. Unfortunately this PR is quite large. I am not aware of a way to break it up into smaller pieces because the various ODK 2 doc pages all reference each other. 

2. I attempted to use the new style-test.py before submitting, but the script crashed.


ankita240796:
Hi @jbeorse, Sorry for the issue here. The test fails because index is not checked before accessing the character at that position. I have issued a PR to fix that. Once that is merged, the build can be restarted. Also you can apply those changes locally to the ``style-test`` script and then run it. That will fix the issue!

jbeorse:
Thanks @ankita240796! I'll check that out locally and start cleaning up any errors I find in my docs.

getting-started-2.rst
```diff
@@ -0,0 +1,473 @@
...
+.. _using-odk-2-config-reset-server:
+
+Resetting the Server
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Once you have the application running on the device, you will typically need to reset the contents of the application server. While the :guilabel:`Reset App Server` button on the device can shuffle the various supporting files between the device and the server, it will not destroy data tables that already exist on the server; this is intentional -- we want to minimize the potential for accidental loss of data.
+
+.. note::
+
+  Whenever you are developing an application, and have found a need to add a new column to an existing table, you will need to manually delete the data tables from the server before using the :guilabel:Reset App Server` button from the device.
```
ankita240796: You have left a backtick here in the `:guilabel:` role due to which the test fails. Please fix that!

jbeorse:
Thanks again. I fixed the error you pointed out, as well as two spelling errors caught by the script. Now when I run it I see 0 errors. 

I do seem to have upped the warnings from 543 to 724.

data-permission-filters.rst
```diff
@@ -0,0 +1,431 @@
...
+  .. code-block:: sql
+
+    SELECT MAX(crop_height) as max_height FROM crop_plantings WHERE _default_access != ? or _row_owner = ? bind parameters = [ "HIDDEN", odkCommon.getActiveUser() ]
+
+.. _data-permission-filters-effective-access:
+
+Effective Access
+-----------------------
+
+As mentioned above, when a SQL query is processed inside the ODK Services layer, it is first examined to see if the result set contains the columns :th:`_sync_state, :th:`_default_access`, :th:`_row_owner`, :th:`_group_read_only`, :th:`_group_modify`, and :th:`_group_privileged`. If it contains all six columns, then a synthesized column, :th:`_effective_access` is added to the result set. That column returns one of *r*, *rw*, *rwd*, or *rwdp* (with the *p* indicating that a user can change permissions for the row as well) to indicate the level of access the current user has on the rows in the result set.
```
ankita240796: Missing backtick in ``:th:`sync state``

jbeorse: Thanks, just pushed a fix.

ankita240796:
@jbeorse Try running on the files you have changed in the following way:
``python style-test.py [file1] [file2]...``
You can also generate a csv output for errors:
``python style-test.py -o out.csv [file1] [file2]...``
It will make it easier for you to see the warnings only in the files changed by you. There are many false positives as well.

adammichaelwood:
This might raise the relative priority of moving the rst files into a `src` subdirectory...

adammichaelwood:
> I do seem to have upped the warnings from 543 to 724.

GH says this PR has 92 files in it, so your warnings-per-file rate is much lower than what we have already.

jbeorse:
@adammichaelwood I agree that the .rst files everywhere are a bit of a mess, and my commit certainly doesn't help with the clutter. I left everything at root level to be consistent with the current style, but will happily move them into a more organized folder structure. 

Is there an open issue describing what the move should look like?

adammichaelwood:
> Is there an open issue describing what the move should look like?

Tentatively, #401 

Basically, I just want to move all the rst files, and conf.py, into a dir called `src` (or something) and all the various and increasing python authoring and build utilities into `util`. That would at least keep things moderately organized. I wouldn't mind topical subdirectories *in theory* if they didn't effect output URI

jbeorse:
I've opened https://github.com/opendatakit/docs/pull/496 to address #401 . If/when that gets merged I'll update this PR to follow suit.

jbeorse:
Since PR #496 has been merged I've updated this PR with those changes, along with the style guide fix in PR #501. This should be ready to go again.

lognaturel:
First, this represents a ton of work and looks really fantastic. Thank you @jbeorse.

The placement right now is a bit odd because the "developing with ODK", "integration" and "reference" sections all focus on 1.0 concerns. Did you consider the option of an analogous but separate docs site? I'm wondering whether that might give each doc set more of an opportunity to be a first class citizen. The website could then link out to the two doc sets side by side. This would also make doc search more useful so that users don't have to very carefully note what suite each page is for (e.g. search for "calculation").

jbeorse:
Just commenting here because I was told that the issue of how to organize the docs and where to put ODK 2 is being discussed by the PMC right now. I can think of pros and cons for both sides, but I'm open to going whichever way the PMC decides.

adammichaelwood:
@yanokwa briefly discussed the possibility of having our cake and forking it too.
That is....

Maybe we could run two Sphinx builds out of the same repo, sharing some resources (like contributor guide, main way-finding, etc.) across the two builds. I'm looking into it and will report back. I tentatively think it is the best idea so far, but the devil will be in the details of actually making it work.

jbeorse:
Hi All,

@wbrunette have implemented @adammichaelwood's plan (read his [forum post](https://forum.opendatakit.org/t/odk-2-0-documentation-migration-to-docs-opendatakit-org-request-for-reveiw/11627/2?u=jeff_beorse)). These changes allow us to share a repo between ODK 1 and ODK 2 but deploy as different sites.

## Change Summary

1. Separated ODK 1, ODK 2, and Shared doc files into three new src folders.
2. Updated the build and utility scripts to work with this new layout. You can choose to either build ODK 1 ODK 2, or both. The first step of a build is to merge the shared-src folder with the appropriate odkX-src. Then the build proceeds much like it did before.
3. Updated the CircleCI configuration to build ODK 1 and ODK 2 separately. Note that we temporarily removed the deploy features so that we don't accidentally deploy docs when doing this migration and testing.

## Notes

### Makefile

We switched to using the make file that was provided, away from the b.sh and away from directly calling the utility scripts. Here are some example commands:

- To build ODK 1
```
make odk1
```

- To build ODK 2
```
make odk2
```

- To check the style of ODK 1
```
make odk1-style-check
```

- To check the spelling of ODK 1
```
make odk1-spell-check
```

- To perform all checks of ODK 1 
```
make odk1-check
```

- To check and build everything
```
make check-all
```

Are there any missing commands that we should add?

### Utility Scripts

 Also, the python scripts for spelling and style checking still work, but require you to specify which src folder to check. For example:
```
python style-test.py -d -r odk2-src
```
But again, you don't need to call them directly; you can use the make file. There more examples in the CircleCI yml file.

### Content changes

We tried to keep changes to actual content to an absolute minimum. The only places where this turned out to be necessary were in shared source files that referenced files in the ODK 1 or ODK 2 source. Mostly this leaned towards examples, and we generally tried to replace these by copying the relevant lines locally.

## To Do

- After hearing feedback we plan to make a matching **make.bat** file for windows command line.
- We also need to update the contributing guide with the new build commands after they have settled.
- Re-enable deploy features.
- Separate further into *user* docs and *developer* docs (this can be done in a future PR).

adammichaelwood:
Very exciting.

jbeorse:
I merged in the latest changes from master, most notably the pdf building. It took some finagling but I think I've got it working for both ODK 1 and ODK 2. Added a new make command for each:
```
make odk1-latex
```

style-test.py
```diff
@@ -503,7 +503,7 @@ def gen_list(paths = None):
     exclude_checks()
 
     # run the checks on docs
-    run_checks(paths, False, False)
+    run_checks(paths, None, False, False)
```
bansalnitish: This should also have ``src_path`` as an argument.

ankita240796: This should also have ``src_path`` as argument.

ankita240796:
Hi @jbeorse, can you also please include ``.pyc`` files in ``gitignore``.

adammichaelwood:
I'd like to get this merged as soon as we can, since this kind of a massive re-structure is going to mean parallel PRs are going to start causing a lot of merge conflicts....

I'm reviewing now. 
@yanokwa needs to weigh in on deployment issues
Anyone else who really needs to get in the loop on this?

adammichaelwood:
I love the `make` commands, btw.

adammichaelwood:
I've reviewed this, and I think this is going to be the best way to move forward.
From my perspective, there's nothing to stop this from being merged in its present state --- 
assuming everything works from a deployment standpoint.

There's a few things I'd like to work out, but they can be done moving forward:

 - At present, there's not a sensible way for `shared-src` documents to reference other pages. 
   This PR removes a number of links to odk1 pages in the contributors guide, 
   and then other links are treated as external links with full URLs. 
   I'd prefer to some way for us to be able to link to either docset from shared docs.
   I suspect this would involve using [intersphinx](http://www.sphinx-doc.org/en/master/ext/intersphinx.html) and either some build-time regex replace
    or a custom role.

 - The template/design/js files are in `shared-src`, which is fine. 
    But as long as we're doing all this build-time file shifting,
    I'd love to see all that extracted out so that `src` directories only contain content.

 - I'm not sure if FAQ needs to be shared. At least, at the moment it probably doesn't.

 - Likewise with the Glossary. Maybe we need separate Glossaries? Or maybe we need little tags or something to differentiate ODK1 and ODK2 terms...

 - Some of the "ODK1 vs. ODK2" content should probably be shared. It doesn't have to be in the same place, BTW. It could be in the front matter on ODK2, like it is now, and in Reference in ODK1.

 - This has been mentioned before, but --- now that I'm seeing it in two tabs on a browser ---
    There really does need to be some visual difference between the two doc builds.
     An ODK2 logo would help (is there one?), and some language in the header. 
    But that wouldn't be enough.
    We need something for those of us who will be regularly dealing with both builds,
    that will immediately clue us in to which set we're looking at,
    like different background colors or a corner ribbon or something...

downeym:
@adammichaelwood @jbeorse I haven't dived into the code, but would a quick CSS color change for background of the sidebar be sufficient? Perhaps using the 3 colors of the current ODK logo (blue, brown, green) the various groups?

adammichaelwood:
> Perhaps using the 3 colors of the current ODK logo (blue, brown, green) the various groups?

I'm not sure exactly what you're proposing here. 


> a quick CSS color change for background of the sidebar be sufficient? 

One of the problems with changing a color in this theme is that color definitions happen in LESS, and then are compiled to CSS. The result is that patching a color in CSS can sometimes require a bunch of declarations and can sometimes be hard to track all of them down. (For example, sometimes a background color is defined for a section, and then re-defined for paragraphs...)

Our handful of design tweaks haven't really touched this too much, and have all been restricted to CSS. 

That doesn't mean what you're suggesting isn't a bad idea. Just that it's hard to know ahead of time what will be "quick" and what will be a pain.

downeym:
So as a specific example, what about changing the value of `background-color` on `div.wy-side-nav-search`? Blue for ODK1, Green for ODK2. 

And maybe the grey sidebar too if the contrast looks wonky.

Hopefully that wouldn't muck with the styling for the actual documents themselves....

adammichaelwood:
The search box isn't always visible --- that is, if you've scrolled down.

The main color on the sidebar would be good, but I don't know what, besides dark grey, would look good. Another option is to change the main section background from light gray to white. I'm worried, though, that this would subtly reinforce "OKD2 is newer and better".

Another option is a sticky nav bar along the top (https://getbootstrap.com/docs/3.3/examples/navbar-fixed-top/ for EXAMPLE, not that we would use Bootstrap). This could be (for example) blue for ODK1 and Green for ODK2, and could additionally have the breadcrumb path to the current page and a search bar.

downeym:
Good points. Perhaps for the sidebar there are appropriate super-pale
shades of the upper logo/search that would work? Here's an example with
blue on one of our sites: http://docs.osc.dial.community/en/latest/

I agree about the main document background color -- you'd probably want
them both at similar saturation values, whatever the shade.

jbeorse:
Thanks for the thorough review and comments @adammichaelwood. I agree that merging sooner and working out these kinks moving forward seems most favorable. 

Just so I'm clear on what needs to be done before and after the merge, my plan right now is to:

- Update style-test.py re @ankita240796's comments
- Add .pyc to gitignore
- Working with @adammichaelwood  and @downeym to update the styling to give a clean visual indication of whether you're looking at ODK1 or ODK2 docs. 

The rest of your bullet points sounded like items for future pull requests? Is that accurate? 

Regarding the styling change, I don't have strong opinions and am generally happy to go with what the groups finds most appropriate. A nav bar and a color change to the side bar seem good to me.

adammichaelwood:
I like that DIAL sidebar ---
I'd support a patch to do something like that here.

That could be in addition to a top nav bar. 
We could settle on a main color for 1 and 2 ---
(Then maybe we could eventually switch to "ODK Blue" and "ODK Green" :) )

I think probably this can all be taken care of in iterative PRs,
and we should merge this one as soon as it is fully vetted.

adammichaelwood:
> Working with @adammichaelwood and @downeym to update the styling to give a clean visual indication of whether you're looking at ODK1 or ODK2 docs.

I *think* this can wait for another PR. 
My suspicion is that it will touch enough things to complicate matters
if we try to add it to this one.
(Not to mention divergent PRs if it takes more than a day to complete.)

adammichaelwood:
BTW - what are the URLs going to be?

jbeorse:
> BTW - what are the URLs going to be?

Good question. Do @wbrunette or @yanokwa have opinions on this? 

>I think this can wait for another PR.

Sure thing. I'll get those small fixes in now.

wbrunette:
I agree that most of these things can be worked on after this PR in smaller PRs that will be easy to merge and have less conflicts after multiiple PRs are done.

I propose we merge this PR as a starting point with a focus on making sure everything on ODK1 documentation is continues to work as expected. We then turn the deploy back on for ODK1 and take care of the suggestions from @adammichaelwood. This will allow work on ODK2 updates with smaller PRs. 

With regards to the URL, it seems like something that can be decided on post merge as future PRs will need to be done. Also, the URL will likely change anyway when ODK2 is renamed and we will likely have re-direct links.

adammichaelwood:
> With regards to the URL, it seems like something that can be decided on post merge as future PRs will need to be done.

Assuming the ODK1 build files all deploy correctly to their current domain... sure.

jbeorse:
I've added the requested changes form @ankita240796 .

yanokwa:
Thanks for the great work on this all! 

Agreed that we should merge and start applying whatever fixes are needed. I *think* it should go smoothly, but to give us the best chance, I'll start that process tomorrow when most of the folks who worked on this are up and about (afternoon, Pacific Time)

PR #496, opened 2018-02-06
Cleanup

jbeorse:
addresses #401 

A new directory is created: `/src`

All `.rst` files are moved into `/src`. The `/_static`, `/img`, and `/downloads` folders are all moved into `/src`. In addition `conf.py` is moved into `/src`. 

`Makefile` and `make.bat` are updated with the new source directory. The contribution guide, README, `b.sh`, and `config.yml` are updated to the new build command: `sphinx-build -b dirhtml src build`

Updates `config.py` to include the root directory in its source path (so that `video.py` can be found). 

This does not move all custom scripts into `/util`. I am unclear on which files this would include.


adammichaelwood:
This seems to work, though -- as mentioned -- there are other files that need to be moved, etc.

 - Various scripts etc need to be moved to `util`
 - `vid` needs to be moved into `src`, but that also requires a config change somewhere
 - style testing needs updating to point to `src`, but the build doesn't break now because style testing just finds no rst files and so produces no errors.

So I'm going to merge this because it affects basically everything and we all need to update our other PRs...

Thanks for taking the time to do this @jbeorse

PR #517, opened 2018-02-12
Add note on filenames to form update guide

### LINKED ISSUE #507: Add a note on not changing filenames when updating forms
https://forum.opendatakit.org/t/new-version-of-form-wont-upload-to-aggregate/4215/5

ankita240796:
closes #507 

Added a note on having same filenames when generating newer versions of form.

### DIFF
--- a/src/form-update.rst
+++ b/src/form-update.rst
@@ -100,9 +100,9 @@ These changes can be made as:
 
 .. note::
 
-  You need to use the same filename when generating newer versions of your form since the XLSForm generator pays attention to the filename. If you use different filenames, you would need to delete the previous version to upload the newer version of form.
+  You need to use the same filename when generating newer versions of your form since the XLSForm converter uses that filename in creating the form structure which is used by the server to help determine uniqueness. If you use different filenames, you would need to delete the old version of the form to upload the newer version.
 
-  In the examples below, we have different filenames since that is necessary on the server.
+  In the examples below, we use different filenames for version upgrade since unique filenames are required for uploading files to the documentation website.
 
 These are the :file:`.xml` files for the above forms:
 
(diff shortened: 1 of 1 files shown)


src/form-update.rst
```diff
@@ -98,6 +98,12 @@ These changes can be made as:
 
   "Example_form", "example_id", "English", "2017120701"
 
+.. note::
+
+  You need to use the same filename when generating newer versions of your form since the XLSForm generator pays attention to the filename. If you use different filenames, you would need to delete the previous version to upload the newer version of form.
```
yanokwa: This is hard thing to communicate!

I'd say, "You need to use the same filename when generating newer versions of your form since the XLSForm converter uses that filename in creating the form structure the server uses to help determine uniqueness. If you use different filenames, you would need to delete the old version of the form to upload the newer version."

src/form-update.rst
```diff
@@ -98,6 +98,12 @@ These changes can be made as:
...
+
+  In the examples below, we have different filenames since that is necessary on the server.
```
yanokwa: It's not clear what the "server" is here. I think you mean the Docs server, right?

So "In the examples below, we use unique filenames since unique filenames are required to upload them to this documentation website". Or something like that.

ankita240796: Yes, that is right. I will fix the language :sweat_smile:

ankita240796:
@yanokwa I made the required changes!

yanokwa:
I edited my comment, but you didn't see the change, so I'll ask here. Do you think adding something like this would be useful?

```
If you use the same filename, you can upload the form and ODK Aggregate will automatically replace the old version and keep the existing form data. If you use different filenames, you would need to delete the old version of the form (and the existing form data) in order to upload the newer version
```

ankita240796:
Yes, this will be definitely useful. I will try add an example as well on some difference in form structure xml due to filename change.

ankita240796:
@yanokwa I have made the changes!

PR #515, opened 2018-02-10
Literate programming

ankita240796:
closes #466 

## What is included in this PR?
- Add function in ``style-test`` script to parse python code-blocks.
- All checks except ``uk vs us`` and ``space check after period`` are generated from style-guide. These checks are pre-defined in ``proselint-extra.py``.
- Checks to be run through proselint are written to file ``style-checks.py``.
- Extra checks are written to file ``extra.py``.
- Script ``extrarun.py`` to run the extra checks.

## What problems did you encounter?
 Warning: uninterpreted text role ``:doc:``, ``:file:`` and ``:rst:dir`` when publishing doctree. I am unable to find a way to get rid of that.
EDIT: Found a fix here: https://github.com/robotframework/robotframework/issues/2093, https://github.com/robotframework/robotframework/pull/2152 :tada:

Issue #402, opened 2017-12-18
Editor's Pass on All the Docs


adammichaelwood:
I need to take several days and just do a close-reading on every page and every picture.

adammichaelwood:
I've done a lot here, and opened issues for specific docs that need to get a once-over.
I'm not sure it makes a lot of sense to keep this issue open.

If you see a doc that needs some work --- open an issue for that doc in particular.

PR #520, opened 2018-02-13
WIP: Style guide updates along with editing pass

adammichaelwood:
Filing as a PR so I don't lose track of it,
but this shouldn't be merged (or much even looked at)
until after editing pass is complete #402  AND
literate programming additions are completed #515


lognaturel:
I finally took a careful look through this and it looks like we've already incorporated most of the substantive changes! Thanks so much @adammichaelwood and hope you're well. 😊

PR #522, opened 2018-02-13
Add spell checker on PRs

### LINKED ISSUE #434: Add spell check on PRs
I was just noticing that we have some spelling errors. It'd be great if we could add a spell check (perhaps as part of the sphinx-build) that would warn when a build has spelling mistakes. 

An ideal solution would integrate into sphinx, be an active/healthy project, and give us the option to warn or fail on projects, and allow us to ignore some words.

@ankita240796 this might be a good issue for you to take on when you have some time!

ankita240796:
closes #434 

- Added ``sphinx-contrib.spelling`` package for checking spellings.
- Custom list of spellings in individual files to be not treated as misspellings.
- New list of words in ``src/spelling_wordlist.txt`` to add to the dictionary.
- Added spell checking to build script.
- Added spell checking to circleci build.


ankita240796:
I have made a separate line edit PR #523. Once that is merged, I will rebase this and the build will pass.

yanokwa:
@ankita240796 This is looking fantastic! I had a couple of quick questions..

1. Can you describe how you decided what goes into an individual file vs what goes into the spelling list? It's not obvious to me, so we should maybe add it to the contributing guide. 
1. Some of the case sensitive spellings seem like they shouldn't be added to the spelling list. For example, `ODk` and `Odk` aren't correct. Is there a reason those are allowed?

ankita240796:
> This is looking fantastic!

Thank you @yanokwa :smile: 

> what goes into an individual file vs what goes into the spelling list? 

The list of spellings like app, Gmail etc was a repetition in all files so that was placed in the spelling list. In some files we have spellings like abcd, un which should be allowed only in those files. So, I decided to place them in individual files. So, my idea was to add a repeated list from each file to an individual file.

> Some of the case sensitive spellings seem like they shouldn't be added to the spelling list.

I have added words like ``ODk`` and ``Odk`` to the individual list since there are cases where these words are used as examples to show wrong usage. Further, this check can be performed by style-guide. So, I thought to exclude them from the spell checker. I am not sure about it though. Another solution would be to add such words in ``spelling`` directive in individual files when they are used as examples otherwise consider them as spelling mistakes.

src/spelling_wordlist.txt
```diff
@@ -0,0 +1,219 @@
+Admin
+admin
+anonymousUser
+app
+App
+apps
+Apps
+appspot
+atleast
```
yanokwa: I think `atleast` is not a word. It's `at least`

src/aggregate-upgrade.rst
```diff
@@ -1,3 +1,24 @@
...
+  init
+  datetime
+  itemset
+  jr
+  ui
+  clickjacking
+  JBoss
+  Resteasy
+  Memcache
+  pre
```
yanokwa: Duplicate

yanokwa:
Good explanation, @ankita240796.

I like the goal of putting repeated words on the spelling list. I think you missed `apk`, `adb`, and `pre`.

I think for words like `Odk` where we've use it as an example of incorrect usage, we should probably just put it in that specific file. But I don't feel strongly about it. You, @adammichaelwood, and @Divya063  are the docs experts, so you should decide!

src/aggregate-upgrade.rst
ankita240796: Yes, I have sorted the ``spelling wordlist``. I missed it here. I will fix that and remove the duplicates as well.

ankita240796:
Thanks @yanokwa!

I feel it would be fine to put words like ``Odk`` in files itself but if there are too many such files we can depend on style guide testing for this.

ankita240796:
Line edit #524 has to be merged before this one for build to pass.

yanokwa:
There seem to be a couple of mispellings when I build it locally...
```
aggregate-data-access.rst:90:acknowledgement:["acknowledgment"]
form-widgets.rst:739:acknowledgement:["acknowledgment"]
form-widgets.rst:1:acknowledgement:["acknowledgment"]
```

.circleci/config.yml
```diff
@@ -19,12 +19,16 @@ jobs:
...
+            sudo apt-get install python-enchant
             sudo pip install -r requirements.txt
       - run:
           name: Style Guide Testing
           command: |
             paths=$(git diff origin/master... --name-only)
             sudo python style-test.py $paths
+      - run:
+          name: Spell Checker
+          command: sphinx-build -W -b spelling src build/spelling
```
yanokwa: This creates a `spelling` folder in `build` and everything in `build` gets automatically uploaded to the S3 and Cloudfront. We probably don't want that `spelling` folder on the website, so in the `exclude_from_upload` section in the `s3_website.yml` file, add `spelling`, so that folder gets excluded.

ankita240796:
https://github.com/opendatakit/docs/pull/522#issuecomment-365758402
This is really strange as I am unable to get these locally. Also, `` acknowledgment`` is the preferred form in US English (default for spelling checker) but still no error is being produced.

yanokwa:
Very weird. I'm running Python 3 with sphinxcontrib-spelling 4.0.1 and pyenchant 2.0.0 on my mac. Those are same versions on CircleCI which works great.

ankita240796:
@yanokwa I am using Ubuntu. I am still not clear about why this is happening. 
http://sphinxcontrib-spelling.readthedocs.io/en/latest/customize.html#input-options
According to this, the default language is set to 'en_US' and in  the code here:
https://github.com/sphinx-contrib/spelling/blob/master/sphinxcontrib/spelling/checker.py#L21 , dictionary is created with the same language.

Also I tried some other things which further confuses me:
```
>>> import enchant
>>> dic = enchant.DictWithPWL('en_US', None)
>>> dic.check('acknowledgement')
True
>>> dic.tag
'en_US'
>>> dic = enchant.Dict('en_US')
>>> dic.check('acknowledgement')
True
>>> dic.check('colour')
False
>>> dic.check('acknowledgment')
True
>>> dic = enchant.request_dict('en_US')
>>> dic.check('colour')
False
>>> dic.check('acknowledgement')
True
>>> dic = enchant.Dict()
>>> dic.check('color')
True
>>> dic.check('colour')
True
>>> dic.tag
'en_IN'

```

yanokwa:
The problem is indeed the versions of the various dictionaries. 

We are using hunspell on CircleCI and when I install myspell (`sudo apt-get install myspell-en-us`) on the build machine and run the spell checker, it correctly flags `acknowledgement` as incorrect. But it also flags other words improperly (e.g., dataset, resize, multi, bulleted).

I think hunspell is the correct choice since it's used on a lot of active projects from LibreOffice to macOS. The real problem lies in Debian's English dictionary. The package is [hunspell-en-us](https://packages.debian.org/stretch/all/hunspell-en-us/filelist) and a search in `/usr/share/hunspell/en_US.dic` shows that both spellings 

```
$ grep acknowledg /usr/share/hunspell/en_US.dic
acknowledgeable
acknowledgedly
acknowledged/U
acknowledge/GZDRS
acknowledger/M
acknowledgment/SAM
acknowledgement/SAM
```

@Divya063 I propose we correct the spelling to acknowledgment in the source, and add it to the style guide checks. Then because we want to be good open source citizens, can you file a bug upstream on the Debian dictionary? I think https://bugs.debian.org/cgi-bin/pkgreport.cgi?pkg=hunspell-en-us is the right place for that.

ankita240796:
@yanokwa I have corrected the spelling in a separate line edit PR #528 . I will add the spelling to style guide checks as well once line edit PR is merged.
I will file a bug upstream on the Debian dictionary for sure!
By the way, I am Ankita :stuck_out_tongue:

yanokwa:
ARGH! I'M SORRY 😅

ankita240796:
That is totally fine :sweat_smile:

ankita240796:
I have filed an issue on the Debian dictionary :tada:

yanokwa:
Nicely done. Once you add the spelling to style guide checks, I'll be happy to merge!

ankita240796:
@yanokwa I have added the spelling in PR #515 (https://github.com/opendatakit/docs/pull/515/commits/9354d3fcbdefdc833a3a790b791c7229f5a0d102) to avoid merging the style guide and spell checker work. Is that fine?

yanokwa:
Totally didn't see that! Merging now :)

PR #537, opened 2018-02-25
PDF output

### LINKED ISSUE #271: Add PDF output

ankita240796:
closes #271 

- Added pdf output generation
- Changed image ordering for html builder to choose gif over png
- Added png images as substitutes for gif for latex builder
- Script to resize images

- Large tables are not rendered properly

- There are still some underfull and overfull issues.
- We need to run xelatex engine twice

EDIT: The first commit needs to be dropped once #536 is merged!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,8 +63,9 @@ jobs:
               cd build/latex
               xelatex OpenDataKit.tex
               xelatex OpenDataKit.tex
-              cd ../..
-              cp build/latex/OpenDataKit.pdf build
+      - store_artifacts:
+          path: build/latex/OpenDataKit.pdf
+          destination: OpenDataKit.pdf
 
   deploy:
     working_directory: ~/work
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,6 +20,7 @@ jobs:
           command: |
             sudo apt-get install pngquant
             sudo apt-get install python-enchant
+            sudo apt-get install texlive-xetex
             sudo pip install -r requirements.txt
       - run:
           name: Style Guide Testing
@@ -31,6 +32,16 @@ jobs:
           command: |
             sphinx-build -b spelling src build/spelling
             python util/check-spelling-output.py
+      - run:
+          name: Build PDF
+          command: |
+            sphinx-build -b latex src build/latex
+            sudo python util/resize.py
+            cd build/latex
+            xelatex OpenDataKit.tex
+            xelatex OpenDataKit.tex
+            cd ../..
+            sudo cp build/latex/OpenDataKit.pdf build
       - run:
           name: Build docs
           command: sphinx-build -W -b dirhtml src build
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,6 +51,7 @@ jobs:
             - build/*
             - s3_website.yml
   build_pdf:
+    working_directory: ~/work
     docker:
       - image: tianon/latex
     steps:
@@ -89,9 +90,13 @@ workflows:
   build_deploy:
     jobs:
       - build
+      - build_pdf:
+          requires:
+            - build
       - deploy:
           requires:
             - build
+            - build_pdf
           # We'd like to also filter by username and remove the if/fi above, but username filtering is not supported in CircleCI
           filters:
             branches:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,9 +53,10 @@ jobs:
   build_pdf:
     working_directory: ~/work
     docker:
-      - image: tianon/latex
+      - image: schickling/latex
     steps:
-      - checkout
+      - attach_workspace:
+          at: ~/work
       - run:
           name: Build PDF
           command: |
@@ -63,7 +64,7 @@ jobs:
               xelatex OpenDataKit.tex
               xelatex OpenDataKit.tex
               cd ../..
-              sudo cp build/latex/OpenDataKit.pdf build
+              cp build/latex/OpenDataKit.pdf build
 
   deploy:
     working_directory: ~/work
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,7 +20,6 @@ jobs:
           command: |
             sudo apt-get install pngquant
             sudo apt-get install python-enchant
-            sudo apt-get install texlive-xetex
             sudo pip install -r requirements.txt
       - run:
           name: Style Guide Testing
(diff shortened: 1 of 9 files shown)


ankita240796:
@Divya063 This would need your feedback and inputs!

Divya063:
@ankita240796   The size of texlive is large, this will slow the build process. Maybe we can use a [minimalistic distribution](https://tex.stackexchange.com/questions/119759/why-cant-tex-lives-size-be-reduced)

ankita240796:
> Sorry for the late response, I'm having my internal exams.

That is totally fine @Divya063. All the best for your exams!

Thanks for the feedback. I will try using a minimalistic distribution.  :tada:

yanokwa:
+1 on a smaller distribution. If that isn't super easy to use, then we should use another container (e.g., https://hub.docker.com/r/tianon/latex/) to build the PDF and configure CircleCI accordingly. That is, instead of build, we could have build_site, then build_pdf, then deploy.

Divya063:
@yanokwa The size of docker image of tianon/latex  is 1.312 GB, also I found [this](https://github.com/thomasWeise/docker-texlive) container which contain installation of texlive as well as other scripts whose image size is 1.046 GB. There are other containers as well which provides builder for HTML and PDF (e.g https://hub.docker.com/r/ddidier/sphinx-doc/ and https://hub.docker.com/r/umbrellium/sphinx-doc/)

yanokwa:
@Divya063 It's a tricky balance. I think we should pick a container that has lots of users and commits because it's more likely to be updated. https://hub.docker.com/r/tianon/latex seems to have that property and it's not that much bigger.

I do not think the latex container needs to have sphinx installed because we can build the latex source files, then pass them to a `build_pdf` machine that has the tianon image and that can do the latex build. Does that seem like a reasonable approach?

ankita240796:
@yanokwa can you please guide us a bit about the build since we are stuck here!

.circleci/config.yml
```diff
@@ -45,6 +50,21 @@ jobs:
           paths:
             - build/*
             - s3_website.yml
+  build_pdf:
+    working_directory: ~/work
+    docker:
+      - image: tianon/latex
```
yanokwa: Looks like tianon's image does not have some of the tools (e.g., git, ssh) that CircleCI needs. See https://circleci.com/docs/2.0/custom-images/#adding-required-and-custom-tools-or-files for that list. I tried installing them after the build but that didn't work.

Let's try `schickling/latex`. It's seems fairly popular and it built in my branch.

bansalnitish: Thank you for trying this out and fixing it! I will make the changes here :smile:

.circleci/config.yml
```diff
@@ -45,6 +50,21 @@ jobs:
...
+    steps:
+      - checkout
```
yanokwa: Instead of `checkout`, which checks out the git repo, you'll need to attach the workspace you persisted earlier. That workspace has `build/*` which is what you need.

```yaml
- attach_workspace:
    at: ~/work
```

.circleci/config.yml
ankita240796: Thank you for trying this out and fixing it! I will make the changes here :smile:

.circleci/config.yml
```diff
@@ -45,6 +50,21 @@ jobs:
...
+      - run:
+          name: Build PDF
+          command: |
+              cd build/latex
+              xelatex OpenDataKit.tex
+              xelatex OpenDataKit.tex
+              cd ../..
+              sudo cp build/latex/OpenDataKit.pdf build
```
yanokwa: I think `sudo` won't be installed on this Docker image, so remove it.

ankita240796:
PDF is copied over to build directory after upload of artifacts is completed. I think that is the reason for this issue.

ankita240796:
Can we upload the artifacts in build pdf job? Sorry, I am not much familiar with the build process :sweat_smile:

yanokwa:
I think instead of the ` cd ../..` and `cp build/latex/OpenDataKit.pdf build` you could do something like this. 

```yaml
- store_artifacts:
    path: build/latex/OpenDataKit.pdf
    destination: OpenDataKit.pdf
```
You can learn more about this part of the build process at https://circleci.com/docs/2.0/artifacts/

yanokwa:
@ankita240796 What is remaining on this PR?

ankita240796:
Sorry for delay the @yanokwa. I was travelling back college so wasn't able to update this. I think this is ready for a review now!

adammichaelwood:
I'd like to review the output PDF, but am unsure how....

ankita240796:
@adammichaelwood You need to install ``texlive-xetex`` on your system.
The commands for generating the output are:
```
sphinx-build -b latex src build/latex
python util/resize.py
cd build/latex
xelatex OpenDataKit.tex
xelatex OpenDataKit.tex
```
After this the PDF file will be generated.

I have also added it as an downloadable option.
For doing that, navigate to the docs repo and then:
```
cp build/latex/OpenDataKit.pdf build
```
and then run the html builder and you can see a downloable option for the PDF.

image

To install xelatex will take up a lot of space. I am not much familiar with docker usage but I will try that and update here so that you can easily test this!

adammichaelwood:
I'm happy to just review the built output from CircleCI, assuming that is the canonical PDF build. 

Unless we think contributors are going to need to build/test the PDF locally. I'm hoping not...

ankita240796:
Yes, the contributors would not really require to that but still I will try to find an easier setup and add to it contributors guide!

adammichaelwood:
Yeah, it'd be nice to let people know how to do it locally if they want to.
More importantly, though, is having relatively easy access to the PDF build from a current PR (which seems like we have, and need to get it into the contrib guide at some point).

ankita240796:
Sure, I will do all that once this is ready to go :smile:

ankita240796:
The tables on page 335, 336, 338, 345, 402 are some which have overlapping text.

src/conf.py
```diff
@@ -234,6 +250,21 @@ def setup(app):
 """
 
 # At bottom of every document
+doc_pdf = """
```
adammichaelwood: This message should just read:
Download this documentation as a PDF.
With the entire sentence being the link anchor text.

adammichaelwood:
@yanokwa and I discussed this and agreed this is good enough even with the tables issue.
The footer link text needs a fix (I made a comment inline).
Since this is mostly CircleCI stuff, I'd like @yanokwa to be the one that hits the big green button.

ankita240796:
I have fixed the footer link :tada:

yanokwa:
@ankita240796 When you get a chance, please file the issues about underfull/overfull problems so someone can take those on!

ankita240796:
> This is fantastic work

Thank you so much @yanokwa :tada: 

>  please file the issues about underfull/overfull problems so someone can take those on
 
Sure, I will do that!

PR #558, opened 2018-03-09
Clean up PDF naming and variables

yanokwa:
addresses #537

Rename file so it's more obvious what it is.
Rename variables so they are consistent.
Add some periods to sentences

### DIFF
--- a/src/_templates/footer.html
+++ b/src/_templates/footer.html
@@ -11,7 +11,7 @@
   {% endif %}
 
   <hr/>
-  <p class="footerText"> {{ doc_pdf }} <a href="{{ odkpdf }}" download> {{ odkdocs }} </a></p>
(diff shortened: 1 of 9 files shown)


util/resize.py
```diff
@@ -0,0 +1,32 @@
+import os
+import sys
+from PIL import Image
+
+def single_resize(imagedir, filename):
+    """Resize a single image."""
+    print("Resisizing %s ..." %filename)
```
ankita240796: ``Resizing`` needs to be fixed here

src/_templates/footer.html
```diff
@@ -11,6 +11,7 @@
   {% endif %}
 
   <hr/>
+  <p class="footerText"> <a href="{{ odkpdf }}" download> {{ download_pdf }} </a></p>
```
ankita240796: This should be ``odk_pdf``.

PR #580, opened 2018-03-19
Update build instructions

lognaturel:
There will need to be corresponding changes to the contribution guide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,12 +13,16 @@ If you can't find what you are looking for, try the [old docs on the ODK website
 ## Building and viewing documentation locally
 See the [Contributor Guide](http://docs.opendatakit.org/contributing) for detailed steps --- no prior experience needed!
 
-Once your environment is set up:
+Once your environment is set up, build and run the doc site with:
 ```
-sphinx-build -b dirhtml src build
-cd build
+make odk1
+cd odk1-build
 python -m http.server 8000
 ```
+
+To perform all spell check and style checks before building, use `make odk1-check` instead or `make odk1`.
+
+To build ODK 2 docs, replace `odk1` with `odk2`.
 
 ## How to contribute?
 


adammichaelwood:
My understanding is we're supposed to use the Make file.
(That's what I've been doing locally, based on the info provided in #480)

To build ODK 1

```
make odk1
```

To build ODK 2

```
make odk2
```

To check the style of ODK 1

```
make odk1-style-check
```

To check the spelling of ODK 1

```
make odk1-spell-check
```

To perform all checks of ODK 1

```
make odk1-check
```

To check and build everything

```
make check-all
```

Serving locally would then be:

```
cd odk1-build
python -m http.server 8000
```


@jbeorse

lognaturel:
That works too! In a local workflow there's no real need to copy the source directory and I do like knowing the underlying commands but I don't feel strongly about it if everyone would prefer using the makefile commands.

jbeorse:
@adammichaelwood is correct, you can no longer just build *odk1-src* with the Sphinx command because you first need to combine it with *shared-src*. Building *odk1-src* by itself should fail or at least complain about links to non-existent files. The same is true of *odk2-src*. They will certainly not create the expected website.

The commands @adammichaelwood listed are correct. The Makefile copies the source directories to a temporary combined source folder and runs that Sphinx command on the combined folder. Similarly, for style and spell checking it handles combing folders for you, but these can still be run manually if you want to check specific files. Notably, you can see that happening [here in config.yml](https://github.com/opendatakit/docs/blob/master/.circleci/config.yml#L29) where we only style check the changed files. 

I'm not sure how much we want to include in a README but I do not think it is correct to include that Sphinx command. At a minimum I would change line 18 to be ```make odk1`` and line 23 to 

> To build ODK 2 docs, replace `odk1` with `odk2`.

lognaturel:
> you first need to combine it with shared-src
Ah, that makes sense, thanks @jbeorse. It works if the shared files are already built, I suppose.

Updated to use the `make` commands. This is just a quick memory jog for people who are already working with docs. The [contribution guide](http://docs.opendatakit.org/contributing/) will go into more detail.

PR #585, opened 2018-03-20
Distinguish ODK1 and ODK2 style

jbeorse:
addresses https://github.com/opendatakit/docs/pull/480#issuecomment-372427994

Creates slightly different styling for ODK1 and ODK 2 to provide a clear visual indication of which the reader is viewing.

Moves the custom.css files from the shared-src directory to each of the odkX-src directories. The ODK2 version is identical, but with the addition of background colors for wy-side-nav-search and wy-side-scroll. The search is changed from blue to green as a relatively arbitrary decision, but it fit with the original comment and the logo has both colors, so it seemed appropriate. The navigation portion (wy-side-scroll) is also darkened a bit, also completely arbitrarily. I'm open to adjusting these colors or adding other style changes if anyone with a more artistic eye has an opinion.


wbrunette:
The green looks fine. When you shrink the window size the side bar moves along the top it is still blue. That probably should change to green.

jbeorse:
Good catch. I've added a rule for wy-nav-top. 

Is the green sufficient? Do we need the color change on the side scroll bar?

wbrunette:
IMHO I think the green is sufficient.

jbeorse:
After looking at it for a while I agree. I've gone ahead and removed that shade change. I think if want a shade change there it should probably be more dramatic, but as the green is ever-present, I think its pretty clear that you're looking at different docs

adammichaelwood:
This looks good. I think there's probably more we can do in the future to update the design of both sites a bit, but this serves the immediate purpose of distinguishing between the two.

PR #604, opened 2018-03-23
remove shared and py from intersphinx

adammichaelwood:
`shared` is not a sphinx site --- it gets pulled into odk1 and odk2 at build time
so we shouldn't include it as an intersphinx mapping

Additionally, I commented out the python mapping,
as I don't see (currently) as reason for us to be referring to it.

### DIFF
--- a/odk1-src/conf.py
+++ b/odk1-src/conf.py
@@ -226,8 +226,7 @@ epub_exclude_files = ['search.html']
 
 # Example configuration for intersphinx: refer to the Python standard library.
 intersphinx_mapping = {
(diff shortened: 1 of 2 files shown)


yanokwa:
@linl33 Do you agree?

cc @jbeorse @wbrunette

jbeorse:
I have no strong opinion about this and defer to @linl33 and @wbrunette

linl33:
I don't have strong opinion on this either. I put it there just so that when the shared site is ready, we won't need to update the mappings.

adammichaelwood:
There is no current plan to have a third "shared" site.

PR #609, opened 2018-03-23
remove agg release notes

### LINKED ISSUE #302: Should we have an Upgrade Note admonition?
#297 added a new admonition style for Upgrade Notes.

I'm not sure this is a good idea, but I merged the PR because the information was good and I wanted it to get published.

FIRST - I am going to let @ankita240796 make a case for adding it.
THEN - I will mention my concerns.
AND THEN - others may comment.

Please note:
 
 - Do not comment until @ankita240796 and @adammichaelwood (me) have both commented.
 - This is not a democracy. The point of a discussion is to arrive at a good decision, not to vote on a course of action.
 - This is also not a dictatorship. You can disagree with me.

### LINKED ISSUE #608: Remove Aggregate release notes from Docs
@yanokwa 

We have:
https://docs.opendatakit.org/aggregate-upgrade/#release-notes
which should go.

What do you think of:
https://docs.opendatakit.org/aggregate-limitations/#issues-with-older-version-of-aggregate

adammichaelwood:
closes #608 
closes #302 

The entire release notes section is removed from aggregate-upgrade,
also, reference to it in the preceding upgrade instructions was removed.

Removes CSS for styling "Upgrade Notes"

Do you want to also pull out https://docs.opendatakit.org/aggregate-limitations/#issues-with-older-version-of-aggregate ?

## other notes

The upgrade aggregate page probably needs a wholesale rewrite. But that requires having some good understanding of what the current install landscape is... do we still have a lot of old agg installs in the wild? Do we need to support an upgrade path for them? (Or can we just say: export to briefcase, start over, import old data)


odk1-src/aggregate-upgrade.rst
```diff
@@ -109,724 +109,19 @@ If you are unable to log onto your server, you will need to search for the versi
...
-3. Search upward in the :ref:`release notes <release-notes>` below, starting from your current ODK Aggregate version to locate the most recent ODK Aggregate version that does not require any manual upgrade steps. 
-4. Log onto your server to confirm that it is still functioning.
-5. Download the next ODK Aggregate version, perform the manual upgrade steps if any and upgrade to that version (and perform any post-install upgrade steps, if any).
-6. Log onto your server to confirm that it is still functioning.
-7. Repeat the steps 3-6 until you have upgraded to the current version.
-8. Enable submissions to ODK Aggregate via the :guilabel:`Form Management` tab.
+#. Disable all submissions to ODK Aggregate, in the :guilabel:`Form Management` tab.
+#. Use :doc:`ODK Briefcase  <briefcase-using>` to pull a copy of all data to your computer. 
+#. Log onto your server to confirm that it is still functioning.
+#. Download the next ODK Aggregate version and upgrade to that version.
```
yanokwa: Maybe we should add a note why we say this. Something like: "Do not simply upgrade to the latest version of Aggregate. We recommend you incrementally upgrade because Aggregate releases may require manual changes to your database"

yanokwa: I would also add a tip (or maybe a new section) on how to determine your Aggregate version. https://github.com/opendatakit/opendatakit/wiki/Aggregate-Release-Notes#determining-your-odk-aggregate-version has copy you can start from.

yanokwa:
I've put each version's release notes at https://github.com/opendatakit/aggregate/releases. I got that copy from https://github.com/opendatakit/opendatakit/wiki/Aggregate-Release-Notes. It looks like v1.3.x were never tagged.

We do have a lot of old installs out in the wild. I think the general upgrade instructions are good enough to get people to upgrade. Of course, the binaries aren't on Github, but I've filed an issue for that https://github.com/opendatakit/aggregate/issues/238.

odk1-src/aggregate-upgrade.rst
yanokwa: I guess we can also add a tip that says if you want to setup a new instance, you can move using  Briefcase. The caveat is that you will lose your URL.

adammichaelwood: > lose your URL

Only if you use the assigned url from your cloud provider, right?
If you've mapped a human-friendly URL to it, you can just move that to the new install right?

yanokwa: Most people on App Engine don't have a URL they can remap, but yeah, others can.

PR #610, opened 2018-03-23
adds note about time zone issue for time widget

### LINKED ISSUE #283: Limitations of time widget
The time widget doesn't work well across daylight savings time boundaries because it does account for timezones but doesn't know what day the time was saved. For example, if a time is saved in summer time and then the form is opened in winter time, the time be shown as off by an hour. However, the underlying value will be correct.

See opendatakit/collect#170 for more details

adammichaelwood:
closes #283 

A note in form-widgets (Question Types) about the effect of time zone and DST.

I don't really have a suggested workaround for this problem.

### DIFF
--- a/odk1-src/form-widgets.rst
+++ b/odk1-src/form-widgets.rst
@@ -479,6 +479,27 @@ A time selector. Captures only a specific time-of-day, not a date and time. For 
 
 The time widget does not accept any :th:`appearance` attributes.
 
+.. note::
+
+  The time widget stores the time along with a time zone.
+  This can cause unexpected behavior around `Daylight saving time`_.
+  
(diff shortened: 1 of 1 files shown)


odk1-src/form-widgets.rst
```diff
@@ -479,6 +479,27 @@ A time selector. Captures only a specific time-of-day, not a date and time. For
...
+  and then view the time after the clock change,
+  it will appear to be an hour off.
+  This happens because the recorded time data
+  is understood as a specific moment in time
+  that is being "translated" into your current, local time zone.
+  
+  A similar problem occurs when moving between geographic time zones.
+  
+  This makes the time widget unsuitable for abstract 
+  time-of-day questions such as *What time do you usually wake up?*
```
yanokwa: One good workaround here is to use a select one or integer for hours and another for minutes.

PR #612, opened 2018-03-24
Edits getting started

### LINKED ISSUE #455: Edit Getting Started Guide
- Procedures are ULs instead of OLs
 - more links

Part of #402

adammichaelwood:
closes #455 

Moved the section about how ODK gets used to the home page,
and cleaned up Getting Started for style, better internal links, and clearer procedure.

### DIFF
--- a/odk1-src/getting-started.rst
+++ b/odk1-src/getting-started.rst
@@ -1,33 +1,7 @@
 Getting Started With ODK
 =========================
 
-.. _using-odk:
 
-Using ODK
------------
-
-In most cases, users of ODK:
-
-- Create survey forms using `Build <https://build.opendatakit.org/>`_ or `XLSForm <http://xlsform.org/>`_
-- Upload forms to an Aggregate server
-- Load forms into Collect on an Android device
-- Use Collect to fill out forms with :term:`participants <participant>`
-- Upload survey data from Collect to Aggregate
-- Analyze or export data in Aggregate
-
-This requires:
-
-- :doc:`Installing Collect on a phone or other mobile device <collect-install>`
-- :ref:`Installing Aggregate on a server <installing-aggregate>`
-
-.. tip::
-
-  While this is the *typical* use pattern, it is not the only way to do things. For example:
-
-  - Forms can be created using tools other than Build or XLSForm
-  - Forms can be transfered directly to Collect, without using Aggregate
-
-  ODK is a very flexible set of tools, and organizations will find their own best practices for adopting it.
 
 .. _install-collect:
 
--- a/odk1-src/getting-started.rst
+++ b/odk1-src/getting-started.rst
@@ -29,9 +29,9 @@ The easiest way to set up Aggregate is to
 You'll set up a new Google Cloud project, and then run the install utility locally. This will connect to your Google Cloud account and install Aggregate there.
 
 Alternatively, if you just want to try things out,
-you can use the `Aggregate demo server`_.
+you can use the `Aggregate sandbox server`_.
 
-.. _Aggregate demo server: https://opendatakit.appspot.com
+.. _Aggregate sandbox server: https://sandbox.aggregate.opendatakit.org:8443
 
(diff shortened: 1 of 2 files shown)


odk1-src/getting-started.rst
```diff
@@ -1,90 +1,81 @@
...
 
-For other installation options, see :doc:`collect-install`.
+.. seealso:: :doc:`collect-install`
 
-.. _installing-aggregate:
+.. _getting-started-install-aggregate:
 
-Install Aggregate
----------------------
+Install Aggregate (optional)
```
yanokwa: If you don't install Aggregate, you can't do the things in this list (e.g., create a form with Build and upload it to Aggregate and so on. 

I think maybe we should have different levels? So Level 1 would be to install Collect, fill out a form, and see it on Aggregate. Then Level 2 would be to a form, upload it to a server, download it to Collect, fill it out, and see it on Aggregate.

The flaw in this plan is that it's not so easy to install Aggregate. And even with the VM it's not that fun to get Collect/Aggregate integration. So maybe we need to have a sandbox server that we delete data from nightly?

yanokwa:
@adammichaelwood I've set up a sandbox at https://sandbox.aggregate.opendatakit.org:8443. It needs some spit and polish, but should at least let people try stuff.

adammichaelwood:
When I try the sandbox link, I get a time out error.

adammichaelwood:
I've updated this to link to the sandbox server. 
Do you think it should be broken out even more (as suggested in your previous comment) or does this work?

odk1-src/getting-started.rst
```diff
@@ -1,90 +1,88 @@
...
-Install Collect
----------------------
+The easiest way to set up Aggregate is to
+:doc:`install it on Google App engine <aggregate-app-engine>`. 
 
-The easiest way to install the ODK Collect App is `to get it from the Google Play store <https://play.google.com/store/apps/details?id=org.odk.collect.android&hl=en>`_.
+You'll set up a new Google Cloud project, and then run the install utility locally. This will connect to your Google Cloud account and install Aggregate there.
 
-For other installation options, see :doc:`collect-install`.
+Alternatively, if you just want to try things out,
```
yanokwa: s/just/g

odk1-src/getting-started.rst
```diff
@@ -1,90 +1,88 @@
...
+Install Aggregate (optional)
+------------------------------
 
-Install Collect
----------------------
+The easiest way to set up Aggregate is to
+:doc:`install it on Google App engine <aggregate-app-engine>`. 
 
-The easiest way to install the ODK Collect App is `to get it from the Google Play store <https://play.google.com/store/apps/details?id=org.odk.collect.android&hl=en>`_.
+You'll set up a new Google Cloud project, and then run the install utility locally. This will connect to your Google Cloud account and install Aggregate there.
```
yanokwa: s/install utility/installer/

odk1-src/getting-started.rst
```diff
@@ -1,90 +1,88 @@
...
 
-For other installation options, see :doc:`collect-install`.
+Alternatively, if you just want to try things out,
+you can use the `Aggregate sandbox server`_.
 
-.. _installing-aggregate:
+.. _Aggregate sandbox server: https://sandbox.aggregate.opendatakit.org:8443
```
yanokwa: We should add a warning that `This is a demonstration server. All forms and data on this server are public and are occasionally deleted.`

odk1-src/getting-started.rst
adammichaelwood: What about "only".

yanokwa: Only works. Just is bad because it makes it sound easy and it might not be easy for someone.

adammichaelwood: Right.

adammichaelwood:
> Shameful delay in reviewing, but it's done.

Let he who has done any actual work in the last two weeks,
cast the first stone.

Issue #616, opened 2018-03-26
Form Logic document


adammichaelwood:
Basically an overview of dynamic form behavior, as a companion to Question Types.
Will cover calculation, substitutions, repeats, relevance, constraints.

adammichaelwood:
I think #630 covers it.

PR #630, opened 2018-04-05
form logic work

adammichaelwood:
addresses #616 

This is currently kind of a mess, and has some half-started notes here and there, and comments to myself. And build errors. But, it would be helpful to get in-progress thoughts on where this doc is going.

I haven't updated the TOC, so you'll just have to navigate to `/form-logic` and `/form-repeats` in the browser nav bar. Even with the build errors, you should be able to build and serve locally.


odk1-src/form-logic.rst
```diff
@@ -0,0 +1,314 @@
...
+:tc:`not(contains(., 'prohibited'))`
+  True if the substring ``prohibited`` does not appear in the response.
+
+  
+.. seealso:: :doc:`form-regex`
+
+
+.. _repeats:
+
+Repeating questions and groups of questions
```
akshay-ap: A separate page is available for repeats, how about moving this content there. Also, this content is a duplicate of http://xlsform.org/#repeats. How about redirecting the user to it and showing unique content in these docs?

odk1-src/form-logic.rst
```diff
@@ -0,0 +1,314 @@
+***********
+Form Logic
+***********
+
+Form logic building blocks
+============================
```
akshay-ap: How about adding numbering? It will help to understand the beginning of blocks.

odk1-src/form-logic.rst
```diff
@@ -0,0 +1,314 @@
...
+  note, tip_18_note, | "Bill: $${bill_amount_1}
+  | Tip (18%): $${tip_18}
+  | Total: $${tip_18_total}",
+
+
+
+
+.. _requiring_responses:
+
+Requiring responses
```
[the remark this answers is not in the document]
akshay-ap: Also, the text size seems to be bigger than other blocks. Would be better to keep consistent text size for all titles of blocks.

odk1-src/form-logic.rst
```diff
@@ -0,0 +1,314 @@
...
+is evaluated dynamically as a form is filled out.
+It can include `XPath functions`_, :ref:`operators`,
+:ref:`values from previous responses <previous-responses>`,
+and (in some cases) :ref:`the value of the current response <current-value>`. 
+
+.. _XPath functions: https://opendatakit.github.io/xforms-spec/#xpath-functions
+
+.. _formula: http://xlsform.org/#formulas
+
+.. rubric:: Example expressions
```
akshay-ap: How about adding links to complete forms here?

odk1-src/form-logic.rst
```diff
@@ -0,0 +1,314 @@
...
+  Evaluates to ``True`` or ``False``, 
+  depending on the value of :tc:`age`.
+
+:tc:`round(${bill_amount} * ${tip_percent} * 0.01, 2)`
+  Calculates a tip amount based on two previously entered values,
+  and then rounds the result to two decimal places.
+
+Expressions are used in:
+
+ - :ref:`calculations`
```
akshay-ap: It doesn't show a clickable link.

odk1-src/form-logic.rst
```diff
@@ -0,0 +1,314 @@
...
+.. rubric:: XLSForm
+
+.. csv-table:: survey
+  :header: type, name, label
+  
+  text, your_name, What is your name?
+  note, hello_name, "Hello, ${your_name}."
+
+ 
+.. _calcuations:
```
akshay-ap: Contains spelling mistake due to which its above `ref` is not clickable.

odk1-src/form-logic.rst
```diff
@@ -0,0 +1,314 @@
...
+.. _variables:
+
+Variables
+-----------
+
+Variables reference the value of previously answered questions.
+To use a variable,
+put the question's :th:`name` in curly brackets preceded by a dollar sign:
+
+:tc:`${question-name}`
```
akshay-ap: Can be bold as seen in `Expressions`

odk1-src/form-logic.rst
```diff
@@ -0,0 +1,314 @@
+***********
+Form Logic
```
akshay-ap: How about notifying the user to refer http://xlsform.org first?

yanokwa: Agreed! The form logic section is currently tightly focused on XLSForm, so we should say that up front. We have a similar language in form-widgets that we can use.

odk1-src/form-logic.rst
yanokwa: I think it's fine to link to xlsform.org for another example, but repeats are such a frequently asked about item, I think we should keep our own content.

odk1-src/form-logic.rst
```diff
@@ -0,0 +1,314 @@
...
+
+ - :ref:`calculations`
+ - :ref:`constraints`
+ - :ref:`relevants`
+
+
+
+.. _variables:
+
+Variables
```
yanokwa: I think the simplest items should go first. So variables, then expressions, then calculations.

yanokwa:
It'd be nice to see a little bit of prose at the beginning about what we are trying to accomplish in the document. This is important because we generally link directly to the docs and some opening context (e.g., adding form logic like branching and variables helps improve data quality) would be nice.

odk1-src/form-logic.rst
yanokwa: Maybe the /form-repeats should be retitled something like /form-repeat-recipes or something along those lines? Because right now we have some good repeat content in /form-logic already.

adammichaelwood: That's literally what I called it at first. Then I thought it was basically an expansion of the repeat section of form-logic, and changed it (I also wasn't sure if "Recipes" was too informal or wouldn't translate well)

adammichaelwood: my intention was to cover the basics of everything in form logic and then add additional pages for specific topics as needed,

yanokwa: That seems like a reasonable approach! Because we could have form topics on repeats, media, styling, etc.

adammichaelwood:
This is ready for some eyes again, @yanokwa and @akshay-ap 
I have:

 - finished the sections I just had quick notes for
 - reorganized a few things
 - eliminated all the sphinx warnings
 - re-worked the form section's TOC

To do later today:

 - screenshots (I have indicated a few places I think i should add them. IF you have suggestions where you think they'd be helpful, make a note)
 - style/spell/typo checking (feel free to point it out, but I haven't even done a quick pass yet, so there's probably a bunch)
 - remove form-interaction (this is mostly now duplicated, but I need to check through each point to make sure the info isn't lost)

If there are any important form logic concepts I missed, please let me know so I can add them.

(Form styling will be another PR.)

odk1-src/form-logic.rst
```diff
@@ -0,0 +1,456 @@
...
+  integer, number_of_children, How many children do you have?
+  begin_repeat, child_questions, Questions about child, ${number_of_children}
+  text, child_name, Child's name,
+  integer, child_age, Child's age,
+  end_repeat, , , 
+
+  
+.. _cascading-selects:
+  
+Filtering options based on previous responses
```
yanokwa: I don't think the `based on previous responses` is needed here because validating and restricting options (one of the previous sections) also has the property.

odk1-src/form-logic.rst
```diff
@@ -0,0 +1,456 @@
...
+  job_titles, asst, Office Assistant, admin
+  job_titles, mngr, Office Manager, admin
+  job_titles, scheduler, Scheduler, admin
+  job_titles, reception, Receptionist, admin
+  job_titles, creative_dir, Creative Director, marketing
+  job_titles, print_design, Print Designer, marketing
+  job_titles, ad_buyer, Ad Buyer, marketing
+  job_titles, copywriter, Copywriter, marketing
+
+.. _launch-apps-from-collect:
```
yanokwa: Feels very out of place. I would move this back. And It's probably sufficient to drop the "Forms" from the title.

adammichaelwood:
This should be ready for review.

odk1-src/form-tools.rst
```diff
@@ -0,0 +1,9 @@
+Form Creation Tools
```
yanokwa: We call the section form building and use the "building" term elsewhere. Seems like Form Building Tools would be more consistent.

adammichaelwood:
> This is looking good. My understanding was that you'd be review form-interaction, make sure the information was captured in form-logic and remove form-interaction. I don't see that change in this PR.

Will do.

PR #632, opened 2018-04-06
Update readme

### LINKED ISSUE #631: Add building requirements & Troubleshooting to README
Add some details in `README.md` such as:
- Python packages requirement
- Troubleshooting about setting up environment

huangyz0918:
closes #631 

I think people who wants to contribute **ODK Doc** may get confused when they get into troubles during setting up the building environment. We should put some details into `README.md`(e.g, pip commands).

You can take a look at new [`README.md` ](https://github.com/huangyz0918/docs/blob/update-readme/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,14 +30,13 @@ $ python -m http.server 8000
 ```
 
 And we also have several `make` options you can choose.
-
-For both odk1 and odk2:
+For both ODK 1 and ODK 2:
 
 |          |    Build     |     Clean     |     Check     |
 | -------- | :---------: | :-----------: | :-----------------: |
 | **Options** | build-all |  clean |  check-all |
 
-For a specific odk:
+For a specific ODK version:
 
 |          |    Copy     |     LaTeX     |     Style Check     |    Spell Check     |    Check All   |
 | -------- | :---------: | :-----------: | :-----------------: | :----------------: | :----------------: |
--- a/README.md
+++ b/README.md
@@ -50,22 +50,7 @@ You can also...
 
 ## Troubleshooting
 
-### **1. command not found**
-#### Error message:
-```bash
-$ make odk1
-rm -rf tmp1-src
-rm -rf odk1-build/*
-mkdir tmp1-src
-cp -rf odk1-src/* tmp1-src
-cp -rf shared-src/* tmp1-src
-python: sphinx-build: command not found
-make: *** [odk1] Error 127
-```
-#### Reason & Fix:
-To build document, make sure you have installed `sphinx` in you current python version already. If you didn't install that, please using our `requirements.txt` file or manually.
-
-### **2. incorrect sphinx version**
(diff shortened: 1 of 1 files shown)


lognaturel:
Thanks for checking out the docs and making these recommendations, @huangyz0918! In general the intent is that someone new to docs contribution would go through the steps at http://docs.opendatakit.org/contributing. Have you looked at those? Perhaps we should start by making sure that's very clear.

I think adding a little bit more information and a troubleshooting section to the README is good but there's a bit too much here. For example, the requirements definitely shouldn't be there because those are likely to become out of date and don't provide actionable information. Adding `pip install -r requirements.txt` should be enough.

huangyz0918:
Yes, I have readed the Doc in detail. 
I agree with you to add just a little more information about setting up environment. It's tiny improvement but necessary.

README.md
```diff
@@ -40,3 +47,28 @@ You can also...
...
+...
+Configuration error:
+There is a programable error in your configuration file:
+...
+from sphinx.util.compat import Directive
+ImportError: cannot import name Directive
+```
+
+#### Reason & Fix:
+It caused by an incorrect` sphinx` version, and you can fix this by reinstalling `sphinx` using `requirements.txt`.
```
lognaturel: It *is* caused...

How about instead of "using `requirements.txt`", "using `pip install -r requirements.txt`" so it's even easier.

adammichaelwood:
I'm a little uncomfortable about providing `pip install` instructions without the context of a virtual environment or python versions or anything else.

lognaturel:
So maybe I misled @huangyz0918 and it's better to include some of that info in the README as well? The way I think of it, the README is a useful memory jog for advanced people who already have the basics set up. For example, I needed a reminder that I had to update requirements when some dependencies changed. For deeper questions or more handholding the complete guide is there.

lognaturel:
Sounds good to me! @huangyz0918 thanks for getting us to have this important conversation!

huangyz0918:
@adammichaelwood I think adding `make` options is great! I list several options, what do you think is needed ?

-` help`
- `odk1_clean`/`odk2_clean`
- `odk1_clean_files`/odk2_clean_files`/`clean`
- `odk1_copy`/`odk2_copy`
- `build-all`
- `odk1-latex`/`odk2-latex`
- `odk1-style-check`/`odk2-style-check`
- `odk1-spell-check`/`odk2-spell-check`
-  `odk1-check`/`odk2-check`/`check-all`

huangyz0918:
I created a table for these `make` options. It's ok for reviewing :)

haha, I have to say, for a Virgo, it is a really tired thing to modify `README.md`, because its appearance always does not make me satisfied. Anyway, it's a joke, I'm very happy to make this improvement (although it is a small one). If you think it still needs improvement, please tell me.

PR #642, opened 2018-04-10
renames form-widgets to form-question-types

adammichaelwood:

 - rename file
 - update refs throughout docs
 - add redirect
 - update TOCs

Some in progress PRs will need some cleanup.


adammichaelwood:
You know, we haven't really been making a practice of renaming image files along with document renaming. (See file name changes associated with Aggregate and Collect reorgs.) 

Maybe we should?

yanokwa:
I don't want to slow this PR down so I've filed another issue.

PR #649, opened 2018-04-11
Describe audit log structure and usage

### LINKED ISSUE #598: Enumerator audit logs
Form specification: https://opendatakit.github.io/xforms-spec/#metadata

XLSForm: 

| survey |       |       |
|--------|-------|-------|
| type   | name  | label |
| audit  | audit |       |

https://github.com/XLSForm/pyxform/pull/136/files

https://github.com/opendatakit/collect/pull/760 has sample forms and output

Sample form: [audit-test-orx.xml.txt](https://github.com/opendatakit/docs/files/1834783/audit-test-orx.xml.txt)

Note that Aggregate 1.5+ is needed or **the attachments are not saved**

lognaturel:
Closes #598 

Feedback wanted on:
- Content: is there enough detail, too much? Is anything missing?
- Formatting: @adammichaelwood you'll probably want to make sure I've used the proper roles and directives


odk1-src/form-audit-log.rst
```diff
@@ -0,0 +1,146 @@
...
++-------------------+------------------------------------------------------------------+-------+----------------+
+| constraint error  | Constraint or required error on finalize                         | No    | ``start`` only |
++-------------------+------------------------------------------------------------------+-------+----------------+
+
+.. _audit-timestamps:
+
+Timestamps
+-----------
+
+Device time can be changed by the user unless Android access controls are put in place. To reduce the ways that audits can be manipulated by modifying device time, clock time is collected once at form launch and all following times are computed relative to that initial reading. This means that while the actual time may potentially be inaccurate, the time elapsed within and between screens is always accurate within one form editing session.
```
yanokwa: s/clock time/device time/

odk1-src/form-audit-log.rst
```diff
@@ -0,0 +1,146 @@
...
+Values in the ``start`` and ``end`` columns are timestamps represented as the number of milliseconds since midnight, January 1, 1970 UTC. This is known as epoch time and provides a standard way of representing date/time even across timezones. The :ref:`audit-timestamps` section contains more information about timestamps.
+
+.. _audit-event-types:
+
+Event types
+--------------
+
+The event column of the audit log can have the following values:
+
++-------------------+------------------------------------------------------------------+-------+----------------+
```
yanokwa: I'm not so sure about the format you use for `start`` here in the timestamps column. Perhaps @adammichaelwood has thoughts on if this is correct.

odk1-src/form-audit-log.rst
```diff
@@ -0,0 +1,146 @@
...
+
+The enumerator spent ``1488761809157 - 1488761807868 = 1289`` milliseconds on the screen showing the ``/data/name`` question. This corresponds to ``1289 / 1000 = 1.289`` seconds.
+
+To convert from epoch time to time in UTC in most common spreadsheet programs, divide the epoch time by 86400000 ms per day and add 25569 days between January 1, 1900 (what most spreadsheet programs use as "day zero") and January 1, 1970. For example, to convert the timestamp ``1488761807868``:
+
+.. code-block:: xml
+
+  (1488761807868 / 86400000) + 25569 = 42800.03944
+
+When the cell is set to type ``date time`` in common spreadsheet programs, it will show ``3/6/2017 0:56:48 UTC``.
```
yanokwa: I think it should be DateTime and not Date Time.

odk1-src/form-audit-log.rst
```diff
@@ -0,0 +1,146 @@
+.. spelling::
+
+  timestamp
+
+Logging enumerator behavior
```
yanokwa: I would add a reference to /form-widgets/#metadata-widgets so people know that they can timestamp in this way.

odk1-src/form-audit-log.rst
adammichaelwood: So, if I was viewing this in Excel, I'd probably add a column of formula cells doing that conversion math, *and then* have that result viewed as `date/time`?

odk1-src/form-audit-log.rst
adammichaelwood: This paragraph seems a little unclear. I read it a few times to grok it, because at first, it sounded like the audit stored elapsed time for each event. But it doesn't (right?) --- it calculates the presumed actual time *using* elapsed time (instead of re-checking the device clock).

I don't have a brilliant suggestion for improving it yet...

odk1-src/form-audit-log.rst
adammichaelwood: I agree its weird. Typically the double-back-tick style is for literal strings (equivalent to `<code>`).

I might use the `:th:` text role, since you are referring back to info in the *start* column of the audit log.

adammichaelwood: I'm also curious why you used this markup style for the table. It's fine, but strikes me as harder to write than the csv table markup.

odk1-src/form-audit-log.rst
lognaturel: Do you mean a reference from there?

odk1-src/form-audit-log.rst
lognaturel: That seems likely. I can make that explicit.

odk1-src/form-audit-log.rst
lognaturel: Your interpretation is correct. Suggestions welcome and I'll try to play with it some more as well! The idea itself is tricky.

yanokwa: What about this...

> If we relied entirely on the time reported by the device for every event, users (or the network) could change the device time and manipulate the correctness of the audit log. For this reason, only use device time for the form start timestamp. All subsequent event timestamps are the result of elapsed time, which users cannot change, added to the form start timestamp. This means that while the timestamps themselves may potentially be inaccurate, the time elapsed within and between the timestamps are always accurate within one form editing session.

odk1-src/form-audit-log.rst
yanokwa: Yes a reference from metadata-widgets pointing to this document.

odk1-src/form-audit-log.rst
lognaturel: Sheets calls it date time and Numbers calls it Date & time. Seems neutral enough this way.

odk1-src/form-audit-log.rst
lognaturel: It didn't feel like a CSV! Do you use it for arbitrary tables? I used https://ozh.github.io/ascii-tables/ so it wasn't a big deal.

odk1-src/form-audit-log.rst
lognaturel: Thanks, @yanokwa, that is clearer!

odk1-src/form-audit-log.rst
adammichaelwood: I see what you mean, but yes --- I use it for arbitrary tables because I find the markup format to just be easier to read, type, and edit in source.

lognaturel:
Changes made!

odk1-src/form-audit-log.rst
```diff
@@ -0,0 +1,146 @@
...
+
+.. image:: /img/form-audit-log/audit-media-icon.png
+  :alt: The Aggregate submissions page with a form that has an audit log. The media icon in the meta audit column is circled.
+
+This displays a popup with the audit contents:
+
+.. image:: /img/form-audit-log/audit-example.png
+  :alt: An example audit log in Aggregate.
+
+If more sophisticated analysis of the logs is required, they can be downloaded along with their submissions using :ref:`Briefcase <pull-from-aggregate>`.
```
yanokwa: ...is required, logs (not they), can be downloaded...

odk1-src/form-audit-log.rst
```diff
@@ -0,0 +1,146 @@
...
+---------------
+
+If a form includes an audit, Collect will create an ``audit.csv`` file as the form is filled out. The ``audit.csv`` file has the following structure:
+
+.. csv-table:: audit.csv
+  :header: event, node, start, end
+
+  question, /data/name, 1523403169208, 1523403170733
+
+Values in the :th:`event` column represent a particular user action such as opening a form, saving a form or displaying a question. Possible event types are described in the :ref:`audit-event-types` section.
```
yanokwa: saving a form COMMA or display a question. Because Oxford comma.

odk1-src/form-audit-log.rst
```diff
@@ -0,0 +1,146 @@
...
++-------------------+------------------------------------------------------------------+-------+-----------------+
+| constraint error  | Constraint or required error on finalize                         | No    | :th:`start` only|
++-------------------+------------------------------------------------------------------+-------+-----------------+
+
+.. _audit-timestamps:
+
+Timestamps
+-----------
+
+If we relied entirely on the time reported by the device for timestamps, users or the network could change the device time and manipulate the correctness of the audit log. For this reason, only use device time for the form start timestamp. All subsequent event timestamps are the result of elapsed time, which users cannot change, added to the form start timestamp. This means that while the timestamps themselves may potentially be inaccurate, the time elapsed within and between the timestamps are always accurate within one form editing session.
```
yanokwa: For this reason, WE only use

odk1-src/form-audit-log.rst
```diff
@@ -0,0 +1,146 @@
...
+
+.. csv-table:: audit.csv
+  :header: event, node, start, end
+
+  form start, , 1488761807863, 
+  question, /data/name, 1488761807868, 1488761809157
+
+The enumerator spent ``1488761809157 - 1488761807868 = 1289`` milliseconds on the screen showing the ``/data/name`` question. This corresponds to ``1289 / 1000 = 1.289`` seconds.
+
+To convert from epoch time to time in UTC in most common spreadsheet programs, divide the epoch time by 86400000 ms per day and add 25569 days between January 1, 1900 (what most spreadsheet programs use as "day zero") and January 1, 1970. For example, to convert the timestamp ``1488761807868``:
```
yanokwa: s/most//g. The most is redundant.

odk1-src/form-question-types.rst
```diff
@@ -1782,7 +1782,9 @@ Metadata widgets
 
 Metadata questions capture information about 
 the device or the survey collection event
-and are not visible to the user.
+and are not visible to the user. 
+A more complete log of user behavior within a form
```
yanokwa: A more complete RECORD (sounds better to me than log). Too much logs.

lognaturel:
Changes made. I would probably squash the commits on merge for this one.

PR #653, opened 2018-04-14
Redirect form-interaction to form-logic

yanokwa:
Addresses #630

Since we removed form-interaction, we should redirect it to form-logic (which is some fantastic content, BTW). I think this is how we do redirects now.

I think @adammichaelwood has a magical script that was supposed to prevent this mistake from happening. Maybe. It'd be good to know where the knowledge gap is.


adammichaelwood:
> I think @adammichaelwood has a magical script that was supposed to prevent this mistake from happening. Maybe. It'd be good to know where the knowledge gap is.

I didn't think of doing it because the move wasn't 1-to-1 (that is, it wan't a rename). But I should have thought of it.

Thanks.

PR #660, opened 2018-04-17
Form functions

adammichaelwood:

 - Function and operator reference
 - Reusable form examples pulled into includes
 - conf.py updated to use JS domain by default (See below)
 - custom css to make function reference easier to read
 - fixed a few form examples (form-logic and form-repeats) that were either wrong or had typos

#### What issues need to be opened?

Opening an ongoing issue to add useful examples.

##### Minor - Hard to follow existing references

A few of the functions listed in the XForm spec weren't clear, and a few weren't relevant --- tracking that down was a minor pain.

cf: https://opendatakit.github.io/xforms-spec/#xpath-functions

##### Sort of major - Domain work

In Sphinx, function references like this are done using language specific domains.
 cf: http://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html

There is no XForm/XPath domain. I tried using the default domain (Python), but the syntax didn't work out. (Python doesn't allow hyphens, and does namespacing differently.) Then I tried to write my own domain, using the Python domain as a starting point. This was a terrible mistake.

Finally, I discovered that the JS domain was *good enough* for current purposes. So `conf.py` has been updated to use that as the default. It might become a problem in the future, depending on how much more XForm/Xpath stuff makes it into the docs. Also, it may become an issue if we decide to pull the ODK XForm spec into Sphinx.

##### Sidenote problem -- git rebase

I did the following:

 - rebased all my work on top of a fresh pull of master
 - fixed conflicts
 - commited locally
 - pushed to GH (--force)

And I *still* had to come fix a merge conflict (The same one) here in GH. There's something in this rebase process I'm missing. It wasn't a big deal, but it's a little annoying.

### DIFF
--- a/odk1-src/form-functions.rst
+++ b/odk1-src/form-functions.rst
@@ -1,0 +1,674 @@
+.. spelling::
+
+  Anaranjado
+  Azul
+  Colores
+  Español
+  Púrpura
+  Qué
+  Rojo
+  Seleccione
+  arctan
+  arg
+  bday
+  bp
+  br
+  chkn
+  colores
+  fsh
+  gustan
+  jr
+  nodeset
+  prefs
+  seleccionados
+  te
+  tres
+  π
+  
+  
+	
+******************************
+Form Functions and Operators
+******************************
+
+:ref:`expressions` in :ref:`calculations`, :ref:`constraints <constraints>`, and :ref:`relevants <relevants>`
+can contain `XPath functions`_ and operators.
+
+.. _XPath functions: https://opendatakit.github.io/xforms-spec/#xpath-functions
+
+.. contents::
+  :local:   
+
+.. _functions:
+  
+Functions
+===========
+    
+.. _control-flow-functions:
+
+Control flow
+--------------
+
+.. function:: if(expression, then, else)
+
+  Returns :arg:`then` if :arg:`expression` evaluates to ``True``. 
+  Otherwise, returns :arg:`else`.
+
+  
+.. function:: not(arg)
+
+  Returns the opposite of :func:`boolean(arg) <boolean>`.
+
+  
+.. function:: coalesce(arg, arg) 	
+
+  Returns first non-empty value of the two :arg:`arg`\ s.
+  Returns an empty string if both are empty or non-existent.
+
+  
+.. function:: true()
+
+  Evaluates to ``True``.
+
+.. function:: false()
+
+  Evaluates to ``False``.
+  
+.. function:: boolean(arg) 
+
+  Returns ``True`` if :arg:`arg` is:
+  
+  - a number other than zero
+  - a non-empty string
+  - a non-empty collection
+  - a comparison or expressions that evaluates to ``True``.
+   
+  Returns ``False`` if :arg:`arg` is:
+  
+  - the number 0
+  - an empty string
+  - an empty collection
+  - a comparison or expression that evaluates to ``False``.
+
+.. _response-access-functions:
+  
+Accessing response values
+--------------------------
+
+.. note::
+
+  The response from most question types
+  can be accessed using :ref:`variables <variables>`.
+  Functions are needed for accessing responses to 
+  :ref:`multi select questions <select-functions>` and
+  questions inside :ref:`repeat groups <repeat-functions>`.
+
+.. _select-functions:
+  
+Select questions
+~~~~~~~~~~~~~~~~~~~
+
+.. function:: selected(select_question, choice_name)
+
+  Returns ``True`` if :arg:`choice_name` 
+  was selected in :arg:`select_question`,
+  otherwise returns ``False``.
+  
+  .. include:: incl/form-examples/constraint-on-selected.rst
+  
+.. function:: selected-at(multi_select_question, n)
+
+  Returns the :th:`name` of the :arg:`n`\ :sup:`th` selected choice of the :arg:`multi_select_question`. (Selected choices are zero-indexed.)
+  
+  .. note::
+  
+    This function returns the :th:`name`, not the :th:`label`,
+    of the selected choice.
+    To get the label in the current language,
+    use :func:`jr:choice-name`.
+  
+  Returns an empty string if the index does not exist.
+  
+  .. image:: /img/form-functions/selected-at-0.* 
+    :alt: A multi-select widget in Collect. The label is "What colors do you like?" Several color names are presented as options. Red, Green, and Purple are selected.
+  
+  .. image:: /img/form-functions/selected-at-1.* 
+    :alt: A note widget in Collect. The label is "Selected Colors". The hint text is "red, green, purple".
+  
+  .. rubric:: XLSForm
+  
+  .. csv-table:: survey
+    :header: type, name, label, hint, calculation
+  
+    select_multiple colors, color_prefs, What colors do you like?, Select three.
+    calculate, color_0, , ,"selected-at(${color_prefs}, 0)"
+    calculate, color_1, , ,"selected-at(${color_prefs}, 1)"
+    calculate, color_2, , ,"selected-at(${color_prefs}, 2)"
+    note, color_note, Selected colors:, ${color_0} <br> ${color_1} <br> ${color_2}  
+
+  .. csv-table:: choices
+    :header: list_name, name, label
+    
+    colors, red, Red
+    colors, blue, Blue
+    colors, yellow, Yellow
+    colors, green, Green
+    colors, orange, Orange
+    colors, purple, Purple
+
+.. function:: count-selected(multi_select_question)
+
+  Returns the number of choices selected in ``multi_select_question``.
+  
+  .. image:: /img/form-functions/count-selected-constraint.* 
+    :alt: A multi-select widget in Collect. The label is "What colors do you like?" The hint text is "Select three." Four colors are selected. A message modal overlays the widget with the text "Select exactly three."
+  
+  .. rubric:: XLSForm
+  
+  .. csv-table:: survey
+    :header: type, name, label, hint, constraint, constraint_message
+    
+    select_multiple colors, color_prefs, What colors do you like?, Select three., count-selected(.)=3, Select exactly three.
+    
+  .. csv-table:: choices
+    :header: list_name, name, label
+    
+    colors, red, Red
+    colors, blue, Blue
+    colors, yellow, Yellow
+    colors, green, Green
+    colors, orange, Orange
+    colors, purple, Purple
+
+.. function:: jr:choice-name(choice_name, 'select_question')
+
+  Returns the label value, in the active language, associated with the :arg:`choice_name` in the list of choices for the :arg:`select_question`.
+  
+  .. note::
+  
+    You have to wrap the :arg:`select_question` reference in quotes.
+    
+    .. code-block:: none
+    
+      '${question_name}'
+  
+  .. image:: /img/form-functions/choice-name-multi-lang-english-0.* 
+    :alt: A multi-select widget in Collect. The label is "What colors do you like?" Several color names are presented as options. Red, Green, and Purple are selected.
+  
+  .. image:: /img/form-functions/choice-name-multi-lang-english-1.* 
+    :alt: A note widget in Collect. The label is "Selected colors". The hint text is "Red, Green, Purple."
+  
+  .. image:: /img/form-functions/choice-name-multi-lang-spanish-0.* 
+    :alt: A multi-select widget in Collect. The label is "¿Qué colores te gustan?" Several color names, in Spanish, are presented as options. Rojo, Verde, and Púrpura are selected.
+    
+  .. image:: /img/form-functions/choice-name-multi-lang-spanish-1.* 
+    :alt: A note widget in Collect. The label is "Colores seleccionados." The hint text is "Rojo, Verde, Púrpura".
+    
+  .. rubric:: XLSForm
+  
+  .. csv-table::  survey
+    :header: type, name, label::English, label::Español, hint::English, hint:Español, calculation
+    
+    select_multiple colors, color_prefs, What colors do you like?, ¿Qué colores te gustan?, Select three., Seleccione tres.
+    calculate, color_0, , , , ,"jr:choice-name( selected-at(${color_prefs}, 0), '${color_prefs}')"
+    calculate, color_1, , , , ,"jr:choice-name( selected-at(${color_prefs}, 2), '${color_prefs}')"
+    calculate, color_2, , , , ,"jr:choice-name( selected-at(${color_prefs}, 2), '${color_prefs}')"
+    note, color_note, Selected colors:, Colores seleccionados:, ${color_0} <br> ${color_1} <br> ${color_2}, ${color_0} <br> ${color_1} <br> ${color_2}
+    
+  .. csv-table:: choices
+    :header: list_name, name, label::English, label::Español
+    
+    colors, red, Red, Rojo
+    colors, blue, Blue, Azul
+    colors, yellow, Yellow, Amarillo
+    colors, green, Green, Verde
+    colors, orange, Orange, Anaranjado
+    colors, purple, Purple, Púrpura
+  
+.. _repeat-functions:
+    
+Repeat groups
+~~~~~~~~~~~~~~~~
+
+.. admonition:: Helpful terms
+
+  .. glossary::
+    :sorted:
+
+    nodeset
+
+      A collection of response values. 
+
+      Outside a :ref:`repeat group <repeats>`, 
+      :ref:`referring to a question by name <variables>`
+      will return a nodeset containing all the responses to that question.
+
+        
+.. function:: indexed-repeat(name, group, i [, sub_grp, sub_i [, sub_sub_grp, sub_sub_i ]])
+
+  Returns the response value of question :arg:`name`
+  from the repeat-group :arg:`group`,
+  in iteration :arg:`i`.
+  
+  Nested repeat groups can be accessed 
+  using the :arg:`sub` and :arg:`sub_sub` parameters.
+
+  .. seealso:: :ref:`referencing-answers-in-repeats`
+
+  .. include:: incl/form-examples/parallel-repeat-groups.rst
+
+.. function:: count(nodeset)
+
+  Returns the number of items in :arg:`nodeset`. This can be used to count the number of repetitions in a :ref:`repeat group <repeats>`.
+
+  .. include:: incl/form-examples/parallel-repeat-groups.rst
+  
+.. function:: count-non-empty(nodeset)
+
+  Returns the number of non-empty members of :arg:`nodeset`.
+
+.. function:: sum(nodeset)
+
+  Returns the sum of the members of :arg:`nodeset`.
+  
+  Can be used to :ref:`tally responses to a repeated select question <counting-answers>`.
+  
+  .. include::  incl/form-examples/sum-to-count-responses.rst
+
+.. function:: max(nodeset)
+
+  Returns the largest member of :arg:`nodeset`.
+  
+  .. rubric:: XLSForm
+  
+  .. csv-table:: survey
+    :header: type, name, label, calculation
+    
+    begin_repeat, child_questions, Questions about child
+    text, child_name, Child's name
+    integer, child_age, Child's age
+    end_repeat
+    calculate, age_of_oldest_child, , max(${child_age})
+
+.. function:: min(nodeset)
+
+  Returns the smallest member of :arg:`nodeset`.
+  
+  .. rubric:: XLSForm
+  
+  .. csv-table:: survey
+    :header: type, name, label, calculation
+    
+    begin_repeat, child_questions, Questions about child
+    text, child_name, Child's name
+    integer, child_age, Child's age
+    end_repeat
+    calculate, age_of_youngest_child, , min(${child_age}) 
+  
+.. _string-functions:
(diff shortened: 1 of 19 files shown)


yanokwa:
This is a great start. I think it's not quite glanceable. I'd like to see a quick table up front sorted by name or organized by theme. It reads a little technical to me and so maybe an example for each might help? https://opendatakit.org/help/form-design/binding I think was pretty good at that.

adammichaelwood:
Still working on examples. 
But --- does this organization make sense?
Section naming here is weird and hard, so suggestion welcome.

adammichaelwood:
Oh, also...

I think this page should also have operators on it.
If so, does the main title change?

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,502 @@
...
+
+
+.. function:: regex(string, expression)
+
+  Returns ``True`` if ``string`` is an exact match for ``expression``.
+
+  .. seealso:: :doc:`form-regex`
+
+
+.. function:: contains(haystack, needle)
```
yanokwa: My guess is that this is a very European figure of speech. Maybe string/substring or string/keyword? What do other docs do?

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,502 @@
...
+  Returns ``then`` if ``expression`` evaluates to ``True``. 
+  Otherwise, returns ``else``.
+
+  
+.. function:: not(arg)
+
+  Returns the opposite of ``boolean(arg)``.
+
+  
+.. function:: coalesce(arg, arg)
```
yanokwa: This feels more like combining strings.

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,502 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~  
+  
+  
+.. function:: format-date(date, format)
+
+  Returns ``date`` as a string formatted as defined by ``format``.
+  
+  The following identifiers are used in the ``format`` string:
+  
+  ``%Y``
```
yanokwa: Can we put this stuff in a table? Current layout feels wasteful.

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,502 @@
...
+.. function:: position(xpath)
+
+  Returns an integer equal to the 1-indexed position of the current node
+  within the node defined by ``xpath``.
+  
+  Most often this is used in the form ``position(..)``
+  to identify the current iteration index.  
+  
+  
+String handling
```
yanokwa: Might be OK to just call this string. Since you don't call it "math and numbers handling"

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,502 @@
...
+.. function:: exp10(number arg)
+
+  Returns ``10^x``.
+
+.. function:: pi()
+
+  Returns an approximation of the mathematical constant π.
+
+
+Control flow
```
yanokwa: This feels out of place here. Maybe it should go first or second (after form data)

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,502 @@
...
+  Returns the response value of question ``name``
+  from the repeat-group ``group``,
+  in iteration ``i``.
+  
+  Nested repeat groups can be accessed 
+  using the ``sub`` and ``sub_sub`` parameters,
+  which can be extended to three levels of nesting.
+
+  
+Counting responses
```
yanokwa: Feels mathy to me.

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,502 @@
...
+.. function:: count(nodeset)
+
+  Returns the number of items in ``nodeset``. This can be used to count the number of repetitions in a :ref:`repeat group <repeats>`.
+
+.. function:: count-non-empty(nodeset)
+
+  Returns the number of non-empty members of ``nodeset``.
+
+ 
+Doing math with responses
```
yanokwa: Feels mathy to me. It even says math in the title :)

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,502 @@
...
+Accessing information about the form
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ 
+ 
+.. function:: jr:choice-name(select_question, choice_name)
+
+  Returns the label value, in the active language, associated with the ``choice_name`` in the list of choices for the ``select_question``.
+
+.. function:: jr:itext(name) 
+
+  Returns the itext in the active language for the named item.
```
yanokwa: The itext here seems self-referential. Maybe just text?

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,502 @@
...
+
+  Returns the sum of the members of ``nodeset``.
+
+.. function:: max(nodeset)
+
+  Returns the largest member of ``nodeset``.
+
+.. function:: min(nodeset)
+
+  Returns the smallest memberd of ``nodeset``.
```
yanokwa: members

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,502 @@
...
+
+.. contents::
+  :local:   
+
+
+  
+Form data 
+----------------
+
+Accessing previous responses
```
yanokwa: Feels like most of this has to do with selects, so why not have a selects section, then have just a repeats section (even if it has one item). Then could put the mathy bits in the math section under their own subsection, move position and itext to utility.

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,502 @@
...
+.. function:: jr:itext(name) 
+
+  Returns the itext in the active language for the named item.
+  
+.. function:: position(xpath)
+
+  Returns an integer equal to the 1-indexed position of the current node
+  within the node defined by ``xpath``.
+  
+  Most often this is used in the form ``position(..)``
```
yanokwa: I love this. Helps the user understand the common case.

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,502 @@
...
+    select_multiple cancer_types, what_cancer, What type of cancer have you experienced?, Select all that apply.,"selected(${what_issues}, 'cancer')",
+  
+  .. seealso:: :ref:`The complete example of this form, with a video. <complex-conditional-example>`
+  
+  
+.. function:: selected-at(multi_select_question, n)
+
+  Returns the :th:`name` of the :arg:`n`\th selected choice of the :arg:`multi_select_question`. (Selected choices are zero-indexed.)
+  
+  .. note::
```
yanokwa: Good!

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,502 @@
+Form Functions
+================
+
+:ref:`expressions` in :ref:`calculations`, :ref:`constraints <constraints>`, and :ref:`relevants <relevants>`
```
yanokwa: Can we make calculations lower case in the rendered view? Looks out of place.

adammichaelwood:
> missing checklist(), weighted-checklist(), and property()

Yes, I couldn't figure out what those did when I was reading the XForm spec, 
but now I'm looking at the old ODK bindings page you linked,
and it is more clear there.

odk1-src/form-functions.rst
adammichaelwood: yeah, i wondered about that...

odk1-src/form-functions.rst
adammichaelwood: i like the semantics of definition lists, so maybe a compact view...

odk1-src/form-functions.rst
adammichaelwood: String/substring seems like a good alternative.

odk1-src/form-functions.rst
adammichaelwood: It would work on non-string values, though too --- right?

adammichaelwood:
see if that organizations structure makes more sense.

adammichaelwood:
This is (I hope) ready to go.
@yanokwa

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,674 @@
...
+---------------
+
+.. csv-table::
+  :header: , Explanation, Example, Notes
+  
+  \+, addition, ${salary_income} + ${self_employed_income}, Numbers only; does not concatenate strings.
+  \-, subtraction, ${income} - ${expenses},
+  \*, multiplication, ${bill} * 1.18,
+  div, division, ${percent_int} div 100, 
+  mod, `modulo`_ (division remainder), (. mod 2) = 0,
```
yanokwa: The use of `.` here sticks out.  I think something like `(15 mod 2) = 1` would be clearer.

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,674 @@
...
+
+  Returns ``True`` if 
+  the sum of the :arg:`weight`\ s 
+  of each :arg:`response` that is exactly the string "yes"
+  is between :arg:`min` and :arg:`max`, inclusive.
+  
+  Set :arg:`min` or :arg:`max` to ``-1`` to make the argument not 
+    
+
+.. _xform-operators:
```
yanokwa: I really like this section. Nice and clear. But it's the most common thing people would look for, so I think it should go first.

And now that I've written that, maybe it should just be in its own section after Form Logic

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,674 @@
...
+  Returns :arg:`then` if :arg:`expression` evaluates to ``True``. 
+  Otherwise, returns :arg:`else`.
+
+  
+.. function:: not(arg)
+
+  Returns the opposite of :func:`boolean(arg) <boolean>`.
+
+  
+.. function:: coalesce(arg, arg)
```
yanokwa: coalesce and boolean feel to my like utility functions. Or rather, they feel like functions that no one uses and they are front and center.

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,674 @@
...
+
+.. function:: number(string)
+
+  Converts a :arg:`string` of digits into a number value.
+
+.. function:: string(arg)
+
+   Converts :arg:`arg` to a string.
+
+.. _math-functions:
```
yanokwa: I was expecting to see count/max/min. Can we put a note here saying that there are some math operations that only work on nodesets?

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,674 @@
...
+  Returns an integer equal to the 1-indexed position of the current node
+  within the node defined by :arg:`xpath`.
+  
+  Most often this is used in the form :tc:`position(..)`
+  to identify the current iteration index
+  within a repeat group.  
+  
+  .. include:: incl/form-examples/parallel-repeat-groups.rst
+
+.. function:: uuid([length])
```
yanokwa: I'd put this above the checklist functions since it's not that used.

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,674 @@
...
+  nodeset
+  prefs
+  seleccionados
+  te
+  tres
+  π
+  
+  
+	
+******************************
```
yanokwa: I doubt this would be possible, but it'd be nice to see an alphabetical list of all the things in a table to make it glanceable.

Maybe a link to https://opendatakit.github.io/xforms-spec/#xpath-functions would be good enough?

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,674 @@
...
+  .. note::
+  
+    This function returns the :th:`name`, not the :th:`label`,
+    of the selected choice.
+    To get the label in the current language,
+    use :func:`jr:choice-name`.
+  
+  Returns an empty string if the index does not exist.
+  
+  .. image:: /img/form-functions/selected-at-0.*
```
yanokwa: Can all these be available as a detail? It makes it hard to skim.

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,674 @@
...
+  tres
+  π
+  
+  
+	
+******************************
+Form Functions and Operators
+******************************
+
+:ref:`expressions` in :ref:`calculations`, :ref:`constraints <constraints>`, and :ref:`relevants <relevants>`
```
yanokwa: Possible to make this lowercase in the display?

adammichaelwood:
Addressed all of the above (I think.)

I can pull operators into their own doc if you want. For now, I just moved them to the top of the page.

adammichaelwood:
(Oh -- except the details thing. I haven't done that. Will add if you want me to. I don't mind scrolling as much as you do, I think.)

adammichaelwood:
ok - i changed my mind.
I tried the details thing and it is totally better.

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,719 @@
...
+  
+	
+******************************
+Form Functions and Operators
+******************************
+
+:ref:`expressions` in :ref:`calculations <calculations>`, :ref:`constraints <constraints>`, and :ref:`relevants <relevants>`
+can contain operators and functions.
+
+.. seealso:: `XPath Functions <https://opendatakit.github.io/xforms-spec/#xpath-functions>`_
```
lognaturel: Maybe "XPath functions in the ODK XForms specification" to make it clear what that doc is?

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,719 @@
...
+Boolean operators
+------------------
+
+.. csv-table::
+  :header: , Explanation, Example
+
+  and, ``True`` if the expressions before and after are ``True``, ${age} > -1 and ${age} < 120
+  or, ``True`` if either of the expressions before or after are ``True``, ${age} < 19 or ${age} > 64
+  
+.. _comparison-operators:
```
lognaturel: I would tend to put comparison operators before boolean operators. Most people aren't familiar with the word "boolean". Perhaps there can be a short sentence when introducing comparison operators that says that the result is always `true()` or `false()` (boolean). Then the boolean operators introduction can say they join two boolean expressions (that evaluate to `true()` or `false()`).

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,719 @@
...
+  =, equal to, ${enrolled} = 'yes', Can compare numbers or strings.
+  !=, not equal to, ${enrolled} != 'yes', Can compare numbers or strings.
+  >, greater than, ${age} > 17, 
+  >=, greater than or equal to, ${age} >= 18,
+  <, less than, ${age} < 65, 
+  <=, less than or equal to, ${age} <= 64,
+
+.. _math-operators:
+  
+Math operators
```
lognaturel: I would put math operators first since they are familiar.

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,719 @@
...
+
+  
+.. _xform-operators:
+
+Operators
+==========
+
+.. _utility-operators:
+
+Utility operators
```
lognaturel: These are XPath references to the current and containing node. They can be used to build up relative paths (e.g. ../../food). It's ok to put them in this section but I think they should have a little more context. See https://opendatakit.github.io/xforms-spec/#xpath-paths. Perhaps this should go last since it's a little more esoteric.

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,719 @@
...
+  questions inside :ref:`repeat groups <repeat-functions>`.
+
+.. _select-functions:
+  
+Select questions
+~~~~~~~~~~~~~~~~~~~
+
+.. function:: selected(select_question, choice_name)
+
+  Returns ``True`` if :arg:`choice_name`
```
lognaturel: I think it's useful to describe this a little more generally and mention that it searches for the second parameter in the space-separated list specified by the first parameter. There are all kinds of creative ways to use `selected` and `selected-at` with that more general concept.

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,719 @@
...
+~~~~~~~~~~~~~~~~
+
+.. admonition:: Helpful terms
+
+  .. glossary::
+    :sorted:
+
+    nodeset
+
+      A collection of response values.
```
lognaturel: It may be worth mentioning that this is as in XPath -- a set of any XML nodes. You can build a nodeset a number of ways. For example, if you have nodes `/data/name` and `/data/age`, you can do `/data/name | /data/age` to build up a nodeset that contains both nodes. Functions like `randomize` return nodesets not necessarily of responses (generally of choices).

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,719 @@
...
+
+.. function:: boolean-from-string(string)
+
+  Returns ``True`` if :arg:`string` is "true" or "1".
+  Otherwise, ``False``.
+
+
+.. function:: number(string)
+
+  Converts a :arg:`string` of digits into a number value.
```
lognaturel: Might want to say the value is `NaN` if it can't be converted to a number. 1 if `true()`, 0 if `false()`. https://www.w3.org/TR/1999/REC-xpath-19991116/#function-number

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,719 @@
...
+Number handling
+~~~~~~~~~~~~~~~~~
+
+.. function:: round(number, places)
+
+  Rounds a decimal :arg:`number` to some number of decimal :arg:`places`.
+
+.. function:: int(arg) 	
+
+  Converts :arg:`arg` to an integer.
```
lognaturel: To be explicit, I might say "truncates the fractional part of a decimal value to make it an integer"

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,719 @@
...
+Date and time
+----------------
+
+.. function:: today()
+
+  Returns the current date without a time component.
+
+.. function:: now()
+
+  Returns the current datetime in the current time zone.
```
lognaturel: in iso 8601 format including the timezone - https://en.wikipedia.org/wiki/ISO_8601

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,719 @@
...
+.. function:: random()
+
+  Returns a random number between 0.0 (inclusive) and 1.0 (exclusive).
+
+  
+.. function:: randomize(nodeset[, seed]) 	
+
+  Returns a shuffled :arg:`nodeset`.
+  
+  A shuffle with a numeric :arg:`seed` is deterministic and reproducible.
```
lognaturel: May want to suggest saving a seed in a previous calculate (e.g. using `random` or `decimal-date-time(now())`) and using it so that the order can be recreated.

odk1-src/form-logic.rst
```diff
@@ -316,7 +308,7 @@ Often, comparison `operators`_ are used in relevance expressions. For example:
 :tc:`${has_children} = 'yes'`
   True if the answer to :tc:`has_children` was ``yes``.
   
-Relevance expressions can also use `XPath functions`_.
+Relevance expressions can also use :doc:`XPath functions <form-functions>`.
```
lognaturel: May be best to use just `functions` here. There's a mix of XPath and XForms and madeup functions, really.

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,719 @@
...
+  
+.. function:: checklist(min, max, response[, response[, response [, ... ]]])
```
yanokwa: I'd move checklist/weighted-checklist to the bottom, above true/false because they are so infrequently used.

odk1-src/form-functions.rst
```diff
@@ -0,0 +1,719 @@
...
+  prefs
+  seleccionados
+  te
+  tres
+  π
+  
+  
+	
+******************************
+Form Functions and Operators
```
yanokwa: Let's split into two documents.

lognaturel: I'm not so sure about this one! Operators can join the results of functions. I think of this document as being about the building blocks of expressions.

yanokwa: If y'all don't want to split, that's fine by me. But at the minimum maybe it should be called `Form Operators and Functions` to match the order?

adammichaelwood: I agree with this last point. I'll change the file name too.

odk1-src/form-functions.rst
adammichaelwood: Could one do:

``${question_name_1} | ${question_name_2}``

?

adammichaelwood: > Functions like randomize return nodesets not necessarily of responses (generally of choices).

odk1-src/form-functions.rst
adammichaelwood: > Perhaps this should go last since it's a little more esoteric

I don't know... in their simple "naive" sense, they get used all the time. You can't really deal with constrains without `.`, and `position(..)` is pretty common inside repeats.

I think we need a doc or section on XPath, separate from this. Once that is in place, I think linking from here to there with a note ("These are actually XPath references.... etc...") would be a good idea. For now, maybe a similar note that links to https://opendatakit.github.io/xforms-spec/#xpath-paths.

adammichaelwood: I'm adding a note, and --- after trying a few arrangements --- ended up putting them last as suggested.

lognaturel:
Looks good to me!

PR #663, opened 2018-04-19
removes section on referencing previous iterations

### LINKED ISSUE #661: Referencing answers from previous iteration example has several problems
https://docs.opendatakit.org/form-repeats/#referencing-answers-from-the-previous-iteration

The calculation is in the wrong column, the parens for the calculated expression aren't right and defaults can't be dynamic!

@adammichaelwood @yanokwa are y'all making sure these examples run on actual devices when you do review? Perhaps we should make that part of either the PR template or a review checklist so we can be sure examples are authoritative.

adammichaelwood:
closes #661 

removes the section about referencing responses from previous repeat iterations

If we think of a good use case for why someone would want to reference a previous iteration,
(other than populating a default, which is what most people want to do, but can't)
we would want to add this section back.

Shame on me. Shame on my family. Shame on my cow.

### DIFF
--- a/odk1-src/form-repeats.rst
+++ b/odk1-src/form-repeats.rst
@@ -127,31 +127,6 @@ use :func:`indexed-repeat` and :func:`position`.
   Most often this is used in the form ``position(..)``
   to identify the current iteration index.
 
-.. _referencing-previous-iteration:
-  
-Referencing answers from the previous iteration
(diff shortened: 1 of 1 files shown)


lognaturel:
Hey, no big deal, we caught it quickly! 😊🐮

PR #666, opened 2018-04-23
Improve ignore

yanokwa:
* I use pyenv, so wanted to add that to the ignore
* Added comments so we know what is being ignored
* Added minor tweak to emacs ignore. Stolen from https://github.com/github/gitignore/blob/master/Global/Emacs.gitignore#L3
* Controversial? Removed style-checks.py and extra.py. If they are important, they maybe should be in the repo? If they are not important, they should be elsewhere?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,7 @@ build/
 tmp*
 
 # emacs backups
-\#*.*\#
+\#*\#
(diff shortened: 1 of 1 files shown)


yanokwa:
This should be ready to go now.

PR #672, opened 2018-04-28
form localization

adammichaelwood:
closes #611 

#### What is included in this PR?

New doc on form localization.

PR #688, opened 2018-05-01
Clarify potential confusion about where localization happens

yanokwa:
Addresses #672

Minor language clean up and clarify potential confusion about where localization happens

I spent way too much time coming up with the right language and placement. It's not perfect, but I think it's good enough.


adammichaelwood:
> I spent way too much time coming up with the right language and placement.

I know that feel.

PR #691, opened 2018-05-02
add tip on default workaround

### LINKED ISSUE #664: Add note about defaults being static values only, possibly with a workaround
Defaults are set once at form beginning so they can't be dynamic. The barrier to making them dynamic is that often the values that would want to be used as defaults are changing and it's not clear when they should update and when the user should have control.

https://forum.opendatakit.org/t/is-it-possible-to-set-a-default-value-for-a-decimal-question-using-the-calculate-column/12141/3 shows one approach that sets the dynamic default whenever the target value is blank. Once it's set, it can be modified by the user. With this approach, the target value cannot ever be blank (it's either the default or a value set by the user).

adammichaelwood:
closes #664

 - added a line about static values only in defaults
 - added a tip section with a workaround and simple example

I'm not sure if it is needed, but --- @lognaturel's original example was much more complicated. I simplified to just calling a simple variable in `once()`. This works (as far as I can tell). I don't know if a more complicated example is definitely needed. It might be helpful... but I found the more complicated example very confusing.

### DIFF
--- a/odk1-src/form-logic.rst
+++ b/odk1-src/form-logic.rst
@@ -211,6 +211,9 @@ Setting default responses
 To provide a default response to a question,
 put the response value in the :th:`default` column.
 
+Default values must be static values,
+not expressions or variables.
+
 .. rubric:: XLSForm
 
 .. csv-table:: survey
@@ -224,6 +227,22 @@ put the response value in the :th:`default` column.
   contacts, phone_call, Phone call
   contacts, text_message, Text message
   contacts, email, Email
+  
+.. tip:: 
+
+  You may want to use a previously entered value as a default,
+  but the :th:`default` column does not accept dynamic values.
+  
+  To workaround this, use the :th:`calculation` column instead,
+  and wrap your default value expression in a :func:`once` function.
+  
+  .. rubric:: XLSForm
+  
+  .. csv-table:: survey
+    :header: type, name, label, calculation
+    
+    text, default_value, Enter any text.,
+    text, calculated_default, Leave the value or change it., once(${default_value})
   
 .. _constraints:
 


odk1-src/form-logic.rst
```diff
@@ -225,6 +228,22 @@ put the response value in the :th:`default` column.
...
+  To workaround this, use the :th:`calculation` column instead,
+  and wrap your default value expression in a :func:`once` function.
+  
+  .. rubric:: XLSForm
+  
+  .. csv-table:: survey
+    :header: type, name, label, calculation
+    
+    text, default_value, Enter any text.,
+    text, calculated_default, Leave the value or change it., once(${default_value})
```
lognaturel: This works if you want the default value to be exactly the FIRST value that ${default_value} is ever set to.

I would highlight two limitations:
- If the value of ${default_value} changes *even before the user gets to the ${calculated_default} question*, the default will match the original value of ${default_value}, not the updated one.
- If ${default_value} has a value, ${calculated_default} MUST also have a value. If the user sets it to blank, it will go back to matching ${default_value} (which may or may not have changed, as described in the first point).

The wording needs some work but hopefully the concepts make sense.

yanokwa: Perhaps we can have an example?

age
gender
malaria_age
favorite_color

If I set age to 8, and go to malaria_age, it will be set to 8.

If I set age to 8, go to gender, go back to age and set it to 12, and go to malaria_age, it will be 8. This is because the expression in once() was evaluated when I first set age.

If I set age to 8, go to gender, go back to age and set it to 12, and go to malaria_age and set it blank, and go to favorite_color, malaria_age will be set to 12. This is because the expression in once() was re-evaluated when I set malaria_age to blank and the value of age was pulled at that time.

adammichaelwood:
See if that helps.

Issue #699, opened 2018-05-03
How much HTML to document and where to put it.


MartijnR:
Maybe this is no longer true after the latest markdown updates (ported from Enketo) in Collect which covered various header issues:
> If a Markdown header is used, the label or hint can only be one line of text. Line breaks in the XLSForm cell will break the header styling.

Also, just FWIW, it would make me very happy if the rogue HTML support was not documented (except for `<span>`)~~. This was an unhappy event of the dark past and it was more or less agreed not to document this if I remember correctly~~, so we don't have to explain to users that one type of formatting doesn't work in e.g. Enketo.

adammichaelwood:
From #700:

>I don't have a strong opinion on the development issue. (Though --- if I had one --- my preference would be to support HTML as much as possible/reasonable.)
>
>I do have a strong opinion about documentation: We should document what is there. If what is there changes, we should change the documentation.
>
>Perhaps mention of (limited) HTML support would be better in another place, instead of in styling. I'm planning to do a small "anatomy of a form question" section, which will include things like "don't put spaces in name". Perhaps that is the right place to mention --- without examples that encourage use --- that Collect supports this small list of HTML tags. (Along with the notice that this is not supported in Enketo.)

yanokwa:
My rough proposal here is that we say...
1. We support and recommend Markdown as the way to style.
1. Add a section under Markdown called "Inline HTML"  (because Markdown does support inline HTML). That section will warn that use of HTML outside of <span> is not ODK standards compliant and it's not recommended. We also note that depending your version of Android some tags might work (and link to latest Android source), so stay away.

MartijnR:
Sounds good to me and will be helpful to users. Maybe good to separate `<span>` plus likely `<sub>` and `<sup>` from the other tags, but that may also makes things too confusing again.

P.S when looking at the ODK Collect code, I think you only support `color` and `font-family` rules as acceptable content of the `style` attribute in a `<span>` (which would be great to add to documentation too - Enketo hasn't implemented such a style filter (unfortunately) so I just won't document other style rules).

PR #700, opened 2018-05-03
Remove html

yanokwa:
Addresses #699

Remove the HTML docs
Reword the section title on custom styling

* Remove the H*s in all-headers-label section. Maybe just drop the screenshot and show an example of an H6 and let the user fill in the blanks.
* Remove the H1 in Color and font styling can be combined
* Adjust style-example.xlsx to remove any HTML. Even the <br>, but I think newlines there should work. I hope.

@adammichaelwood There will be a tweet about styling going out on Monday, so ideally we can merge this as soon as possible and do the rest in a follow-on PR?

### DIFF
--- a/odk1-src/form-styling.rst
+++ b/odk1-src/form-styling.rst
@@ -165,13 +165,13 @@ you'll need to escape that too.
   
   note, escape_md, \# This headline is normal sized, \*Asterisks\* and \_underscores\_ and one slash: \\
   
-.. _style-attribute:
+.. _custom-styling:
   
-Styling with the style attribute
+Custom styling
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 To add custom styling to hint, label, and choice labels,
-use `the style attribute`_.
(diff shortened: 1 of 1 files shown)


yanokwa:
Fair point, @adammichaelwood. @MartijnR can we strike a balance here? Could we add a warning that says that because ODK Collect renders in a WebView, by default it will render this list of tags, but that behavior is not standards-complaint?

MartijnR:
Thanks for this. I agree that the `<span>` support is an opening for further HTML tags support. I'm fine with agreeing on other tags when necessary, i.e. if there is no pure markdown syntax available for them. `<sup>` and `<sup>` are okay with me if that's what we agree on in the forum discussion. (I'd just not process them if they have attributes, I think.)

With `<br>` in conjunction with literal line breaks, I'm not sure how to deal properly with a combination of both (as is common in HTML where the literal line break would be ignored). Currently, two linebreaks create a paragraph.

P.S. Note that the weird thing we do is that we actually support both `&lt;span&gt;` and `<span>` (Pyxform encodes HTML entities).

PR #701, opened 2018-05-05
Update pdf header, footer and content listing

### LINKED ISSUE #557: Include note in the PDF about the content being out of date
Currently, the PDF (#537) that is being generated doesn't let users know that it might be out of date. We discussed this in the original issue at https://github.com/opendatakit/docs/issues/271#issuecomment-334850319.

I'd propose a few things...
1. Remove the `Release 1.0` text from the PDF. It makes it seem like things are versioned in an orderly fashion.
1. To let people know that the PDF isn't up-to-date, we could put something in the footer that says 
   > ODK software is always being updated, so this PDF is likely out of date. Get the latest guide at https://docs.opendatakit.org.

ankita240796:
closes #557 

- Update pdf header and footer.
- Add a note in pdf footer about latest documentation.
- Update commands in ``config.yml`` so that content list is generated in pdf.

Here are the screenshots for the same:
image
image
image

(It feels great to be back :tada: )


odk1-src/conf.py
```diff
@@ -66,7 +66,7 @@
 # The short X.Y version.
 version = '1.0'
 # The full version, including alpha/beta/rc tags.
-release = '1.0'
+#release = '1.0'
```
yanokwa: I get a little nervous when something is commented out! Why was this removed?

odk1-src/conf.py
```diff
@@ -160,6 +160,18 @@
     # Additional stuff for the LaTeX preamble.
     'preamble': '''
     \usepackage{fontspec}
+    \makeatletter
+      \\fancypagestyle{normal}{
+        \\fancyhf{}
+        \\fancyfoot[LE,RO]{{\py@HeaderFamily\\thepage}}
+        \\fancyhead[LO]{{\py@HeaderFamily\\nouppercase{\\rightmark}}}
+        \\fancyhead[RE]{{\py@HeaderFamily\\nouppercase{\\leftmark}}}
+        \\fancyfoot[CE,CO]{{\py@HeaderFamily{Get the latest guide at https://docs.opendatakit.org}}}
```
yanokwa: It would be nice the text explained why and included a clickable link (if possible).

> We update our documentation daily, so this file is likely out of date! Get the most current version at [https://docs.opendatakit.org](https://docs.opendatakit.org).

Also, if possible, let's make the text not-bold and a little smaller since it's on every page.

adammichaelwood: I agree with this.

adammichaelwood:
> It's so nice to have you back, @ankita240796. We missed you!

+1

(I also agree with @yanokwa's other comments.)

----

Should we add this to `odk2-src/conf.py` as well?

odk1-src/conf.py
ankita240796: Sphinx has defined its own release command for latex for customization. I have renewed the command to take no arguments and display no release information. So, I commented this out since it was providing a release number argument for the release command.

odk1-src/conf.py
ankita240796: Sure, I will fix this!

ankita240796:
Hi @adammichaelwood @yanokwa, I missed ODK too!

I have updated the PR - the footer is displayed on chapter start pages as well it is explained with a link.

image

I would suggest we should have the changes for odk2 too, shall I make the changes?

yanokwa:
Better to have this warning than not. I'll merge this and follow on with my proposed changes.

ankita240796:
Sorry for the delay on this. I will submit a follow on PR for the same, Thanks!

PR #712, opened 2018-05-19
adds inline html

### LINKED ISSUE #699: How much HTML to document and where to put it.

adammichaelwood:
closes #699 

 - mention of inline html, with lots of caveats
 - link to list of supported tags
 - mention that span/style only supports color and font-family
 - remove the html xlsform sample for header size comparison
   (left the image, because I think it is useful)

### DIFF
--- a/odk1-src/form-styling.rst
+++ b/odk1-src/form-styling.rst
@@ -58,14 +58,6 @@ Labels and hints can be styled with one of six header levels.
   A comparison of headline sizes. This exact effect :ref:`cannot be produced using Markdown <one-headline-only>`.
   
   
-.. rubric:: XLSForm
-
-.. csv-table:: survey
-  :header: type, name, label, hint
-  
-  note, h1_label, # Largest Headline - H1, This note has a label with a Markdown-style header.
-  note, all_headers, <h1>Largest Headline - H1</h1><h2>H2 Headline</h2><h3>H3 Headline</h3><h4>H4 Headline</h4><h5>H5 Headline</h5><h6>H6 Headline</h6>, Headers in the label.
-  
 .. warning::
   :name: one-headline-only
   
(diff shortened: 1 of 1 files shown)


odk1-src/form-styling.rst
```diff
@@ -164,15 +156,36 @@ you'll need to escape that too.
...
+
+Many Markdown implementations support inline HTML,
+but Collect only supports a small subset of HTML elements.
+Support of HTML is further limited because:
+
+- Your exact Android device, operating system version,
+  and other device-related factors
+  will affect what HTML can be rendered, and how it is rendered.
+- HTML is not supported
+  by other form rendering tools in the XForm ecosystem.
```
yanokwa: XForms not XForm

odk1-src/form-styling.rst
```diff
@@ -164,15 +156,36 @@ you'll need to escape that too.
...
+  For example, HTML elements that work in Collect may not work in Enketo.
+
+For these reasons, we do not recommend using HTML in forms (except the ``<span>`` element :ref:`noted below <custom-styling>`).
+
+.. seealso:: `The list of HTML tags currently supported in Collect <https://www.grokkingandroid.com/android-quick-tip-formatting-text-with-html-fromhtml>`_.
+
   
 .. _custom-styling:
```
yanokwa: I think that this should go under Markdown in the ToC, not between Markdown and Emoji as it is currently. @lognaturel thinks it's fine. What do you think, @adammichaelwood?

odk1-src/form-styling.rst
adammichaelwood: right

odk1-src/form-styling.rst
adammichaelwood: I wrote the text thinking it was under Markdown, 
I just did the wrong headline underline.

adammichaelwood: I don't think of the `<span>` construct as being a part of Markdown support,
but I don't have a strong opinion on the matter.

adammichaelwood: It does occur to me that this would be more clear if it was "Custom font styling"

yanokwa: Maybe then the text at the very top of the document can read, "Labels, hints, and choices in an XLSForm can all be styled using Markdown, fonts and colors, and emoji." Then we change the title to "Fonts and colors"?

yanokwa: That is, form styling is already custom styling, so it's a bit redundant to call it that.

adammichaelwood: I like this suggestion.

PR #717, opened 2018-05-28
adds file widget

### LINKED ISSUE #667: Widget that allows attaching files of any type
On the Collect's side, we are working on `ArbitratyFileWidget` https://github.com/opendatakit/collect/issues/2135 
It's a widget that allows selecting any type of file.

Once it's merged it should be added to the doc also.

adammichaelwood:
closes #667 

A new widget section for the file widget.

[This conversation about Pyxform support](https://github.com/XLSForm/pyxform/issues/194) and the [PR that satisfies it](https://github.com/XLSForm/pyxform/pull/198) suggests that a form designer could specify allowed file types, but I can't find an example of how it would be done. The PR includes a sample XLSX file, but it doesn't specify any file types.

#### Note --- this is a new feature

This PR should not be pushed until the associated Collect and Pyxform features are released.

### DIFF
--- a/odk1-src/form-question-types.rst
+++ b/odk1-src/form-question-types.rst
@@ -1521,6 +1521,10 @@ Records video, using the device camera.
 File upload widget
 --------------------
 
+.. versionadded:: 1.15
+
+  `ODK Collect v1.15.0 <https://github.com/opendatakit/collect/releases/tag/v1.15.0>`_
+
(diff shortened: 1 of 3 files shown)


lognaturel:
> a form designer could specify allowed file types

Collect currently ignores this so let's add it once there's client support.

PR #725, opened 2018-05-29
adds note about geopoint dialog

### LINKED ISSUE #704: Geopoint dialog now shows number of satellites and time elapsed
https://github.com/opendatakit/collect/pull/2146

Updates the dialog to add
```
Satellites available: n/24
Time elapsed: mm:ss
```

24 is the total number of satellites used by the Global Positioning System. The number before it represents how many satellites are used in the fix. At least 4 are required to get a location. If that number is small, an enumerator should make sure there is nothing obstructing the device's view of the sky.

Time elapsed is the amount of time since the dialog was opened. It can be used to help troubleshoot issues. For example, the enumerator may tell their supervisor that after 5 minutes of waiting, the accuracy was 15m. This is useful information to share on the forum, for example.

adammichaelwood:
closes #704 

Added a `tip` about the satellite count and elapsed time.

### DIFF
--- a/odk1-src/form-question-types.rst
+++ b/odk1-src/form-question-types.rst
@@ -1088,7 +1088,9 @@ see :ref:`placement-map-widget`
   - Number of satellites found
   - Elapsed time
   
-  If the device can not find at least four (4) satellites,
+  To get an accurate fix on your location, 
+  the device must find at least four (4) satellites.
+  If the device can not find at least four satellites,
   make sure there is nothing obstructing the device's view of the sky.
   
   If you have problems with the geopoint widget, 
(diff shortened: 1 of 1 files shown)


odk1-src/form-question-types.rst
```diff
@@ -1078,6 +1078,25 @@ Captures the current location from the device.
...
+
+  .. versionadded:: 1.15
+
+  While determining the current coordinates,
+  you will see a dialog box displaying:
+  
+  - Number of satellites found
+  - Elapsed time
+  
+  If the device can not find at least four (4) satellites,
```
yanokwa: Why 4 (vs 3 or 6)?

Issue #456, opened 2018-01-30
Edit ODK Build page


adammichaelwood:
Part of #402 

 - Follow OL/screenshot/details style for procedures

PR #727, opened 2018-05-29
Standardize on tool URLs

yanokwa:
Addresses #456 

* Change URL to match other tools. I noticed this in the launch of the website.
* Remove tools that are no longer relevant

I'm not 100% sure the redirects that I've put in work, so I'd appreciate a closer look there.


odk1-config/s3_website.yml
```diff
@@ -38,4 +38,5 @@ redirects:
   briefcase-forms/index.html: briefcase-using
   briefcase-vs-aggregate/index.html: briefcase-and-aggregate
   form-widgets/index.html: form-question-types
-  form-interaction/index.html: form-logic
\ No newline at end of file
+  form-interaction/index.html: form-logic
+  odk-build/index.html: /build-intro
```
adammichaelwood: I can't test this until it goes live, but 
(based on the existing precedent of the preceding lines)
I think this should be:

yanokwa:
@adammichaelwood Are you sure there are no slashes? The other links above that section do...

adammichaelwood:
Are we looking at the same thing?
The line directly above it does not have a slash on the right side of the colon.

yanokwa:
See https://github.com/yanokwa/odk-docs/blob/HEAD/odk1-config/s3_website.yml#L32

adammichaelwood:
Well, at least neither of us is crazy. Seems to not matter --- they both work. I'm merging. (Other than cleanliness, is there a reason to prefer one or the other?)

PR #734, opened 2018-06-07
Adds section about empty number values

### LINKED ISSUE #731: Blank values are nil, not 0 and can't be used in math expressions
https://forum.opendatakit.org/t/calculation-questions/13294

Would be good to add a tip, perhaps at https://docs.opendatakit.org/form-operators-functions/#math-operators. I'm not sure where a user is most likely to find it.

adammichaelwood:
closes #731 

 - add a "gotcha" subsection to Form Logic building blocks
 - demote "when expressions are evaluated" to subsection of gotcha
 - add section on nil values to form logic gotchas
 - fix some incorrect section header underlining in that file
 - add NaN to spelling list

There's no obviously good place to put this info. @lognaturel suggested a warning in the math operators section, but it applies to math functions and potentially comparisons as well. 

What I came up with makes sense to me, but if anyone disagrees I am open to other ideas.

### DIFF
--- a/odk1-src/form-logic.rst
+++ b/odk1-src/form-logic.rst
@@ -188,19 +188,23 @@ That is, they have no value.
 When a :ref:`variable <variables>` referencing an empty value is used 
 in a math :ref:`operator <math-operators>` 
 or :ref:`function <math-functions>`,
-it is treated as Not a Number (NaN).
+it is treated as Not a Number (``NaN``).
 The empty value **will not** be converted to zero.
-The result of a calculation including NaN 
-will also be NaN, 
+The result of a calculation including ``NaN`` 
+will also be ``NaN``, 
 which may not be the behavior you want or expect.
 
 To convert empty values to zero,
-use the :func:`coalesce` function.
+use either the :func:`coalesce` function
+or the :func:`if` function.
 
 .. code-block:: none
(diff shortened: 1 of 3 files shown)


lognaturel:
Nice. That place looks good to me but to increase the changes a user would see it, how about adding a warning in each of the relevant sections that links to this text? I think it would also be helpful to include the `if(${value} = '', 0, ${value})` structure because it may be easier for users to remember. I never recall `coalesce` and spelling it is super awkward.

adammichaelwood:
> spelling it is super awkward.
this is vrey true

tiritea:
oh man! I **just** discovered `coalesce()`, and now I gotta go back to using clunky ol' if/then/else's ... sigh. :-(

But yes, `if(...)` is a bit more obvious. But maybe we can add "...or use coalesce(${value},0) to achieve same result" ?

lognaturel:
@tiritea Not sure I understand your comment! Did you look at the diff? The page is also live at https://docs.opendatakit.org/form-logic/#empty-values-in-math

lognaturel:
Sounds good! I encourage you to link to that section, too. And hopefully we'll see less questions about it though it's certainly an odd concept for non-developers.

PR #755, opened 2018-06-29
link update

danbjoseph:
previous link goes to a partially broken page with no apparent way to download

not sure if link should be updated to http://opendatakit.org/software/odk2/ (currently in this PR) or straight to https://github.com/opendatakit/services/releases (which you end up clicking through to from the former link) or something else?


danbjoseph:
same link appears on 
- `/odk2-src/survey-install.rst`
- `/odk2-src/tables-install.rst`

yanokwa:
Thanks for sending this in @danbjoseph. I've merged and fixed the other broken links in #756!

PR #766, opened 2018-07-08
Compress images

### LINKED ISSUE #765: Compress surprisingly large images
There are some PNGs that are pretty big. You can find them with this command: `find . -type f -name "*.png" -exec du -h {} + | sort -r -h`. These files don't see to respond to lossless compression, so we shoud use lossy but with reasonable defaults (maybe 90%) to bring the size down.

yanokwa:
Closes #765

Most of the images that are surprisingly large are in odk2-src. I'm not sure why these images are so big and don't compress initially. 

These files did not compress at all losslessly when I started this PR, so I ran all these through pngquant with `--strip --speed 1 --quality 90 --force --ext .png` and that brought the file size down. Rather than take the biggest ones arbitrarily, I compressed groups of images (e.g., all the cold-chain-tour pngs) so they look reasonable next to each other.

This change saves about 30 MB. The files are not as small as I'd like, but every little bit counts :)


lognaturel:
Darn, should have merged this before removing the SHA images. @yanokwa can you please fix conflicts?

PR #782, opened 2018-07-23
Add basic docs for SMS submission

lognaturel:
partially addresses #781

#### What is included in this PR?
First pass at SMS docs.

#### What is left to be done in the addressed issue?
Some of the components described in #781 will need to be expanded on. All screenshots of Server settings will need to be updated.

#### What problems did you encounter?
It's hard to know where this should be linked from and how it should be split up. Perhaps the form design part should go in form design, for example. 

CC @jd-alexander

PR #783, opened 2018-07-23
Modify use of short_tag to compact_tag per current pyxforms

milesstewart:
addresses #782

Modification to documentation to use ``compact_tag`` instead of ``short_tag`` for ODK Collect to properly send data fields over SMS.


yanokwa:
Good catch, @milesstewart! @lognaturel claims she made that mistake to make sure that people were paying attention 😂

lognaturel:
🙏

PR #785, opened 2018-07-31
Added redirect mechanism for anchor links

### LINKED ISSUE #643: JS redirect solution for anchor links
When a section title is changed it breaks anchor links.
(Note, this has to do with the actual title of the section, not the in=sphinx section reference name.)

We can't fix this server side (anchor links are handled on the client).
It would be good to have some kind of anchor link redirect feature,
with a simple mapping file that content authors can update if they re-title a section.

ankita240796:
closes #643 

This PR adds redirect mechanism for anchor links.


ankita240796:
Hi @adammichaelwood can you please review this. Also can you tell me from where can I prepare a list of old and new anchor links to add them in the json file. Thanks!

ankita240796:
Hi @yanokwa @ln, can you help me here, Thanks!

lognaturel:
Thanks for doing this @ankita240796 and I apologize for the delay in getting a response!

This looks good to me. I have a slight concern about performance because this means the full DOM has to be parsed on load but I think clients are fast enough these days that it's probably not an issue. I also don't really see any alternative.

I verified this by adding `"hidden-questions": "text-widgets"` and then loading `http://localhost:8000/form-question-types/#hidden-questions`. It works very well on first load. And then it seems like if I paste in the same URL to the address bar, it doesn't reload the document so the redirect doesn't happen. But weirdly, if I do that again, it does. In other words, I get the expected behavior every other time that I paste the URL in the address bar. I don't think this is a huge deal since it's not a likely scenario but thought I would call it out.

This only works when anchors are unique across the site and across time. That is, I can't have #foo on one page that redirects to #bar and then reuse #foo on another page. Again, not sure it's a deal breaker but I wanted to point it out.

What do others think of these limitations?

In terms of anchors that need redirecting, I'm afraid I didn't keep track of any so that's more a question for @adammichaelwood or @yanokwa.

ankita240796:
Hi @lognaturel, Thanks for the review.

I will try to figure out a solution for the first problem you mentioned.

Regarding the second issue we can fix it by having a page wise redirect mapping in the json file, I will update the PR if you think this is the right fix.

Thanks!

lognaturel:
👍 I think it'd be helpful to hear from others whether the limitations are important to address. @yanokwa @adammichaelwood perhaps have some thoughts on this?

ankita240796:
Thanks for the review @yanokwa! I will make the changes suggested by you!

ankita240796:
Hi @yanokwa, I have made the changes as suggested by you.
Please take a look, Thanks!

yanokwa:
@ankita240796 I have the following in hash-redirects.json
```json
{
	"form-question-types": {
		"old": "geopoint-widget"
	}
}
```

When I run the docker build (`./run-task.sh odk1-autobuild`) and go to http://0.0.0.0:8080/form-question-types/#old, it doesn't redirect. Are you able to reproduce this behavior?

ankita240796:
Hi @yanokwa, sorry for the delay in the reply. I was out due to a college trip. 

I was not able to reproduce the behaviour as mentioned by you. Redirect occurs as expected when I run the docker build and go to http://0.0.0.0:8080/form-question-types/#old.

Here is gif for the same:

video

ankita240796:
Hi @yanokwa, do you still get the issue for redirects as mentioned in your previous comment?

ankita240796:
Hey @yanokwa, is this still required (I can fix the merge conflicts if that's the case) or can I close the PR?

yanokwa:
Thanks so much for following up, @ankita240796! 

I think after 3 years of inaction, we should probably close it.

PR #784, opened 2018-07-29
Added info to build pdf locally

ankita240796:
closes #556 

#### What is included in this PR?
Adds documentation on building PDF locally.

PR #789, opened 2018-08-03
Move PDF to after building docs, fix case sensitivity, use Docker ima…

yanokwa:
Addresses #784 

Couple of changes to tighten the prose...
* I moved the PDF to after building the docs. Feels like a better place.
* I made sure we are using the correct case for all the things (e.g., MacTex instead of mactex).
* I used the Docker image we use in CircleCI so we know it works.

https://github.com/opendatakit/docs/pull/788 has to be merged before the `make odk1-pdf` works.

### DIFF
--- a/shared-src/docs-tech-guide.rst
+++ b/shared-src/docs-tech-guide.rst
@@ -819,49 +819,6 @@ to get any help about style guide testing:
 
   $ python style-test.py -h
 
-.. _docs-pdf:
-
-Building PDF for docs
-~~~~~~~~~~~~~~~~~~~~~~
-
-The docs can be downloaded as PDF for the ease of users.
-
-To build the pdf locally, follow these steps:
-
-.. note::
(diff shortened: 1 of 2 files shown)


shared-src/docs-tech-guide.rst
```diff
@@ -937,6 +894,45 @@ It's a good idea to delete the ``build`` directory before each rebuild.
...
+  - For Windows, use `MiKTeX <https://miktex.org/>`_.
+  - For Ubuntu (or Debian), run
+
+    .. code-block:: console
+
+      $ sudo apt-get install texlive-xetex
+
+  .. note::
+
+    Instead of installing XeLatex directly on your system, you can also this `Docker image <https://github.com/schickling/dockerfiles/tree/master/latex>`_.
```
ankita240796: also **use** this

PR #790, opened 2018-08-03
PDF make fix

### LINKED ISSUE #787: Add make commands for pdf build
Make command should be added to build pdf for odk1 and odk2 docs.

yanokwa:
Closes #787 
Addresses #788

* Ensure sure the Docker image has make
* Persist the Makefile

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,9 +35,8 @@ jobs:
           name: Check spelling of ODK1 Documentation
           command: make odk1-spell-check
       - run:
-          name: Build PDF of ODK1 Documentation
-          command: |
-            make odk1-latex
+          name: Build LaTex of ODK1 Documentation
+          command: make odk1-latex
       - run:
           name: Compress images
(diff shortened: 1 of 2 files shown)


yanokwa:
@ankita240796 I think this should fix the issues you are seeing in your PR!

yanokwa:
I should have explained this better, @ankita240796! 

The `make odk1-pdf` command in your commit was being run on the `schickling/latex` Docker container. That command was failing because the latex container is very minimal and it doesn't have make installed. 

My goal behind suggesting we add the make command was that it'd simplify the docs you wrote and building locally, and it did. But running make in the latex container would have required we persist the src folders to the latex container, and also have sphinx installed on that container. It would have made for a very complicated build process, so I ended up going back to the old way of building the PDF in CI, but the new way for the Makefile that people use locally. 

Hope that is a bit clearer, if not, ask another question :)

ankita240796:
Thanks @yanokwa, made things quite clear for me :smile:

PR #794, opened 2018-08-14
drop google accounts from docs

wbrunette:
Support for google accounts was removed from ODK Services, therefore, the documentation should only talk about Aggregate usernames.


yanokwa:
Relatedly, we'll be removing this in future versions of Aggregate. https://github.com/opendatakit/aggregate/issues/286

PR #808, opened 2018-08-20
Typo and grammar fixes

lesley-morgan:
addresses #96 

Typos and grammar.

### DIFF
--- a/shared-src/docs-tech-guide.rst
+++ b/shared-src/docs-tech-guide.rst
@@ -11,7 +11,7 @@ when contributing to ODK Documentation.
 .. note::
 
   Developer and authoring tools 
-  have lots of options and alternatives. 
+  have a lot of options and alternatives. 
   Local tools and workflows presented in this guide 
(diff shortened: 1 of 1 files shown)


shared-src/docs-tech-guide.rst
```diff
@@ -463,11 +463,11 @@ Clone to local
 
 From your own fork of the repo on GitHub, 
 select the :guilabel:`Clone or download` button. 
-Copy the URI from the text box that opens up. 
+Copy the URL from the text box that opens.
```
adammichaelwood: URI is correct.

yanokwa: For those following along, if the web address has a protocol in front of it, it's a URL. If it doesn't, it's a URI. In the case of this dialog, if you use clone with HTTPS, then it's a URL because the string GitHub shows starts with https://. If you clone with SSH, then it's a URI because there is no ssh://.

So @adammichaelwood is correct, but unecessarily pendatic.

PR #809, opened 2018-08-20
Update links that are no longer relevant

yanokwa:
* Release notes are now split between old versions and new versions
* These are user docs, so reference the most likely use of an emulator
* Our source is on GitHub. We can say that more directly
* XLSForm link is local. Also, Gumroad isn't a thing we use anymore
* The developer wiki is out of date and not that useful

### DIFF
--- a/shared-src/docs-style-guide.rst
+++ b/shared-src/docs-style-guide.rst
@@ -1036,7 +1036,7 @@ XLSForm
 
 - The `XLSForm format for describing form in an Excel spreadsheet <http://xlsform.org/>`_
 - A spreadsheet file that describes a form using the format.
-- An `online tool <https://docs.opendatakit.org/xlsform/>`_ and an `offline tool <https://gumroad.com/l/xlsform-offline>`_ for converting :file:`*.xls(x)` files to XForm documents. 
+- A :doc:`tool <xlsform>` for converting :file:`*.xls(x)` files to XForm documents.
 
 When writing about any of these things, make sure you are clear --- in your mind as well as in your writing --- which one you are talking about.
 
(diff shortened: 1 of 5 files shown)


shared-src/docs-style-guide.rst
```diff
@@ -1036,7 +1036,7 @@ XLSForm
 
 - The `XLSForm format for describing form in an Excel spreadsheet <http://xlsform.org/>`_
 - A spreadsheet file that describes a form using the format.
-- An `online tool <https://docs.opendatakit.org/xlsform/>`_ and an `offline tool <https://gumroad.com/l/xlsform-offline>`_ for converting :file:`*.xls(x)` files to XForm documents. 
+- A `tool </xlsform/>`_ for converting :file:`*.xls(x)` files to XForm documents.
```
lognaturel: Should use the doc directive here.

adammichaelwood: i *think* cross reference links work for shared src stuff... (If i remember correctly, it ends up getting resolved via intersphinx)

lognaturel: Hmm, @yanokwa, there was an error when you tried it, right? Do you still know what the error was?

yanokwa: Finally figured it out. For future reference...

failed:
```
:doc:`tool </xlsform>`
```

worked: 
```
:doc:`tool <xlsform>`
```

lognaturel: 👍 Thanks @adammichaelwood for letting us know it was possible! Also, it looks much better without the `/` -- I didn't even recognize it as a link at first because of that.

Issue #810, opened 2018-08-20
Improve footer


yanokwa:
I think the footer could be simplified. To me, it should say there things.
1. We have a PDF.
2. If you have a problem, file an issue. 
   * I think a call to contribute is a bit too much.
3. Join the forum for x, y, z reasons.

PR #811, opened 2018-08-20
Remove FAQ because it is no longer relevant

### LINKED ISSUE #462: Edit/re-org/re-think FAQ
The FAQ page was brought over wholesale from the old site and is a bit all over the place in what it covers and doesn't.

yanokwa:
Closes #462 

* Remove FAQ page because the questions have either been answered in the obvious places  in the doc (e.g., how to use ODK, do we have an API), are no longer relevant (e.g., other data collection systems, why delayed list messages), or aren't actually that frequent (how to cite ODK).

I think we should improve the footer so there's a bit more guidance for things not convered in the docs. I've filed it here: https://github.com/opendatakit/docs/issues/810


lognaturel:
I think this is a good idea. I've filed https://github.com/opendatakit/website/issues/190 and https://github.com/opendatakit/website/issues/191 to keep track of some information I think would be worth keeping on the website.

Going to give this a couple days to see if anyone else has comments.

PR #814, opened 2018-08-20
Randomize choices

### LINKED ISSUE #754: Improve randomize docs
The docs on randomize() aren't correct. We should start from https://github.com/XLSForm/pyxform/issues/183 to improve them.

lognaturel:
Closes #754 

The first commit adds a section describing the XLSForm syntax for randomizing the order of choices. The second commit removes an example usage of the `randomize` function that doesn't work and clarifies that `calculate`s can't reference nodesets. 

I believe randomizing the order of choices is currently the only practical use for `randomize` and it gets hidden away by the XLSForm syntax but I think it's ok to leave the function in for completeness.


lognaturel:
Thanks so much, @adammichaelwood! 🎉

PR #820, opened 2018-08-21
fix function path references

### LINKED ISSUE #738: fix function refs
Several pages talk about functions,
but were written before the function reference,
so the functions didn't get linked.
They should be linked.

This is mostly the case on:

 - form-logic
 - form-repeats
 - form-question-types (i think there's some there)

lesley-morgan:
closes #738 

Fixed references for now, random, once


lognaturel:
Right on! 👍 🎈

Issue #539, opened 2018-02-27
style versionadded directive


adammichaelwood:
![screen shot 2018-02-27 at 9 21 48 am](https://user-images.githubusercontent.com/1776241/36743699-c12beac4-1b9f-11e8-80b8-3586b210b3c2.png)

Not sure exactly how. Maybe something similar to admonition styling? 

(whatever is done should take into account that the directive might not have content aside from the version number)

PR #821, opened 2018-08-22
Document special uri_data parameter used to specify intent data when launching external app

### LINKED ISSUE #775: Document Collect external application special key uri_data
This can be used to set the data for an implicit intent.

https://github.com/opendatakit/collect/pull/2351

lognaturel:
closes #775 

Adds `uri_data` docs and examples.

I also added some more detail on how this appearance relates to intents because I think it's hard to generalize usage without that context.

I removed the XML output for the simple case but left it for the `field-list` case because it's not currently documented anywhere else.

### DIFF
--- a/odk1-src/launch-apps-from-collect.rst
+++ b/odk1-src/launch-apps-from-collect.rst
@@ -47,20 +47,7 @@ XLSForm
 
   integer, counter, Click launch to start the counter app, "ex:org.opendatakit.counter(form_id='counter-form', form_name='Counter Form', question_id='1', question_name='Counter')"
 
-XForm XML
-~~~~~~~~~~~
-
-.. code-block:: xml
-
-  <input appearance="ex:org.opendatakit.counter(form_id='counter-form', form_name='Counter Form', question_id='1', question_name='Counter')" ref="/counter/counter">
-      <label>Click launch to start the counter app</label>
-  </input>
-
-In the examples above, the parameters specified are ``form_id``, ``form_name``, ``question_id`` and ``question_name``. Any number of extra parameters can be specified. The parameter values can be:
-
-  - XPath expressions referring to other fields and including function calls
-  - String literals defined in single quotes
-  - Raw integers or decimals
+In the examples above, the extras specified have names ``form_id``, ``form_name``, ``question_id``, and ``question_name``.
 
 .. _launch-apps-multiple-fields:
 
--- a/odk1-src/launch-apps-from-collect.rst
+++ b/odk1-src/launch-apps-from-collect.rst
@@ -21,7 +21,23 @@ Launching External Apps
 Launching external apps to populate single fields
 ------------------------------------------------------
 
-ODK Collect can launch external applications to populate string, integer or decimal fields using the ``ex:intentString`` appearance. A ``value`` parameter that holds the current value for that field is passed to the application. Since v1.4.3, additional parameters can be specified. The names of these parameters are user defined and ``value`` is the only reserved name. 
+ODK Collect can launch external applications to populate string, integer or decimal fields using the ``ex:intentAction(intentExtras)`` appearance.
+
+This feature configures an Android intent. Intents are messaging objects used to request an action from another app component. Learn more in `the Android docs <https://developer.android.com/guide/components/intents-filters.html>`_.
+
+Collect builds the action using the text immediately after ``ex:`` and before an opening parenthesis (if it exists). For example, in the string ``ex:org.opendatakit.counter(intentExtras)``, the action name is ``org.opendatakit.counter``. An action name must include a namespace, such as ``org.opendatakit`` or ``com.google``. For example, ``ex:org.opendatakit.counter`` is valid but ``ex:counter`` is not.
+
+The parameters defined in the optional parentheses represent extended data ("extras") to be added to the intent. Extras are specified by a comma-delimited list of ``name=value`` pairs. The text to the left of the equals sign represents the extra name and may require a namespace. The text to the right of the equals sign represents the extra value.
+
+The values of extras can be:
+
+- XPath expressions referring to other fields and including function calls
+- String literals defined in single quotes
+- Raw integers or decimals
+
+An extra named ``value`` that holds the current value for the current field is always passed with the intent. Since v1.4.3, additional parameters with user-defined names can be specified. There are two reserved names: ``value`` and ``uri_data``.
+
+Since Collect v1.16.0, the data that the intent operates on can be set by :ref:`using the reserved uri_data parameter <launch-apps-uri-data>`. This is particularly useful for implicit intents such as ``android.intent.action.SENDTO``.
 
 XLSForm
 ~~~~~~~~~
--- a/odk1-src/launch-apps-from-collect.rst
+++ b/odk1-src/launch-apps-from-collect.rst
@@ -81,3 +81,21 @@ XLSForm
 The ``intent`` attribute is only used when the group has an ``appearance`` of ``field-list``. The format and the functionality of the ``intent`` value is the same as above. If bundle of values returned by the external application contains values whose keys match the type and the name of the sub-fields, then these values overwrite the current values of those sub-fields.
 
 The external app is launched with the parameters that are defined in the intent string plus the values of all the sub-fields that are either text, decimal, or integer. Any other sub-field is invisible to the external app.
+
+.. _launch-apps-uri-data:
+
+Specifying a URI as intent data
+---------------------------------
+
+Since Collect v1.16.0, the value for the reserved parameter name ``uri_data`` is converted to a URI and used as the data for the intent. The intent data determines which application to launch when using implicit intents such as `SENDTO <https://developer.android.com/reference/android/content/Intent#ACTION_SENDTO>`_. For example:
+
+``ex:android.intent.action.SENDTO(uri_data='smsto:5555555', sms_body=${message})``
+  Launches a new message in an SMS app with the destination number set to ``5555555`` and the message body set to the contents of the ``message`` field.
+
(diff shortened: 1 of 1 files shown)


odk1-src/launch-apps-from-collect.rst
```diff
@@ -21,7 +21,21 @@ Launching External Apps
 Launching external apps to populate single fields
 ------------------------------------------------------
 
-ODK Collect can launch external applications to populate string, integer or decimal fields using the ``ex:intentString`` appearance. A ``value`` parameter that holds the current value for that field is passed to the application. Since v1.4.3, additional parameters can be specified. The names of these parameters are user defined and ``value`` is the only reserved name. 
+ODK Collect can launch external applications to populate string, integer or decimal fields using the ``ex:intentAction(intentExtras)`` appearance.
+
+Under the covers, this feature configures an intent, a messaging objects used to request an action from another app component. For more details on intents,  refer to `these Android docs <https://developer.android.com/guide/components/intents-filters.html>`_.
```
adammichaelwood: I think "under the hood" might be more clear than "under the covers".

adammichaelwood: (Or "behind the scenes")

adammichaelwood: > an intent, a messaging objects used to request an action

I *think* this is supposed to be `object` not `objects`.

But I think even then, the multiple commas make it a little unclear.
Maybe something like...

```
Under the hood, this feature configures an `intent`_. An intent is a messaging object used to...

.. _intent: https://developer.android.com/guide/components/intents-filters.html
```
(And then remove  "For more details...")

(Could put the link on the explanatory sentence instead of the word "intent" if that feels better.)

odk1-src/launch-apps-from-collect.rst
```diff
@@ -21,7 +21,21 @@ Launching External Apps
...
+
+Collect uses the text immediately after ``ex:`` and before an opening parenthesis if it exists as the action the intent is built with. This action must be namespaced. That is, ``ex:org.opendatakit.counter`` is valid but ``ex:counter`` is not.
```
adammichaelwood: > if it exists

Does that refer to the opening parenthesis?

I think that notes would be more clear if that was in parens.

``` 
an opening parenthesis (if it exists)
```

(And if it doesn't refer to the opening parens, the sentence needs more help.)

adammichaelwood: > Collect uses the text immediately after ``ex:`` and before an opening parenthesis if it exists as the action the intent is built with.

Maybe it's because I'm not Java-y, but I found this confusing.

Think about:

```
The action is built using the text immediately after ``ex:``, up to (but not including) any opening parenthesis. For example, in the string ``ex:org.opendatakit.counter(something)``, the action name is ``org.opendatakit.counter``.

An action name must include a namespace, such as ``org.opendatakit`` or ``com.google``. For example, ``ex:org.opendatakit.counter`` is valid but ``ex:counter`` is not.

File: odk1-src/launch-apps-from-collect.rst [hunk 3]
```diff
@@ -21,7 +21,21 @@ Launching External Apps
...
+
+The parameters defined in the optional parentheses represent extended data ("extras") to be added to the intent. The value to the left of the equals sign represents the extra name and may require a namespace. The value to the right of the equals sign represents the extra value. Multiple extras can be configured using a comma-separated list. The extra values can be:
```
adammichaelwood: A suggestion...

```
Inside the optional parentheses you can define extended data which can be added to the intent. These are called `extras`_ and are represented by a comma-delimited list of ``name=value`` pairs. For example ``(please=think, of=examples, more=relevant, than=this)``. 

The values of extras can be:

- XPath expressions referring to other fields and including function calls
- String literals defined in single quotes
- Raw integers or decimals

```

adammichaelwood: PS: 
Unlike markdown, bullet lists in RST should not be indented. Doing so will (sometimes) cause them to be wrapped in a blockquote.

```
- Bullets
- Like
- This

 - Not
 - Like
 - This
```

odk1-src/launch-apps-from-collect.rst
```diff
@@ -21,7 +21,21 @@ Launching External Apps
...
+
+  - XPath expressions referring to other fields and including function calls
+  - String literals defined in single quotes
+  - Raw integers or decimals
+
+An extra named ``value`` that holds the current value for the current field is always passed with the intent. Since v1.4.3, additional parameters with user-defined names can be specified. There are two reserved names: ``value`` and ``uri_data``.
+
```
adammichaelwood: Any reason these two references (above and below this comment) to versions aren't using the [versionadded](http://www.sphinx-doc.org/en/stable/markup/para.html#directive-versionadded) or [versionchanged](http://www.sphinx-doc.org/en/stable/markup/para.html#directive-versionchanged) directives?

(It's okay if the reason is that we still haven't styled versionadded (#539) and we just don't like how it looks...)

odk1-src/launch-apps-from-collect.rst
```diff
@@ -31,20 +45,7 @@ XLSForm
...
-  <input appearance="ex:org.opendatakit.counter(form_id='counter-form', form_name='Counter Form', question_id='1', question_name='Counter')" ref="/counter/counter">
-      <label>Click launch to start the counter app</label>
-  </input>
-
-In the examples above, the parameters specified are ``form_id``, ``form_name``, ``question_id`` and ``question_name``. Any number of extra parameters can be specified. The parameter values can be:
-
-  - XPath expressions referring to other fields and including function calls
-  - String literals defined in single quotes
-  - Raw integers or decimals
+In the examples above, the parameters specified are ``form_id``, ``form_name``, ``question_id`` and ``question_name``. Any number of extra parameters can be specified.
```
adammichaelwood: I would *think about* making that list bullets instead of inline. 
If you don't.... at least you need to include the serial comma.

cf: https://docs.opendatakit.org/docs-style-guide/#serial-comma

> A bulleted list is often more clear than an inline list.

odk1-src/launch-apps-from-collect.rst
```diff
@@ -81,3 +82,16 @@ XLSForm
 The ``intent`` attribute is only used when the group has an ``appearance`` of ``field-list``. The format and the functionality of the ``intent`` value is the same as above. If bundle of values returned by the external application contains values whose keys match the type and the name of the sub-fields, then these values overwrite the current values of those sub-fields.
 
 The external app is launched with the parameters that are defined in the intent string plus the values of all the sub-fields that are either text, decimal, or integer. Any other sub-field is invisible to the external app.
+
+.. _launch-apps-uri-data:
+
+Specifying a URI as intent data
+---------------------------------
+
+Since Collect v1.16.0, it is possible to specify a URI that will be used as the data for the intent using the reserved ``uri_data`` parameter. This URI should include a scheme so that external applications will be able to identify the type of content to act on. That is, ``mailto:example@example.com`` is a valid URI but ``example@example.com`` is not.
```
adammichaelwood: See comment above re: versionadded/versionchanged directive.

odk1-src/launch-apps-from-collect.rst
yanokwa: * "Under the covers" is colloquial. You can just drop it.
* s/objects/object
* I think you have too many spaces before refer.

odk1-src/launch-apps-from-collect.rst
yanokwa: I think you need a comma after parenthesis and before exists.

odk1-src/launch-apps-from-collect.rst
adammichaelwood: > it is possible to specify a URI that will be used as the data for the intent using the reserved ``uri_data`` parameter. 

This sentence confused me. I think it is the `used...using` and (i don't know what to call it) daisy chain handoff of subject/predicate.

(Also, the style guide prefers "You can..." to "It is possible..." cf: https://docs.opendatakit.org/docs-style-guide/#direct-address)

Moreover. I don't know, it was just kind of unclear. I think it is the front-loading of information (URI requiring a scheme before any examples).

Maybe something like...

```
you can specify a URI in the extra parameters using the reserved name ``uri_data``. 
This is particularly useful when using an implicit intent such as `sendto `_, which launches different applications for different types of URIs. 

.. _sendto: https://developer.android.com/reference/android/content/Intent#ACTION_SENDTO

For example: 

``ex:android.intent.action.SENDTO(uri_data='smsto:5555555', sms_body=${message})``
  Launches a new message in an SMS app with the destination number set to ``5555555`` and the message body set to the contents of the ``message`` field.

``ex:android.intent.action.SENDTO(uri_data='mailto:example@example.com?subject=${subject}&body=${message})``
  Launches a new message in an email app with destination address set to ``example@example.com``, the subject set to the contents of the ``subject`` field and the body set to the contents of the ``message`` field.

``ex:android.intent.action.DIAL(uri_data='tel:5555555')``
  Launches a phone dialer with the number ``5555555`` as the number to dial.

Notice that the URI must include a `scheme`_, such as ``mailto:`` or ``https://``.

.. _scheme: https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml

```

Notice that I changed your list to a definition list. In general if a list has an item:value format, definition lists are more clear.

If you want to keep the bullets, I prefer an em-dash (wrapped in spaces) instead of a colon because I think it is more clear (since it provides more horizontal separation).

(I don't think we've solidified that preferences in the style guide, though. So if you feel strongly about colons, that's probably fine.)

lognaturel:
Thanks very much @yanokwa and @adammichaelwood! I made nearly all the changes you suggested and forced push to maintain the history. If you have any questions about why I may not have adopted a certain suggestion, please ask. Thanks in particular to @adammichaelwood for pointing out the definition list -- I wanted something like that but couldn't quickly find it.

PR #826, opened 2018-08-24
Add version information for Aggregate

### LINKED ISSUE #770: Specify versions of Aggregate stack for local deployment
The Aggregate docs page about [Installing on Tomcat](https://docs.opendatakit.org/aggregate-tomcat/) should specify a minimum compatible versions of the Aggregate stack for local deployments, which is composed of:
- Java JDK 8u171/ 8u172
- Apache Tomcat 8.5
- Database options:
  - PostgreSQL 9.6
  - MySQL 5.7.22

yanokwa:
Closes #770

* Clean up whitespace
* Use sentence case
* Move security note to the security section
* Add minimum version information. 

I start with v1.4.13 because that's when MS/Azure SQL, Tomcat 8, MySQL 5.7, and PostgreSQL 9.4 were added, so it's a [known checkpoint](https://github.com/opendatakit/aggregate/releases/tag/v1.4.13_Production). I toyed with the idea of adding a v1.5.0 header, but I'm pretty sure the minimum versions are the same.

I say _should_ in my prose because we haven't verified.

### DIFF
--- a/odk1-src/aggregate-tomcat.rst
+++ b/odk1-src/aggregate-tomcat.rst
@@ -55,6 +55,10 @@ Encrypted forms can be used in conjunction with either of the first two suggesti
 
 If you are not using encrypted forms and are handling sensitive data, a computer security specialist should review your system and your security procedures. 
 
+.. note::
+
+  Use of an SSL and `https` is recommended for any deployment accessed from the internet.
+
 Database Systems
 ~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


ggalmazor:
Regarding adding a 1.5.0 column, while it's true that we're officially working with PostgreSQL 9.6 (VMs, Docker, etc.), I agree that the versions for 1.4.13 should work. Unless we can verify they don't, I also prefer to have less columns in that table to make it less confusing.

PR #838, opened 2018-09-05
Central is now in beta, but is ready for production use

yanokwa:
Neutral language about how early Central is without having alpha/beta everywhere


clint-tseng:
it's up to y'all! i abstain from judgement.

PR #867, opened 2018-09-27
Dockerized workflow

ggalmazor:

This PR includes a Dockerfile container definition that produces a Docker container with all the tools we need to work with the docs.

The container's entrypoint is `make`, which enabled using this container as a direct substitute of running make locally. The only requirement is to bind the docs repo folder to the container's `/mnt` volume:

| normal command | becomes |
| --- | --- |
| `make odk1` | `docker run -it -v /path/to/docs:/mnt odk-docs odk1` |
| `make odk1-pdf` | `docker run -it -v /path/to/docs:/mnt odk-docs odk1-pdf` |
(etc.)

This PR also adds a new `make build-env` task to build the Docker container.

Also in this PR, two new tasks to run sphinx autobuild:
- It watches the sources folder and rebuilds the docs (only what's changed)
- It server the docs with a live-reload server that will refresh your browser whenever the automatic build process ends.

Instructions to start the autobuild task:
- Execute `docker run -it -v /path/to/docs:/mnt -p 8080:8080 -d --name odk-docs odk-docs odk1-autobuild`
- Open http://localhost:8080 in your browser
- If you want to see the log messages while sphinx-autobuild works, then run `docker logs -f odk-docs`
- To stop the autobuild process run `docker kill odk-docs`

The only requirement to use this is to have installed Docker.


requirements.txt
```diff
@@ -1,23 +1,43 @@
 alabaster==0.7.10
```
yanokwa: All these new requirements will probably add to the CircleCI build time, but it's probably not a thing I should be complaining about.

yanokwa:
One thing we should probably add as part of those instructions is how users know what is going wrong. Maybe that bash/batch file can wrap a docker log file or something?

Makefile
```diff
@@ -19,6 +19,15 @@ help:
 
 .PHONY: help Makefile
 
+build-env:
```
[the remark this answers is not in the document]
ggalmazor: Having the odk prefix on the image name could prevent collisions with other images, but I guess we could do that.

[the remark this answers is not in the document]
yanokwa: Fair enough. Then maybe `odk-docs`? Reads better.

requirements.txt
ggalmazor: I think they're transitive dependencies, although I'm not an experienced Pythonite.

Basically, I've run `pip freeze > requirements.txt` after installing sphinx-autobuild.

yanokwa: https://nvie.com/posts/pin-your-packages/ suggests that I'm wrong and we should pin the dependencies.

ggalmazor: Anyway, I see that `pip freeze` has arguments. Maybe adding `--user` gets less stuff in.

ggalmazor: OK, `--user` gives nothing, and `--local` gives the same output. I don't know if we can try anything else.

It feels like it's OK, though.

ggalmazor: Nah. I wish I knew more Python! I'm sorry for all the confusion. With the new requirements it doesn't work right. Let me fix it

PR #871, opened 2018-09-27
Briefcase split select multiples

yanokwa:
Addresses https://github.com/opendatakit/briefcase/pull/649

* Match shipping code

### DIFF
--- a/odk1-src/briefcase-using.rst
+++ b/odk1-src/briefcase-using.rst
@@ -186,7 +186,7 @@ Export forms to CSV
     - :guilabel:`Ovewrite existing files` enables overwriting form instance data in the output files. The default behavior is to append data.
     - :guilabel:`Split select multiples` enables splitting select multiple fields. Enabling this setting will create an extra output column per select choice, with a `1` if the choice was selected, or `0` otherwise.
 
-      This will affect only select fields that have a static list of choices, as opposed to selects that use dynamic `<itemset>` choices.
+      This only affects select fields without a choice filter and that are not from an external file.
 
     - :guilabel:`Pull before export` enables trying to pull the selected forms in case there are new form instances to be exported.
 
--- a/odk1-src/briefcase-using.rst
+++ b/odk1-src/briefcase-using.rst
@@ -186,7 +186,7 @@ Export forms to CSV
     - :guilabel:`Ovewrite existing files` enables overwriting form instance data in the output files. The default behavior is to append data.
     - :guilabel:`Split select multiples` enables splitting select multiple fields. Enabling this setting will create an extra output column per select choice, with a `1` if the choice was selected, or `0` otherwise.
 
-      This will affect only to select fields that have a static list of choices, as opposed to selects that use dynamic `<itemset>` choices.
+      This will affect only select fields that have a static list of choices, as opposed to selects that use dynamic `<itemset>` choices.
 
     - :guilabel:`Pull before export` enables trying to pull the selected forms in case there are new form instances to be exported.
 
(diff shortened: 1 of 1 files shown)


odk1-src/briefcase-using.rst
```diff
@@ -184,9 +184,9 @@ Export forms to CSV
 
     - :guilabel:`Export media files` enables exporting media files into the chosen export directory
     - :guilabel:`Ovewrite existing files` enables overwriting form instance data in the output files. The default behavior is to append data.
-    - :guilabel:`Split multiple selects` enables splitting multiple select fields (a.k.a. choice lists). Enabling this setting will create an extra output column per select option, with a `1` if the option was selected, or `0` otherwise.
+    - :guilabel:`Split select multiples` enables splitting select multiple fields. Enabling this setting will create an extra output column per select choice, with a `1` if the choice was selected, or `0` otherwise.
 
-      This will affect only to select fields that have a static list of options, as opposed to selects that use dynamic `<itemset>` options.
+      This will affect only to select fields that have a static list of choices, as opposed to selects that use dynamic `<itemset>` choices.
```
lognaturel: I don't think users are likely to know what this means. Maybe "select multiples without a choice filter?"

@ggalmazor, the determination is based on whether there's an `itemset` defined, right? Unfortunately, neither of the two unofficial ways to use external CSVs ([`search()`](http://xlsform.org/en/#dynamic-selects-from-pre-loaded-data) and ["external selects"](http://xlsform.org/en/#external-selects) use `itemset`s so I believe those would be split.

ggalmazor: Thanks for the clarification, @lognaturel! I didn't know about that.

JR's API provides two separate methods to get the choices list. One works with static items and the other works with dynamic choices, and we use only the one that returns the static choices.

I've tried the new split select multiples with a form that had external instances and I've verified that it doesn't split those selects because the static choices list is empty. This is the form: [external_instances.zip](https://github.com/opendatakit/docs/files/2427122/external_instances.zip)

ggalmazor:
[external_instances.zip](https://github.com/opendatakit/docs/files/2427122/external_instances.zip)

odk1-src/briefcase-using.rst
ggalmazor: Also related: in the XSLForms spec you've linked, they describe a  `select_one_external`. Fields of this type would never be split because they're not select multiples.

ggalmazor: Also also related: forms that use external csvs combined with `search()` appearance don't really work (I'll create an issue on briefcase to track this).

Take this form, for instance: [search_and_select.zip](https://github.com/opendatakit/docs/files/2427166/search_and_select.zip)

You get this CSV export:
```csv
SubmissionDate,fruit,fruit/name_key,note_fruit,meta-instanceID,KEY
28-sep-2018 9:21:04,mango oranges,0,"",uuid:4151c9a4-0f86-47a1-b434-9f69f0309028,uuid:4151c9a4-0f86-47a1-b434-9f69f0309028
```

- There's actually a static item:

  ```xml
  <select appearance="search('fruits')" ref="/search_and_select/fruit">
    <label>Choose a fruit</label>
    <item>
      <label>name</label>
      <value>name_key</value>
    </item>
  </select>
  ```
- This static item produces the extra `name_key` column in the output CSV.
- At least we're not getting the choices in the external CSV file as static choices.

lognaturel: > they describe a `select_one_external`. Fields of this type would never be split because they're not select multiples.

Ah, great! I always forget that this odd feature is only supported for select ones. That's probably a good thing.

lognaturel: Thanks for doing that exploration and filing #656. How about this for the wording: "This only affects select fields without a choice filter and that are not from an external file" I think that would have more meaning to users. 

Also note "affect to" -> "affect" (no se admite la forma "afectar a" en inglés 😊)

ggalmazor: Thanks, @lognaturel! I've added a pair of commits with your suggested changes.

lognaturel:
This still has a WIP label but I think it's ready to merge.

PR #872, opened 2018-09-27
Update security privacy

### LINKED ISSUE #832: Update Security and Privacy with Briefcase usage data gathering
Briefcase gathers anonymous usage data. We should update the [Security and Privacy](https://docs.opendatakit.org/security-privacy/) section to add a note about this.

This is related to changes from https://github.com/opendatakit/briefcase/issues/357

yanokwa:
Closes #832 

Update to current practice and re-arranging so it's easier to read. The big changes are 
* Combining all the Aggregate things in one section and moving the bits that are focused on encryption into the general encryption section so it's easier to read.
* Dropped the Google Account section because it's no longer relevant for Aggregate.
* Update the rest of the tools language to match current practice
* Creating a new cross-tool concerns section

I think the easiest way to parse these changes is to read the docs at https://docs.opendatakit.org/security-privacy and compare it to what this PR has.

### DIFF
--- a/odk1-src/security-privacy.rst
+++ b/odk1-src/security-privacy.rst
@@ -36,7 +36,7 @@ All our installers, programs, source code, and documentation are provided AS-IS 
 Communication Channels
 ---------------------------
 
-None of the downloadable ODK software transmits or communicates any information back to us and the software we have written does not have any mechanisms that might allow us to access or control your devices or systems.
+Outside of usage analytics and crash reports, ODK software does not transmit or communicate any information back to ODK's maintainers. Further, the software we have written does not have any mechanisms that might allow us to access or control your devices or systems.
 
 There is always the possibility that hackers can discover and exploit deficiencies or bugs in our software or in 3rd-party libraries to access or control your devices or systems.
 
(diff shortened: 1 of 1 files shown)


odk1-src/security-privacy.rst
```diff
@@ -63,7 +67,7 @@ Google Play Store
 
 Downloads from the Google Play Store are compiled into aggregated usage statistics.
 
-Crash reports you elect to send are provided to us as anonymous crash reports. By design, these do not contain survey field values or other device- or user- specific data.
+Crash reports you elect to send are provided to us as anonymous crash reports. By design, these do not contain device or user specific data.
```
lognaturel: "as anonymous crash reports" -> anonymously

> these do not contain device or user specific data.

I don't think this is so accurate. Sometimes crash reports include network information which could be identifying. I also don't know what "device-specific data" is. We do see what devices get what crash. That could potentially be identifying.

yanokwa: Google Play Store reports are anonymized by default. It's Firebase reports that are not.

yanokwa:
I disagree that we should put the password warning here. We should instead make the warning on Aggregate more severe. I'll file an issue there. I will also file an issue to put the security and privacy document somewhere on the website.

PR #875, opened 2018-10-04
Correct sms submission custom delimiter example

alxndrsn:
I think this corrects the example of an SMS form submission with custom delimiter.  The text implies that fields should be delimited by `+`, but the example shows a `|` (pipe).  Pipe wouldn't be a great delimiter to pick, as it's not included in the [standard GSM SMS character set](https://en.wikipedia.org/wiki/GSM_03.38#GSM_7-bit_default_alphabet_and_extension_table_of_3GPP_TS_23.038_/_GSM_03.38).


lognaturel:
🎉 Thank you!!

PR #882, opened 2018-10-07
Update the command to run the spelling checker.

### LINKED ISSUE #881: Spelling checker command is incorrect
In https://docs.opendatakit.org/docs-tech-guide/, the recommended command to run the spelling checker is:

    `$ sphinx-build -b spelling src build/spelling`

This command doesn't work, because there is no `src` directory.  The command should specify the `odk1-src` or `odk2-src` directory.

zestyping:
Also fixes a few nearby grammar errors.

Closes #881.

This updates the recommended command to run the spelling checker.

None.

Nothing.

The spelling checker does not run at all with the current command.  The new command runs the spelling checker, though it generates a lot of complaints about misspelled words.  The next step is to add the list of all the misspelled words to the `spelling_wordlist.txt` file as recommended by https://docs.opendatakit.org/docs-tech-guide/.

### DIFF
--- a/shared-src/docs-tech-guide.rst
+++ b/shared-src/docs-tech-guide.rst
@@ -682,32 +682,38 @@ Write and edit files in your favorite code editor.
 
 .. _spell-check:
 
-Spell check your work
-~~~~~~~~~~~~~~~~~~~~~~
+Check your spelling
+~~~~~~~~~~~~~~~~~~~
 
 After making changes to the docs,
-you need to run the spell checker.
-To run the spell checker:
+run the spelling checker with:
 
 .. code-block:: console
 
-  $ sphinx-build -b spelling src build/spelling
+  $ sphinx-build -b spelling odk1-src build/spelling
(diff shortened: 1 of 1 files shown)


shared-src/docs-tech-guide.rst
```diff
@@ -682,32 +682,38 @@ Write and edit files in your favorite code editor.
...
+~~~~~~~~~~~~~~~~~~~
 
 After making changes to the docs,
-you need to run the spell checker.
-To run the spell checker:
+run the spelling checker with:
+
+.. code-block:: console
+
+  $ sphinx-build -b spelling odk1-src build/spelling
```
ankita240796: Hi @ping, thanks for the PR.

This command produces spelling errors since the `spelling_wordlist.txt` file is not present in `odk1-src`. You can update the command to `make odk1-spell check` which will run all the commands needed to copy the required files before running the spell checker.

PR #891, opened 2018-10-16
Google Cloud now requires a credit card or bank account

### LINKED ISSUE #850: Billing account now required for Aggregate mapping
According to https://forum.opendatakit.org/t/odk-aggregate-map-visualize-cant-open-correctly/14828 and https://developers.google.com/maps/billing/important-updates#billing_changes, all projects must be associated with a billing account. 

It would be good to add a warning to https://docs.opendatakit.org/aggregate-deployment-planning/ and maybe https://docs.opendatakit.org/oauth2-service/

yanokwa:
Closes #850

Add language in the places where a user would look to install Aggregate on GAE.

It would be good to ask if the community has been able to set up new servers without a credit card.

### DIFF
--- a/odk1-src/aggregate-app-engine.rst
+++ b/odk1-src/aggregate-app-engine.rst
@@ -14,7 +14,9 @@ Installing on Google App Engine
   
   - You will need a Gmail account to use App Engine. This Gmail account will be the owner of the Google Cloud Platform project under which your App Engine will execute.
 
-  - Read the `Google Cloud Platform term of service <https://cloud.google.com/terms/>`_.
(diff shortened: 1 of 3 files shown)


lognaturel:
@yanokwa, you said you spoke to someone at Google about this, right? So it seems really unlikely anyone is able to set up without a credit card?

Do you want to post a support question about this and see what others' experience has been or just merge it?

yanokwa:
I have spoken to folks at Google and read their docs. Both say that you need a credit card. Please merge and I'll write a support post point to the docs and see if folks have had a bad experience.

PR #893, opened 2018-10-19
remove misleading note

danbjoseph:
see the discussion at:
https://forum.opendatakit.org/t/value-produced-by-decimal-date-time/15751

should this edit add the additional detail about "it's the format used by Excel but off by 70 years" instead of just removing the sentence?

### DIFF
--- a/odk1-src/form-operators-functions.rst
+++ b/odk1-src/form-operators-functions.rst
@@ -646,7 +646,7 @@ Converting dates and time
   
 .. function:: decimal-date-time(dateTime)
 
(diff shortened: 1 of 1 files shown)


lognaturel:
I think we can start by merging this and then perhaps file an issue to extract the part that starts with "To convert from epoch..." at https://docs.opendatakit.org/form-audit-log/#timestamps somewhere else so it can be linked to from both places.

yanokwa:
I've punted on filing an issue because...
1. There isn't a great place to put it.
2. The conversion is easily Googleable. 

Instead, I sent in https://github.com/opendatakit/docs/pull/896

PR #903, opened 2018-10-25
Issue 902 aggregate versions

### LINKED ISSUE #902: Update info about form versions and Aggregate
We should explain what version formats are expected in Aggregate, how comparing two versions work, and what can be expected when downloading blank forms from Aggregate (losing zero padding)

ggalmazor:
closes #902

In this PR:
- Fix a language nuance around blank form versions 
- Explains how blank forms downloaded from Aggregate can (and will) be different from the original blank forms that are uploaded

### DIFF
--- a/odk1-src/aggregate-forms.rst
+++ b/odk1-src/aggregate-forms.rst
@@ -52,8 +52,15 @@ From the :ref:`form list <aggregate-view-blank-form-list>`,  click on form's :gu
 .. image:: /img/aggregate-use/xml-viewer.*
    :alt: Image showing xml viewer for form.
 
+.. note::
+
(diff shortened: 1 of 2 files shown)


odk1-src/aggregate-forms.rst
```diff
@@ -53,6 +53,11 @@ From the :ref:`form list <aggregate-view-blank-form-list>`,  click on form's :gu
    :alt: Image showing xml viewer for form.
 
 .. _aggregate-toggle-form-download:
+
```
yanokwa: Maybe wrap this in a note?

ggalmazor: Done! I've also fixed the label on line 55, which should be below the text I've added :P (RST is hard!)

Issue #938, opened 2019-01-08
Explain how to filter with range widgets and avoid errors with selected()


ggalmazor:
We have found that users can mistakenly use the `selected()` function with a range widget to filter a secondary itemset, which won't usually work.

We should explain how to filter secondary itemsets with range data and make it explicit that the `selected()` function is intended to be used with select widgets.

PR #939, opened 2019-01-08
Update the range widget to explain how it can be used in filters

ggalmazor:
addresses #938

This should help users to avoid errors by mistakenly using the `selected()` function with range widgets in itemset filters, which won't usually work. The `selected()` function will only work with text data from other `select1` widgets.


yanokwa:
Also, @ggalmazor, I do this too, but we should be careful with branches against master as owners of a repo. It means that long lived branches like this live in other people's repo and it feels a little dirty.

PR #943, opened 2019-01-21
age should be a variable

yanokwa:
Minor fix from a user comment at https://forum.opendatakit.org/t/unexpected-behaviour-of-the-now-function/17302


yanokwa:
Bypassing this review because it's the safest PR ever and it's been 17 days :)

PR #947, opened 2019-01-24
clarification on aggregate

khmermenten:
closes #

addresses #

Added clarification for what an aggregate server is/what it's used for in the getting started user guide.

### DIFF
--- a/odk2-src/getting-started-2-user.rst
+++ b/odk2-src/getting-started-2-user.rst
@@ -38,7 +38,7 @@ A major goal of the 2 tools was to eliminate the need for any software engineeri
 Joining a device to an Existing Aggregate Server
 ------------------------------------------------------
 
-ODK uses a server called Aggregate to distribute blank forms, accept finalized forms, publish data, and more. We will use an Aggregate Server to get the data for a sample application.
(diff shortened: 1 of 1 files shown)


odk2-src/getting-started-2-user.rst
```diff
@@ -38,7 +38,7 @@ A major goal of the 2 tools was to eliminate the need for any software engineeri
 Joining a device to an Existing Aggregate Server
 ------------------------------------------------------
 
-ODK uses a server called Aggregate to distribute blank forms, accept finalized forms, publish data, and more. We will use an Aggregate Server to get the data for a sample application.
+Aggregate is a server that distributes blank forms, accepts finalized forms, publishes data, and more. We will use an Aggregate Server to get the data for a sample application for ODK Tables and ODK Survey.
```
ggalmazor: You might want to remove the last part about "for ODK Tables and ODK Survey", since support for those is being dropped in the next Aggregate v2.0 release (Feb 4th)

ggalmazor: Well, I guess you could remove the whole parragraph unless you want to specify that it should be an Aggregate v1.x version (legacy)

Issue #603, opened 2018-03-23
better solution for shared template and css files


adammichaelwood:
Currently odk1 and and odk2 have duplicated css and template customizations.

There are some things (like the color scheme changes) that need to be odk1/odk2 specific, but I think we want to keep shared styles/customizations shared.

PR #948, opened 2019-01-24
Issue 944 Aggregate v2.0 changes

### LINKED ISSUE #944: Make changes for Aggregate v2.0
We should document the changes for the upcoming Aggregate v2.0 version, including:
- Remove info about ODK Tables, and Google Fusion Tables, and Google Account credentials, which are no longer supported.
- Remove info about deploying in AppEngine, which is no longer supported
- Explain how Aggregate can be deployed in Digital Ocean and other cloud providers
  - Explain how to use the Aggregate updater tool 
- Explain how temporal filters work and what values should be used to filter temporal data

ggalmazor:
closes #944

Review all Aggregate docs and:
- Remove any mention to AppEngine support
  - Gather all removed AppEngine information into a new page
  - Tag as "legacy" two pages that were already focused on AppEngine
- Remove any mention of publishers that don't exist anymore
- Simplify some guides by reducing the number of suggested options or by linking to better or more complete guides
- Do a general review of language


ayushi-jain97:
I want to unsubscribe but it seems no info is provided for the same in the
mail.

adammichaelwood:
> /home/circleci/work/tmp1-src/aggregate-app-engine-legacy.rst:89:Explicit markup ends without a blank  line; unexpected unindent.

I've seen this before, but I don't remember what the problem is.
I'll Google around and see if anything rings a bell.

> How could I get a table cell's (long) text to have some breaks? I'm getting tables that overflow horizontally, which is not very nice.

This is the single worst thing about rst.

The answer is to use the [line block syntax](http://docutils.sourceforge.net/docs/user/rst/quickref.html#line-blocks).

```rst

.. csvtable::
  :header: Column1, Column2

   "|This is 
| in column1", "|This is
| in column2"

```

That will normally also indent the text in that cell, but I *think* I added some CSS to put it back the way it should be.

I've done this in a few places, so if you search around you'll find more realistic examples.


> Is there any way to present instructions for different operating systems in a nicer way? I remember you mentioned something about this, but I can't find it anywhere...

TABS!

Look at the contributors technical guide for realistic examples.

```rst

.. tabs::
   
   .. group-tab:: Windows

   Windows instructions

   .. group-tab:: Mac

   Mac Instructions

   .. group-tab:: Linux

   Linus Instructions

```

ggalmazor:
> Do those line breaks need to be there? They don't look semantic, but rather formatting-driven. Did we not solve the problem of too-wide tables?

Indeed. I suspect that it's making super wide tables because the text doesn't have enough natural breakpoints. That's why I'm trying to insert artifical breaks everywhere.

ggalmazor:
OK, I've solved the problem for now, but I'm open to restructuring that table. The current (crazy) code is this:

```rst
.. csv-table::
  :header: "Variable", "Description", "Example value"

  "| `{{pubKey}}`",                 "| SSH public key to access
  | the machine once it's running",                                 "| ssh-rsa AAAAB3NzaC1y...
  | ...sn3PMkIj user@servername"
  "| `{{aggregateVersion}}`",       "| ODK Aggregate version you
  | want to deploy",                                                "| v2.0.0"
  "| `{{domain}}`",                 "| The domain under which you
  | will be serving ODK Aggregate",                                 "| aggregate.opendatakit.org"
  "| `{{forceHttps}}`",             "| Force using the HTTPS
  | schema when producing links.
  |
  | **Always set this to true**",                                   "| true"
  "| `{{httpPort}}`",               "| The port http port number
  | to be used.
  |
  | **Always set this to 80**",                                     "| 80"
  "| `{{aggregateWarUrl}}`",        "| The URL to be used to
  | download ODK Aggregate.
  |
  | **Always set this to**
  | **https://github.com/opendatakit**
  | **/aggregate/releases/download**
  | **/{{aggregateVersion}}**
  | **/ODK-Aggregate-**
  | **{{aggregateVersion}}.war**",                                  "| https://github.com/opendatakit
  | /aggregate/releases/download
  | /v2.0.0/ODK-Aggregate-v2.0.0.war"
```
image

Some details have to be dealt with (removing the link from the email and the first part of the last example URL), but I'm happy with it as it is.

The tabs worked wonderfully, by the way :)

adammichaelwood:
>  I suspect that it's making super wide tables because the text doesn't have enough natural breakpoints.

There MUST be some CSS and/or JS solution to sane table sizing. I remember looking into this a while ago, but I don't recall my findings.

ggalmazor:
I'll try to do some research. How would could I go about adding some custom css to that page to try things out?

adammichaelwood:
We don't have a special way to add CSS to one specific page. (Though, since raw HTML works, you *could* just add a `<style>` block. But don't do that....)

Custom CSS goes in:
docs/odk1-src/_static/css/custom.css
docs/odk2-src/_static/css/custom.css

(Side note, these files need to be combined and moved into shared src. - cf. #603 )

Experiment in there and build locally to see what works.
Maybe we should file an issue specifically about fixing table width?

adammichaelwood:
> /home/circleci/work/tmp1-src/aggregate-app-engine-legacy.rst:89:Explicit markup ends without a blank line; unexpected unindent.

Things to try that *might* solve the problem:

 - Does adding an extra line break before or after the entire toctree block help?
 - Does this work:

```rst

Installing Aggregate
--------------------

: 

 .. toctree::
  :maxdepth: 1

   aggregate-app-engine

: 

Aggregate Preferences
---------------------
```

 - What about adding section labels?

```rst

.. _legacy-installing-aggregate:
Installing Aggregate
--------------------

 .. toctree::
  :maxdepth: 1

   aggregate-app-engine

.. _legacy-aggregate-preferences:

 Aggregate Preferences
---------------------
```

Things which don't solve the problem, but provide more information:

 - Does it occur if you add a second page to the toc list?
 - Does it still occur if you remove the `maxdepth` attribute?
 - Does it occur if there is a paragraph of text after the the toctree?

ggalmazor:
@adammichaelwood I have news about the issue with the table size:
- For some reason, the theme has the following class:

  ```css
  .wy-table-responsive table td, .wy-table-responsive table th {
      white-space: nowrap;
  }
  ```
  This looks like a very bad idea. The white-space rule should be `normal`.

- Adding `word-break: break-all` will break content in any place, which is what I need for my table.

  Unfortunately, this can't be applied everywhere because it will break words anywhere and we don't want that in all tables that can use natural breaking (using whitespaces) of sentences.

  How can I apply a css class or style to just one table?

ggalmazor:
@adammichaelwood, regarding the `toctree` issue, I was having problems because the previous section had some syntactic errors! (embarrasing, sorry!)

ggalmazor:
There you go, @yanokwa :D

It's a bunch of commits. I hope it's more manageable now. I recommend setting the diff view to ignore whitespaces and such.

odk1-src/aggregate-app-engine-legacy.rst
```diff
@@ -14,61 +14,13 @@ Remember that all the information in this page and others linked here refers onl
...
-.. _aggregate-deployment-ongoing-support:
-
-Ongoing support
-~~~~~~~~~~~~~~~~
-
-Tomcat deployments require periodic backups of your data. If data security is a concern, you should have a system administrator or database administrator periodically review logs and look for malicious activity.
-
-.. _aggregate-deployment-availability:
-
-Availability
```
ggalmazor: create-backup

odk1-src/aggregate-tomcat.rst
```diff
@@ -33,143 +34,96 @@ Database systems
...
+    .. group-tab:: Ubuntu
 
-   - For Microsoft SQL Server or Azure SQL Server, you should configure these with UTF-8 character sets and to use Windows authentication. When using Windows authentication, the user under which the webserver executes must be granted permissions to access the SQL Server instance. The install wizard for ODK Aggregate will produce a :file:`Readme.html` file that contains additional information on how to complete the configuration of the database and webserver service.
+      .. code-block:: console
 
-5. Download and install `ODK Aggregate <https://github.com/opendatakit/aggregate/releases/latest>`_. Select the latest release for your operating system.
+        sudo apt-get install postgresql-10
 
-   .. note::
+7. Install `ODK Aggregate <https://github.com/opendatakit/aggregate/releases/latest>`_. Select the latest release for your operating system.
```
ggalmazor: Should we also mention the Aggregate-CLI as an alternative method to install Aggregate on a server?

PR #989, opened 2019-03-15
Document front appearance for barcode widget

### LINKED ISSUE #853: Add front appearance for barcode
As of Collect v1.17.0, the `front` appearance can be used with a barcode question to force the front camera to be used. More info and sample form at https://github.com/opendatakit/collect/pull/2453

This should be added to https://docs.opendatakit.org/form-question-types/#barcode-widget

grzesiek2010:
closes #853

It documents front appearance for barcode widget.

### DIFF
--- a/odk1-src/form-question-types.rst
+++ b/odk1-src/form-question-types.rst
@@ -1743,8 +1743,7 @@ Uploads any file from the device to the form.
 Barcode widget
 ----------------
 
-Scans, decodes, and captures the content of a barcode,
-using the device camera.
+Scans, decodes, and captures the content of a barcode, using the device camera.
 
 The following barcode formats are supported:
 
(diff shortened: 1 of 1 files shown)


lognaturel:
The images are quite large. Other question type images are about 28kbs but these are over a mb. In general, images need to be resized and compressed before upload.

In this case, how about not including images at all since they're the same as the rear-facing camera screenshots?

lognaturel:
@grzesiek2010 want to make sure you saw my comment so we can get this merged! 😊

grzesiek2010:
Sorry for the delay! I'll fix it soon.

grzesiek2010:
> In this case, how about not including images at all since they're the same as the rear-facing camera screenshots?

yeah maybe you are right it's not really needed. Fixed.

PR #991, opened 2019-03-15
Describe how to handle a repeat_count that may be decreased by the user

### LINKED ISSUE #726: Describe how to handle a repeat_count that may be decreased by the user
https://docs.opendatakit.org/form-repeats#using-a-constraint-to-limit-repetitions should include a note or a section pointing out that if the count is decreased by a user, no groups will be deleted. 

A technique to address that is in the "repeat-groups are not respecting the jr:count setting after a change" at https://opendatakit.org/help/form-design/workarounds/.

An XLSForm is linked from https://github.com/opendatakit/docs/issues/181#issuecomment-332163330

grzesiek2010:
closes #726

The pr just documents how to handle a repeat_count that may be decreased by the user.

### DIFF
--- a/odk1-src/form-repeats.rst
+++ b/odk1-src/form-repeats.rst
@@ -38,7 +38,25 @@ and set a constraint on that question.
   text, guest_name, Guest's name, ,
   text, guest_dietary, Does this guest have any dietary restrictions?, , 
   end_repeat,,,,
+
+.. note::
+
+  If the count is decreased by a user, no groups will be deleted. This avoids accidental data loss: a user who accidentally sets the count too low can set it to a higher number and still have the repetitions that were previously created.
+
+A recommended way to handle this case is using relevance to hide any extra values:
   
+.. csv-table:: survey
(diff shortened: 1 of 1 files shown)


odk1-src/form-repeats.rst
```diff
@@ -38,7 +38,25 @@ and set a constraint on that question.
   text, guest_name, Guest's name, ,
   text, guest_dietary, Does this guest have any dietary restrictions?, , 
   end_repeat,,,,
+
+.. note::
+
+  If the count is decreased by a user, no groups will be deleted. It's an approach we agreed on in order to avoid accidental data loss.
```
lognaturel: To be a little more explicit, how about: "It's an approach we agreed on in order to avoid accidental data loss." -> "This avoids accidental data loss: a user who accidentally sets the count too low can set it to a higher number and still have the repetitions that were previously created."

grzesiek2010: Improved.

PR #995, opened 2019-03-27
Clarified which sheet in a spreadsheet is filled

### LINKED ISSUE #994: ODK Collect sends data to the first sheet in a spreadsheet.
If we send data to Spreadsheet ODK Collect tries to send it to the first sheet no matter what is the name. 
It should be documented because some users might have problems:
https://forum.opendatakit.org/t/spreadsheet-is-missing-the-following-columns-bdd-date/18560/11

grzesiek2010:
closes #994 

This pr clarifies which sheet in a spreadsheet is being filled during sending forms.

### DIFF
--- a/odk1-src/collect-forms.rst
+++ b/odk1-src/collect-forms.rst
@@ -169,6 +169,9 @@ If you are connected to :doc:`an ODK Aggregate server  <collect-connect-aggregat
   - Use the `TRIM <https://support.google.com/docs/answer/3094140?hl=en>`_ function in the google sheets to remove the leading and trailing spaces from the cells.
   - Define empty cell in your tests to be a cell that is either empty or contains a single space.
 
+.. note::
+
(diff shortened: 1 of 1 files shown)


odk1-src/collect-forms.rst
```diff
@@ -169,6 +169,9 @@ If you are connected to :doc:`an ODK Aggregate server  <collect-connect-aggregat
   - Use the `TRIM <https://support.google.com/docs/answer/3094140?hl=en>`_ function in the google sheets to remove the leading and trailing spaces from the cells.
   - Define empty cell in your tests to be a cell that is either empty or contains a single space.
 
+.. note::
+
+  Using Google Drive as a server, filled forms are sent to the first sheet in a given spreadsheet, no matter what is its name. If you use one spreadsheet to keep an original form and to collect filled forms make sure the sheet you expect to be filled is in the first place.
```
lognaturel: "what is its name" -> "what its name is"

How about "an original form" -> "a form definition"? I had to think through what "original" would mean in this context.

grzesiek2010: You are right, fixed. Thanks!

PR #1018, opened 2019-04-23
Replace occurrences of ODK 2 with ODK-X

ray-fung:
closes #

addresses #

Replaces the majority of occurrences of ODK 2 with the new ODK-X.

Links with existing ODK-2 in the address have been left purposefully. Might want to change those to ODK-X in the future to avoid confusion. As noted by @yanokwa, all instances of ODK 1 also need to be removed.


yanokwa:
Thanks for sending in this PR, @ray-fung! One quick thing to add to left to be addressed is removing all instances of "ODK 1".

wbrunette:
@yanokwa we trying to split up the work, @ray-fung is only doing part of the work.

PR #1026, opened 2019-04-30
Clarified the difference between .xlsx files and worksheets. Ensures …

ishannla:
…that user doesn't accidentally create multiple .xlsx files instead of worksheets. Build tested.

closes #

addresses #

A change to the documentation that clarifies that the worksheets are to be created within the .xlsx file. 

None

None

None

### DIFF
--- a/odk2-src/conf.py
+++ b/odk2-src/conf.py
@@ -282,7 +282,7 @@ Download this documentation as a PDF.
 """
 odk_pdf = """
 
-_downloads/ODK2-Documentation.pdf
+_downloads/ODKX-Documentation.pdf
 
 """
 prob_in_doc = """
--- a/odk2-src/index.rst
+++ b/odk2-src/index.rst
@@ -2,7 +2,7 @@
 
   geotagging
 
-Welcome to Open Data Kit 2's documentation!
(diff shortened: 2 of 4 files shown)


odk2-src/index.rst
```diff
@@ -2,7 +2,7 @@
 
   geotagging
 
-Welcome to Open Data Kit 2's documentation!
+Welcome to Open Data Kit X's documentation!
```
wbrunette: Should be "Open Data Kit Tool Suite X 's"

odk2-src/conf.py
```diff
@@ -282,7 +282,7 @@ def setup(app):
 """
 odk_pdf = """
 
-_downloads/ODK2-Documentation.pdf
+_downloads/ODKX-Documentation.pdf
```
wbrunette: First it should be "ODK-X".

Second, for the generated file to have that name you will need to change the build file so when it copies it into place it had the correct name.
See:
	mv OpenDataKit2.pdf ../_downloads/ODK2-Documentation.pdf

https://github.com/opendatakit/docs/blob/f77ead63c5272c14ebb767479ffdc1c2de0d8842/Makefile#L90

PR #1029, opened 2019-05-12
Added hint for using Git on windows

e2themillions:
Hello,

Sorry if I'm not following procedure - I am new to this, so please bear with me :)
I added a warning about Git for windows substituting LF with CRLF, since that will cause the containers to be unable to start. Perhaps it could be useful to someone else - it did at least cost me quite a lot of time to figure out that this was the issue ;-)

(on a side note; it is mentioned that sync-endpoints should be compiled with java jdk 8, but I could only get it working with jdk 12? If I used jdk 8, it crashed with a lot of "duplicate class definition" errors - this could however also be a side effect of my messing up of line endings with autocrlf=true in git)

A note in the documentaion

### DIFF
--- a/odk2-src/sync-endpoint.rst
+++ b/odk2-src/sync-endpoint.rst
@@ -44,6 +44,8 @@ ODK Sync Endpoint requires a database and a *LDAP* directory, you could follow t
 
   All of the following commands should be run on your server.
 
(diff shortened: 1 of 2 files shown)


wbrunette:
@e2themillions thank you for helping to improve ODK-X's documentation!!!  

You are following correct procedure. Please fix your build error and we will approve the request. (Click on the details link on the failed CircleCI build to see what is wrong).

e2themillions:
@wbrunette thanks a lot. Sorry for committing a couple of times, I was a little confused ;-) But I believe the build errors are now fixed :)

PR #1034, opened 2019-05-22
Warn about ips

### LINKED ISSUE #1030: Correction in Aggregate AWS installation docs
Hi,

On this page: https://docs.opendatakit.org/aggregate-aws/, there is a slight correction needed. 

In section CREATE AN EC2 MACHINE, line 8 says: 
>Select Enable in the Auto-assign Public IP dropdown.

This is functionally correct step. However, there is a problem associated with it. By assigning IP like this, AWS assigns you a temporary IP which vanishes if you reboot the machine (a new temp IP is assigned). So this will cause the problem. 

Solution: Instead of assigning IP here in EC2 creation dialogue, you should create the machine without a public IP. Once the machine creation is completed, from the left menu, go to NETWORK & SECURITY > ELASTIC IP, and reserve/allocate an IP there. This IP reservation is permanent for your account (unless you release it). Once allocated, you can then click on the IP and click ASSOCIATE ADDRESS to attach it to your machine (which will be listed in the dropdown). This way, the IP will not vanish on reboot.

yanokwa:
Closes #1030.

I would love a review that confirms that:
1. Making this a tip not a warning is the right thing to do.
1. The placement in the docs is correct.
1. The language I use is consistent and the links are to useful places.

Here's what it looks like...
Screen Shot 2019-05-22 at 12 29 52 PM

### DIFF
--- a/odk1-src/aggregate-aws.rst
+++ b/odk1-src/aggregate-aws.rst
@@ -132,6 +132,8 @@ Create an EC2 machine
 Set up your domain
 ------------------
 
+.. tip:: EC2 machines use IP addresses which can change if you stop and start (but not reboot) the machine. To ensure your Aggregate install will always be reachable using the same IP address, use an Elastic IP by following `these instructions <https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/elastic-ip-addresses-eip.html>`_.
+
 1. Go to the `EC2 - Instances <https://console.aws.amazon.com/ec2/v2/home#Instances:>`_ tab and find your machine.
 
 2. Take note of the IPv4 Public IP address (e.g., 12.34.56.78) and set a *DNS A record* pointing to it.
--- a/odk1-src/aggregate-azure.rst
+++ b/odk1-src/aggregate-azure.rst
@@ -62,6 +62,8 @@ Create a VPC
 Set up your domain
 ------------------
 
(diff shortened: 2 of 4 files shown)


saadaan:
Hi Yaw et all,

While you are right that AWS IP changes only when the instance is stopped and restarted (and not on reboot), please note that there are a lot of settings for which it is mandatory to stop the instance, apply the change, and then restart the instance (meaning REBOOT will not work). An example I can give out of my experience is when you have to change the instance type/size, lets say from micro to small (or any other). You cannot do it with a reboot. Effectively, you would have to stop the instance, make the change via console, and then restart the instance. There are other settings like this as well. If the public IP is assigned on the fly, you will lose it this way. It is much safer to assign the IP by allocating it from within AWS account separately, and then attach it with the instance, so that we don't lose it.

Cheers,
Saad

yanokwa:
@saadaan I agree that resizing the instance will require an IP change, but I don't expect that to be a common operation. My goal in the documentation is to have a consistent process across the various cloud providers (of which AWS is one) that captures the common case and provides hints to users who don't have the common case.

@ggalmazor Do you agree?

saadaan:
@yanokwa Very much agreed. I am just concerned that this situation of IP switch (if it happens) would be categorized/flagged as a possible problem with the aggregate server, when actually it actually is not an aggregate issue. But your idea of keeping a consistent guide is also valid. Maybe you could also update it in docs if such issues appear in the forum frequently.

The alternate process (adding IP separately), however, is just a couple of additional steps, not more.

PR #1040, opened 2019-05-28
Document answer tracking in audit file

### LINKED ISSUE #1031: Update Logging enumerator behavior section
We are going to add an option to log new answers in audit file: https://github.com/opendatakit/collect/pull/3072 That will require [Logging enumerator behavior section](https://docs.opendatakit.org/form-audit-log/?highlight=audit#logging-enumerator-behavior) to be updated.

grzesiek2010:
closes #1031 

It just documents answer tracking in the audit file: https://github.com/opendatakit/collect/pull/3072

### DIFF
--- a/odk1-src/form-audit-log.rst
+++ b/odk1-src/form-audit-log.rst
@@ -78,6 +78,17 @@ For the most accurate locations, set :tc:`location-priority` to `high-accuracy`.
 
 .. _viewing-audit-logs:
 
+Answer tracking
+~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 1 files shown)


lognaturel:
In my eagerness I forgot that both `pyxform` and Collect are unreleased. 😢

I think it's ok, though. The Collect release will happen very soon. The `pyxform` release is not scheduled but if users do try to follow the instructions and get confused, they should come to the forum and that will give us a reason to do the `pyxform` release as soon as possible.

Issue #710, opened 2018-05-17
Display only newest form versions in Collect


grzesiek2010:
On the Collect's side, we are going to hide old form versions. I mean if a user has a few forms with the same `formId` and `version number` only the newest one should be displayed. By `the newest one`, I mean
the form with the highest `version number` or the newest one if all `version numbers` are the same.
More info: https://github.com/opendatakit/collect/issues/1694

lognaturel:
https://docs.opendatakit.org/form-update/ is where the update needs to go.

PR #1044, opened 2019-05-30
Only newest form versions is visible in Collect

grzesiek2010:
addresses #710 

The pr documents the change we implemented in collect:
-only the newest form version is displayed on the form list.
-updates have an additional message on the list of forms we can download


lognaturel:
Thanks, @grzesiek2010, and sorry I missed this!

(Hah, looks like buttons aren't debounced and I accidentally double clicked the review button)

PR #1057, opened 2019-06-24
Release tweaks

yanokwa:
* Slightly more robust language about what Aggregate displays
* Match formatting of Briefcase CLI docs
* Document that we can validate multiple forms

### DIFF
--- a/odk1-src/validate.rst
+++ b/odk1-src/validate.rst
@@ -45,9 +45,6 @@ Setting up Validate
 Using Validate
 ---------------
 
-GUI
-~~~
-
 #. Open :program:`Validate`.
 #. Find your XForms :file:`*.xml` file using :guilabel:`Choose file`, 
    and :guilabel:`Open` it.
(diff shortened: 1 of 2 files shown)


odk1-src/validate.rst
```diff
@@ -59,14 +56,15 @@ GUI
...
 ~~~~~~~~~~~~
 
-``java -jar ODKValidate.jar [--fail-fast] path/to/xform.xml [FORM...]``. The
-``--failFast/--fail-fast`` command line argument tells the validator to abort
-on the first error rather than validating all files and reporting an error at
-the end.
-  
+.. code-block:: console
+
+  $ java -jar {path/to/validate-jar-file} [--fail-fast] {path/to/xform1.xml} {path/to/xform2.xml}
```
lognaturel: How about making it clearer that an arbitrary number of forms can be specified? I think there should be some text to describe that.

yanokwa: Change made and force pushed. If you are happy with the changes, please review and merge!

PR #1068, opened 2019-07-05
Indent Briefcase details for better presentation

yanokwa:
Minor change whiles I was confirming if indentation was working as expected.

Screen Shot 2019-07-05 at 2 20 12 PM

Screen Shot 2019-07-05 at 2 20 33 PM


lognaturel:
This looked like it was a no-brainer but now I'm confused. It looks like some of the changes were already made in master? It would be good to make sure that my conflict resolution does what was intended. 

Does it really make sense to hide one-line details? It takes up less space to just put it inline.

PR #1083, opened 2019-08-08
Adds a Dockerfile for the common portions of the ODK1 and ODK2 builds

programmablemike:
Adds a Dockerfile for the common portions of the ODK1 and ODK2 builds

https://github.com/opendatakit/docs/issues/1084

Testing on the build machine and fixing any issues.


lognaturel:
Hi @programmablemike! Would love to learn more about the status here -- this has the potential to be really valuable and I want to make sure we don't lose momentum. Do you think you might still have time/interest to devote to it? If not, could you please write out a few notes for the next person to look into it?

lognaturel:
We have now split up the two tool suites so this repo is only for ODK. Thanks again for looking into this, @programmablemike!

PR #1086, opened 2019-08-08
Use JDK 11 because JDK 8 is no longer supported

### LINKED ISSUE #1085: Deploy step is broken on CI because openjdk-8-jre-headless is no longer available
~Should be openjdk-11-jre-headless~

Unfortunately s3_website needs jdk8 so we'll need to do something like https://github.com/opendatakit/website/commit/f6c41afe61da627e86c3f2143ef40408e875a836. I'll give @programmablemike a moment to look further into #1084 because then we'll be able to make this change in one place.

lognaturel:
closes #1085

I did a search and replace for `openjdk-8-jre-headless` to `openjdk-11-jre-headless`. I don't think there's a great way of verifying this other than merging and having it run on CircleCI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -151,7 +151,7 @@ jobs:
           name: Install deploy requirements
           command: |
               if [[ "$CIRCLE_PROJECT_USERNAME" == "opendatakit" ]]; then \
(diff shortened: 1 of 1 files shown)


lognaturel:
I cancelled the CI run because the deploy step isn't run on branches.

PR #1092, opened 2019-08-13
Clarify ex:prefix use

atl2017:
Add text "Notice that the ex:prefix is not included" to Clarify that ex: prefix is only for external apps that populate one field.

closes #

addresses #


lognaturel:
Thanks so much, @atl2017! 🎉

PR #1095, opened 2019-08-14
Raw values are sent from Collect to Google Sheets as of v1.23

### LINKED ISSUE #1077: Raw values are sent from Collect to Google Sheets as of v1.23
Prior to Collect v1.23, the behavior for values sent to Google Sheets was: "The values will be parsed as if the user typed them into the UI. Numbers will stay as numbers, but strings may be converted to numbers, dates, etc. following the same rules that are applied when entering text into a cell via the Google Sheets UI."

Now, they are sent raw. The implication there is that a tick before each cell's value (see https://github.com/opendatakit/collect/pull/3128#issuecomment-498589849 for details). Users may want to format columns before doing analysis. The format will not be applied to new submissions.

grzesiek2010:
closes #1077 

It just documents that not blank values will be preceded by an apostrophe to avoid parsing them.

### DIFF
--- a/odk1-src/collect-forms.rst
+++ b/odk1-src/collect-forms.rst
@@ -169,6 +169,8 @@ If you are connected to :doc:`an ODK Aggregate server  <collect-connect-aggregat
   - Use the `TRIM <https://support.google.com/docs/answer/3094140?hl=en>`_ function in the google sheets to remove the leading and trailing spaces from the cells.
   - Define empty cell in your tests to be a cell that is either empty or contains a single space.
 
+  Other values will be preceded by an apostrophe. Thanks to that they won't be parsed automatically in google sheets (e.g. strings to dates) and will be stored as-is.
+
 .. note::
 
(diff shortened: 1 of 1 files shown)


odk1-src/collect-forms.rst
```diff
@@ -169,6 +169,8 @@ If you are connected to :doc:`an ODK Aggregate server  <collect-connect-aggregat
   - Use the `TRIM <https://support.google.com/docs/answer/3094140?hl=en>`_ function in the google sheets to remove the leading and trailing spaces from the cells.
   - Define empty cell in your tests to be a cell that is either empty or contains a single space.
 
+  Other values will be preceded by an apostrophe. Thanks to that they won't be parsed automatically in google sheets (e.g. strings to dates) and will be stored as-is.
```
lognaturel: I think this doesn't make it clear enough why a user should care and what they might do about it. How about "Other values will be preceded by an apostrophe. This prevents Google Sheets from guessing at the data type and applying a format that may not be appropriate (e.g. making ``1940 10 5`` into a date when it actually is the value of a select multiple). Raw values may be used in any kind of computation and the apostrophes only are visible when editing a value. You may choose to manually apply a format for certain rows if desired." If you can think of a way to make it shorter but include all the info, that would be even better.

grzesiek2010: Ok seems fine. Done.

PR #1101, opened 2019-09-02
v0.6 supports encryption and push/pull for Briefcase

yanokwa:
Addresses #1076 

We now support encrypted forms and ODK Briefcase v1.17 (which is ready, but hasn't been released) has push/pull support.


issa-tseng:
oops. thanks!

PR #1113, opened 2019-10-03
Add documentation for media and autoplay

### LINKED ISSUE #1112: Document audio, video media and autoplay
There are some examples of using audio and video media in forms as part of "Form Styling" and also with specific examples for "Select widgets" but there isn't any thorough documentation on using audio/video in forms. Things I think we should cover are:

* How to put audio/video in at the XLSForm stage
* How audio/video gets presented in Collect
* How to have audio and video autplayed (and how it behaves in Collect)

seadowg:
Closes #1112

Additional documentation (and some reworking) around using media (image, audio, video) in forms for Collect. 

I tried to think about the journeys for learning about adding media and was happy with "Form Styling" providing most of the info and also with having documentation around adding files themselves to a media folder living elsewhere. I've kept everything linked so the journeys through the docs work like this:

* Media -> Loading Form Media
* Loading Form Media -> Media
* Including Media in choices -> Media -> Loading Form Media

### DIFF
--- a/odk1-src/form-styling.rst
+++ b/odk1-src/form-styling.rst
@@ -32,7 +32,7 @@ Files referenced should be included :ref:`in your form's media folder <loading-f
 Images
 ~~~~~~~~
 
-Adding image to a question adds a displays the image as part of the question.
+Adding an image to a question displays the image as part of the question.
 
 .. image:: /img/form-styling/media-image.*
   :alt: A single select widget in Collect. The label text is "Do you want coffee?" The label text is accompanied by a picture of a mug of coffee. The options are "yes", "no", and "I don't know".
--- a/odk1-src/form-styling.rst
+++ b/odk1-src/form-styling.rst
@@ -79,10 +79,8 @@ Adding video to a question adds a button that will play the video clip full scre
 
   integer, people, How many people do you see in the video?, people.mp4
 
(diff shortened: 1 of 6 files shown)


seadowg:
Hmmmm loooks there the build is not a fan of the word "autoplay". Give we the forms use `autoplay` as a column though I think it would be good to use it in that form. Is there a way to add exceptions to the spell check?

odk1-src/form-styling.rst
```diff
@@ -27,23 +27,57 @@ Media
 
 You can include questions in your form that display images or that play video or audio files by including a :th:`media` column in your `XLSForm <http://xlsform.org/#media>`_.
 
-.. image:: /img/form-styling/media-image.* 
+Images
+~~~~~~~~
+
+Adding image to a question adds a displays the image as part of the question.
```
lognaturel: "Adding image to a question adds a displays" -> "Adding image to a question displays"

seadowg: Good catch!

seadowg:
@lognaturel ah yeah I can expand the docs for the GridWidget audio. It's a bit of a special case but I'll have a think about where it fits in.

seadowg:
Added a note about the audio features in `GridWidget`. Let me know what you think!

Issue #1016, opened 2019-04-20
Add section on field-list updates and remove existing warning


lognaturel:
Once collect#3023 is merged, https://docs.opendatakit.org/form-question-types/#grouping-multiple-widgets-on-the-same-screen will need to be modified. A section on field list updates and its limitations should also be added.

lognaturel:
Free-entry fields update as each character change is made. This means tricks with `once()` don't work. See https://forum.opendatakit.org/t/calculation-error-with-new-version-since-1-24-0/20562/3?u=ln

PR #1126, opened 2019-10-17
Logic within a field list is evaluated in Collect v1.22+

lognaturel:
Begins addressing #1016

How to frame the current limitations with `once` requires some thought but I want to make sure that it's clear that newer versions of Collect do support logic evaluation within a screen.

### DIFF
--- a/odk1-src/form-question-types.rst
+++ b/odk1-src/form-question-types.rst
@@ -2249,16 +2249,7 @@ The :tc:`field-list` appearance attribute, applied to a group of widgets, displa
 
 .. warning::
 
-  Do not include a question that depends on a previous answer
-  in the same field list as the previous question it depends on.
-
-  Relevance and calculation values are determined
-  when the screen advances,
-  which does not happen between questions
-  that are grouped on the same page.
-  Therefore,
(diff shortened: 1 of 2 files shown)


odk1-src/form-question-types.rst
```diff
@@ -2249,16 +2249,7 @@ The :tc:`field-list` appearance attribute, applied to a group of widgets, displa
...
-  in the same field list as the previous question it depends on.
-
-  Relevance and calculation values are determined
-  when the screen advances,
-  which does not happen between questions
-  that are grouped on the same page.
-  Therefore,
-  including a question and its dependent together in a field list
-  will not work as expected.
+  Relevance, constraint and calculation evaluation within the same screen is supported in Collect v1.22 and later.
```
seadowg: Feels to me it would be good to also have this in the relevant form logic sections. I'm worried about someone learning about form logic and then not realizing it won't work in certain versions. What do you think?

lognaturel:
Good point, @seadowg. I added a warning to the top of https://docs.opendatakit.org/form-logic/.

Issue #1133, opened 2019-11-06
add more details for Central install on AWS


danbjoseph:
would be helpful to have some more details about the options for installing elsewhere

PR #1134, opened 2019-11-06
installing Central on AWS

danbjoseph:
addresses #1133

I've added some details that helped me get Central running on AWS EC2.

I feel like it's slightly too much text to be a subsection on the page I placed it. But would need to be filled out more (and end up duplicating a lot of the DigitalOcean instructions) if given its own page such as `central-install-aws.rst`. Where should the added information be placed?

I'm also not entirely sure I haven't ignored any best practices when setting up something in production on an AWS EC2. Does anyone know of anything that works but wasn't a recommended thing to do?


issa-tseng:
i think this is a great start!

eventually, it would be nice to inline some of the steps that are currently linked out, just in case those pages disappear or change, and perhaps to be a little more granular and specific with the Elastic IPs step. but i see no reason not to merge this as-is for now.

thank you!!

Issue #1135, opened 2019-11-08
Likert documentation


shouryaj:
Documentation for issue #1272 in ODK Collect 

https://github.com/opendatakit/collect/issues/1272

lognaturel:
Ah, I see why @grzesiek2010 ended up with a duplicate. @shouryaj, please make sure to claim an issue you're working on with `@opendatakit-bot claim` and put `Closes #<issue number` at the top of a PR you submit so it's clear someone is working on it!

shouryaj:
Oh oops! My bad will be more careful in the future!

PR #1136, opened 2019-11-08
Adding likert widget documentation

shouryaj:
addresses #1135 

Added likert widget documentation

### DIFF
--- a/odk1-src/form-question-types.rst
+++ b/odk1-src/form-question-types.rst
@@ -805,6 +805,43 @@ See :ref:`image-options` to learn more about including images in surveys.
   abcd_icon,d,D,d.jpg
 
 
+.. _likert-image-widget:
+
+Select widget with likert
+""""""""""""""""""""""""""""""""""
+type
+ :tc:`select_one {list_name}`
+appearance
+ :tc:`likert`
+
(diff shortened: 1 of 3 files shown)


odk1-src/form-question-types.rst
```diff
@@ -805,6 +805,43 @@ See :ref:`image-options` to learn more about including images in surveys.
...
+Select widget with likert
+""""""""""""""""""""""""""""""""""
+type
+ :tc:`select_one {list_name}`
+appearance
+ :tc:`likert`
+
+Select widget supports likert options. Options can include text, images or both. If both are provided, images appear above text.
+
+Take into consideration that likert options DO NOT wrap to next line.
```
lognaturel: It would be helpful to clarify the implications of this. Maybe something like "Take into consideration that likert options DO NOT wrap to next line and will overlap instead. If you have long option text, make sure to try the question on the device you will use for data collection."

lognaturel: It doesn't look like the update for this got committed, @shouryaj!

shouryaj: My bad! Not sure what happened there. Just pushed this change up again.

PR #1138, opened 2019-11-08
Clean up repeats section and add recipes for avoiding dialog

### LINKED ISSUE #1042: Add repeat recipes for avoiding "add new repeat" dialog
See examples and discussion at https://forum.opendatakit.org/t/xls-form-repeat-select-one-question-until-a-is-selected/15318

lognaturel:
closes #1042

- an introduction to repeats
- a rework of the repeat count tip which wasn't right
- links between the repeat tips and repeat introduction sections
- an example for indefinite repeats (#1042)
- an example for 0 or more repeats

I'd like to keep adding recipes but I think this adds a lot of value already.

We're doing this now in the context of @seadowg taking a look at the repeat user experience in Collect.

### DIFF
--- a/odk1-src/form-logic.rst
+++ b/odk1-src/form-logic.rst
@@ -536,7 +536,7 @@ You can ask the same question or questions multiple times by wrapping them in :t
 .. csv-table:: survey
   :header: type, name, label 
 
-  begin_repeat, my_repeat, Repeat label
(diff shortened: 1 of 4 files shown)


lognaturel:
Huh. I double clicked the PR creation button, I guess?

PR #1156, opened 2019-12-09
Add identify user/track changes reasons docs

### LINKED ISSUE #1155: Add documentation for Collect 1.25 audit features
Collect 1.25 contains two new audit features that need documented:

- [x] [Identify User](https://github.com/opendatakit/collect/pull/3439)
- [ ] [Track Changes Reasons](https://github.com/opendatakit/collect/pull/3477)

seadowg:
Closes #1155.

A new section in the form audit docs around `identify-user`. Was originally thinking of including screenshots but not sure we need it for this. Interested in if others disagree!

### DIFF
--- a/odk1-src/form-audit-log.rst
+++ b/odk1-src/form-audit-log.rst
@@ -93,6 +93,10 @@ You can enable change tracking so that old answers and new answers will be added
 Enumerator identification
 ~~~~~~~~~~~~~~~~~~~~~~~~~
 
+.. versionadded:: 1.25
+
+  `ODK Collect v1.25.0 <https://github.com/opendatakit/collect/releases/tag/v1.25.0>`_
+
 If your form needs a record of the identity of the enumerator you can use :tc:`identify-user=true`.
 
 .. csv-table:: survey
--- a/odk1-src/form-audit-log.rst
+++ b/odk1-src/form-audit-log.rst
@@ -97,12 +97,12 @@ Tracking reason for edit
 
   `ODK Collect v1.25.0 <https://github.com/opendatakit/collect/releases/tag/v1.25.0>`_
 
-You can add to :tc:`track-changes-reason=on-form-edit` to prompt enumerators to enter a reason before they save changes to a form:
+You can add to :tc:`track-changes-reasons=on-form-edit` to prompt enumerators to enter a reason before they save changes to a form:
 
 .. csv-table:: survey
   :header: type, name, parameters
 
-  audit, audit, track-changes-reason=on-form-edit
+  audit, audit, track-changes-reasons=on-form-edit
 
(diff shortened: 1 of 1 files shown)


odk1-src/form-audit-log.rst
```diff
@@ -90,6 +90,17 @@ You can enable change tracking so that old answers and new answers will be added
 
   audit, audit, track-changes=true
 
+Enumerator identification
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If your form needs a record of the identity of the enumerator you can use :tc:`identify-user=true`.
```
lognaturel: I think it would be great to add some context as to how this might be useful and to contrast it with `username` in https://docs.opendatakit.org/collect-settings#form-metadata-settings and accessed by https://docs.opendatakit.org/form-question-types/#metadata.

Maybe something like this:

In some data collection workflows, devices are passed between different enumerators for data verification or completion. If you need to know who had the device when data was entered or changed, you can use :tc:`identify-user=true`.

If each device will be used by a single enumerator, you can use the server username or [settings username](https://docs.opendatakit.org/collect-settings#form-metadata-settings) and the [`username` metadata question](https://docs.opendatakit.org/form-question-types/#metadata). This will write the username to each submission instead of to the audit log.

seadowg: Will have a look at this

seadowg: Chose to do this in a "tip". Felt to me like it was nice to have the core text be fairly functional and then have the tip provide information around the other possible feature.

seadowg:
@lognaturel I've added a note about the Collect version. To me, it makes sense to keep the documentation we have rather than the `bind` syntax and just hold this until https://opendatakit.org/xlsform/ is up to date? Does that sound good?

lognaturel:
Sounds great to me! Let's work to get that `pyxform` release out asap. Lots of good stuff in there.

seadowg:
@lognaturel seeing as we're holding this I'm acutally going to do the "Track Changes Reasons" docs in this PR as well so we don't have another PR to keep track of. Hope that makes sense to you!

odk1-src/form-audit-log.rst
```diff
@@ -90,6 +90,22 @@ You can enable change tracking so that old answers and new answers will be added
...
   audit, audit, track-changes=true
 
+Tracking reason for edit
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 1.25
+
+  `ODK Collect v1.25.0 <https://github.com/opendatakit/collect/releases/tag/v1.25.0>`_
+
+You can add to :tc:`track-changes-reason=on-form-edit` to prompt enumerators to enter a reason when they save changes to a form:
```
lognaturel: Should be `track-changes-reasons` with an `s` at the end!

seadowg: Ach! Every time.

PR #1157, opened 2019-12-10
Add form map documentation

### LINKED ISSUE #1144: Document the new map view of saved forms (MapActivity)
This functionality is added in https://github.com/opendatakit/collect/pull/3447.

lognaturel:
closes #1144

I didn't add any screenshots because they add a maintenance burden, make it harder to scan and I think that in this case text is sufficient to describe what's going on. If reviewers think it would be helpful to have some visuals, let me know.

I tried to be really thorough about documenting all states that forms could be in and what the behavior should be. Gaps there would be something to look for.

I also tried to link out to any related documentation.

FYI @zestyping if you're interested in taking a peek.

### DIFF
--- a/odk1-src/collect-form-map.rst
+++ b/odk1-src/collect-form-map.rst
@@ -28,7 +28,8 @@ Tapping on a map marker will open the filled form for viewing or editing dependi
 
 The bar at the bottom of the screen displays the total number of saved forms and how many of these are shown on the map. All filled forms with a value for the first geopoint question will be displayed. To ensure that all filled forms are displayed, make the identifying geopoint question :ref:`required <requiring-responses>`.
 
-Deleted or :doc:`encrypted <encrypted-forms>` filled forms are not shown on the map. However, forms that were successfully sent and then deleted and forms that are encrypted both contribute to the total number of saved forms. See :ref:`deleting-forms` for more on how filled form deletion works. 
+.. note::
+  Deleted or :doc:`encrypted <encrypted-forms>` filled forms are not shown on the map. However, forms that were successfully sent and then deleted and forms that are encrypted both contribute to the total number of saved forms. See :ref:`deleting-forms` for more on how filled form deletion works.
 
 There are three control buttons clustered at the top right of the map. The top button is used to zoom to the current location. The middle button adjusts the zoom level to ensure that all mapped filled forms are displayed on the screen. The last button is used to change layers if :doc:`offline layers <collect-offline-maps>` are available. The basemap and reference layer settings are used across all of Collect so the same ones will be used for this form map as for :ref:`location widgets with maps <location-widgets>`.
 
--- a/odk1-src/collect-form-map.rst
+++ b/odk1-src/collect-form-map.rst
@@ -17,6 +17,11 @@ If you are collecting data related to physical locations, you can see a map of a
 
 In the :guilabel:`Fill Blank Form` list, form definitions with a geopoint question outside of any repeat have a map button. Tap on the map button to open a map showing the filled forms for that form definition. The first geopoint question in the form definition that is outside of any repeat is used to map filled forms. The first geopoint question can be one that either :ref:`requires user intervention <location-widgets>` or :ref:`gets the location in the background <metadata-start-geopoint>`. On the map, you will also see your current position and be able to fill a new instance of the current form definition.
 
+.. form-instance-map-markers:
+
+Form instance map markers
+----------------------------
+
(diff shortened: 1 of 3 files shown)


odk1-src/collect-form-map.rst
```diff
@@ -0,0 +1,36 @@
...
+Viewing filled forms on a map
+================================
+
+.. versionadded:: 1.25
+
+  `ODK Collect v1.25.0 <https://github.com/opendatakit/collect/releases/tag/v1.25.0>`_
+
+If you are collecting data related to physical locations, you can see a map of all filled forms from the :doc:`Fill Blank Form <collect-filling-forms>` menu. Use the map view to see your progress, to identify filled forms with bad locations and to plan your travel to the next location to survey.
+
+.. note::
```
seadowg: Maybe I'm being too lasse fair about this but it feels to me like we don't need this level of detail around it only working in 1.25. I'd imagine at least one person will see this problem and ask about it in the forum but it feels like the sort of thing that will get asked whether it's in the docs or not. What do you think?

odk1-src/collect-form-map.rst
```diff
@@ -0,0 +1,36 @@
...
+
+Tapping on a map marker will open the filled form for viewing or editing depending on the form status. By default, forms with :formstate:`Saved`, :formstate:`Finalized`, and :formstate:`Submission Failure` status are opened for edit and :formstate:`Sent` forms are opened for viewing. However, if the :guilabel:`Edit Saved Form` button is made unavailable from the :ref:`admin settings <user-access-control-settings>`, forms with :formstate:`Saved`, :formstate:`Finalized`, and :formstate:`Submission Failure` status are opened as view-only.
+
+The bar at the bottom of the screen displays the total number of saved forms and how many of these are shown on the map. All filled forms with a value for the first geopoint question will be displayed. To ensure that all filled forms are displayed, make the identifying geopoint question :ref:`required <requiring-responses>`.
+
+Deleted or :doc:`encrypted <encrypted-forms>` filled forms are not shown on the map. However, forms that were successfully sent and then deleted and forms that are encrypted both contribute to the total number of saved forms. See :ref:`deleting-forms` for more on how filled form deletion works.
```
seadowg: I think this could be a `note`/`warning`.

odk1-src/collect-form-map.rst
```diff
@@ -0,0 +1,36 @@
+.. spelling::
+
+  unfinalized
+  basemap
+
+Viewing filled forms on a map
+================================
+
+.. versionadded:: 1.25
```
seadowg: Ah this is a really nice way of showing where a feature is added. Totally stealing this.

lognaturel:
> One thing I'm thinking is that there is a lot of text describing the UI. Do you think we might be able to reduce using a screenshot/gif? 

I think the text would still be needed to describe HOW the different UI elements behave. Just a screenshot doesn't really help because it's what users see when they get to that screen anyway. I think screenshots are important when there's ambiguity around which screen is being described or which button does what. Does that make sense? In this case I think it's pretty clear that there's only one relevant screen.

odk1-src/collect-form-map.rst
lognaturel: I think this behavior needs to be documented somewhere as being intentional and because it's not necessarily totally intuitive what the fix is. That is, is it downloading the form with a particular version? Pressing a button somewhere? ...? I agree that it's not critical to have it in user docs but I can't really think of a better place for it to go. Because we have such a long tail of people staying on old versions, it's bound to come up years from now and I'm not sure I'd be able to remember the details!

odk1-src/collect-form-map.rst
lognaturel: Agreed! Pushed in a new commit.

lognaturel:
@seadowg Thanks for the review! Couple of responses above. 🙏

odk1-src/collect-form-map.rst
seadowg: Fair enough. I'm convinced! I think a note is the right formatting a well if it's going in.

seadowg:
@lognaturel yeah I think you're right about the screenshot. What do you think of trying to break the text into headed sub sections around the different features on screen (like "Map Markers")? It just feels a little hard to scan right now.

lognaturel:
> What do you think of trying to break the text into headed sub sections around the different features on screen (like "Map Markers")

Ahh, brilliant! See what you think of the 3 sections I broke it into.

PR #1166, opened 2020-01-09
tip for getting Central version

danbjoseph:
apologies if this is already noted somewhere.
also, will gladly close and resubmit this pull request if there's a better place in the docs for this.
i asked in Slack about getting the version of an installed Central and @yanokwa helpfully told me about this.


issa-tseng:
:) i think i missed this—thank you!

PR #1173, opened 2020-02-07
Dynamic defaults and ${last-saved#question_name}

lognaturel:
Adds documentation for dynamic defaults and last-saved.

I decided to put the version requirements in warning blocks in this case because using dynamic defaults or last-saved with tools that don't support it will have unpredictable results.

I'd like to add a section about dynamic defaults in repeats but so far the examples I've come up with either require writing XPath expressions like `if(position(..) >= 1, /data/child_details[position() = position(current()/..) - 1]/child_school, '')` or suffer from the issue I described at https://github.com/XLSForm/pyxform/issues/182#issuecomment-583560793. There's also a bug I discovered at https://github.com/XLSForm/pyxform/issues/427. So I think we should come back to that.

### DIFF
--- a/odk1-src/form-logic.rst
+++ b/odk1-src/form-logic.rst
@@ -240,44 +240,69 @@ without answering the question.
 Setting default responses
 ===========================
 
-To provide a default response to a question,
-put the response value in the :th:`default` column.
+To provide a default response to a question, put a value in the :th:`default` column. Defaults are set once when each instance of a form definition is first opened for filling and can then be changed by enumerators. Defaults can either be fixed values (:ref:`static defaults <static-defaults>`) or the result of some expression (:ref:`dynamic defaults <dynamic-defaults>`).
 
-Default values must be static values,
-not expressions or variables.
+.. _static-defaults:
 
-.. note::
+Static defaults
+----------------
 
-  The content of the :th:`default` row in a question
-  is taken literally as the default value.
-  Quotes should **not** be used to wrap string values,
-  unless you actually want those quote marks to appear
-  in the default response value.
+The text in the :th:`default` column for a question is taken literally as the default value. Quotes should **not** be used to wrap values, unless you actually want those quote marks to appear in the default response value.
 
-.. rubric:: XLSForm
+In the example below, the "Phone call" option with underlying value ``phone_call`` will be selected when the question is first displayed. The enumerator can either keep that selection or change it.
+
+.. rubric:: XLSForm to select "Phone call" as the default contact method
 
 .. csv-table:: survey
   :header: type, name, label, default
-  
+
   select_one contacts, contact_method, How should we contact you?, phone_call
-  
+
 .. csv-table:: choices
   :header: list_name, name, label
-  
+
   contacts, phone_call, Phone call
   contacts, text_message, Text message
   contacts, email, Email
-  
-.. tip:: 
-  :name: dynamic-defaults
 
-  You may want to use a previously entered value as a default,
-  but the :th:`default` column does not accept dynamic values.
+.. _dynamic-defaults:
+
+Dynamic defaults
+----------------
+
+.. warning::
   
-  To work around this, use the :th:`calculation` column instead,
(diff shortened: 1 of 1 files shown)


odk1-src/form-logic.rst
```diff
@@ -240,44 +264,71 @@ without answering the question.
...
+
+.. rubric:: XLSForm to set the default username as the server username
+
+.. csv-table:: survey
+  :header: type, name, label, default
+
+  username, username
+  text, confirmed_username, What is your username?, ${username}
+
+If enumerators are creating records representing groups of things with the same property, dynamic defaults can be combined with :ref:`last saved <last-saved>`.
```
yanokwa: Drop the groups of things language.

odk1-src/form-logic.rst
```diff
@@ -240,44 +264,71 @@ without answering the question.
...
+
+Dynamic defaults in repeats are evaluated when a new repeat instance is added.
+
 .. tip:: 
-  :name: dynamic-defaults
+  :name: defaults-from-form-data
 
-  You may want to use a previously entered value as a default,
-  but the :th:`default` column does not accept dynamic values.
+  You may want to use a value filled out by the enumerator as a default for another question that the enumerator will later fill in. Dynamic defaults can't be used for this because they are evaluated once when a record is first created which is before an enumerator fills in any data.
```
yanokwa: It'd be nice to say some of this upfront. "Dynamic defaults cannot be used to pre-populate answers to questions that an enumerator has not yet seen. This is because defaults are evaluated once when a record is first created"

odk1-src/form-logic.rst
```diff
@@ -140,6 +140,30 @@ you must first use a :tc:`calculate` row and then a variable.
   | Tip (18%): $${tip_18}
   | Total: $${tip_18_total}",
 
+.. _last-saved:
+
+Values from the last saved record
```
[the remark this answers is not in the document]
lognaturel: Alternatives would be "filled form" or "form instance". I've used record consistently in this PR. Do you have another preference?

odk1-src/form-logic.rst
```diff
@@ -240,44 +264,71 @@ without answering the question.
 Setting default responses
 ===========================
 
-To provide a default response to a question,
-put the response value in the :th:`default` column.
+To provide a default response to a question, put a value in the :th:`default` column. Defaults are set once when a record is first created from a form definition. Defaults can either be fixed values (:ref:`static defaults <static-defaults>`) or the result of some expression (:ref:`dynamic defaults <dynamic-defaults>`).
```
yanokwa: `Defaults are set when a record is first created from a form definition. Defaults are set once and only once. Defaults can be...`

odk1-src/form-logic.rst
```diff
@@ -240,44 +264,71 @@ without answering the question.
...
+.. _dynamic-defaults:
+
+Dynamic defaults
+----------------
+
+.. warning::
   
+  Support for :ref:`dynamic defaults <dynamic-defaults>` was added in Collect v1.24.0. Form conversion requires XLSForm Online ≥ v2.0.0 or pyxform ≥ v1.0.0. Using older versions will have unpredictable results.
+
+If you put an expression in the :th:`default` column for a question, that expression will be evaluated once when a record is first created from a blank form definition. This allows you to use values from outside the form like the current date or the :ref:`server username <metadata>`. Dynamic defaults can't be used to base the default value of one field on the value of another field in the form. For example, if you have an ``age`` question and you want to default the diagnosis age to the value of the ``age`` question, see :ref:`the tip below <defaults-from-form-data>`.
```
yanokwa: I still feel like this "once" is hard to understand. It'd be good to break it out. 

s/"used to base"/"used to set"

yanokwa: Drop the age question example. You say you can't do something, then you provide an example of what you can do. 

```
field in the form. If you wish to do that, see :ref:`the tip below <defaults-from-form-data>`.

PR #1177, opened 2020-02-18
Clarify that last saved references can be used anywhere

lognaturel:
Augment last-saved documentation based on feedback at https://forum.opendatakit.org/t/some-feedback-on-the-last-saved-feature/24818

@danbjoseph it would be great to get your review! Let me know if you see any other gaps.


yanokwa:
I think `For the first record ever saved for a form definition, the last saved value for any field will be blank` is a little hard to parse, but I don't have a good alternative.

PR #1192, opened 2020-03-11
Update settings documentation for Collect v1.26.0

### LINKED ISSUE #1164: Include material design icon font
It would be helpful to be able to put icons inline to refer to UI elements. It's easier to maintain than screenshots and helps the user quickly find what they want more information about.

https://google.github.io/material-design-icons/#icon-font-for-the-web describes how to use Google's material design font.

https://stackoverflow.com/questions/47643795/how-to-use-custom-woff2-and-oft-fonts-with-sphinx has some ideas on how to do that in Sphinx.

### LINKED ISSUE #1174: Document changes to admin settings configuration
https://github.com/opendatakit/collect/pull/3628#issuecomment-584574810

In particular, the .settings file export is separated from the QR code configuration.

### LINKED ISSUE #1182: Update screenshots to match main screen without "Main Menu" title and tagline
Change as of v1.26

### LINKED ISSUE #1183: Document option to scan configuration QR code from main screen
Added in v1.26

https://github.com/opendatakit/collect/pull/3637

lognaturel:
This PR should be merged concurrently with the Collect v1.26.0 release

Closes #1182
Closes #1174
Closes #1183
Closes #1164 - it's not exactly a Material Design icon font but most of the material design icons we may need have close analogs in fontawesome
Updates settings QR code keys and puts them in the same order as the settings menus.

I decided to do all of these together to avoid conflicts and because I thought it'd be easier to review a more complete set of changes. It also avoids conflicts.

This PR introduces a dependency on a fontawesome wrapper. This will make it easier to refer to icons inline without having to take screenshots. That library is not maintained but it doesn't matter because it's tiny and doesn't need to change.

I added a page on configuring ODK Collect because I always expect that settings will be described in the `Setting Up ODK Collect` section but in the current docs they're in `Collect Menus, Settings, and Security` which I don't find intuitive. I think we may want to move some of that around and refine it some more but it doesn't have to be done now.

I updated several screenshots as part of this PR. I took them with Android Studio  running https://github.com/opendatakit/collect/commit/fc67d5b6f90a4edd7e2a297f598c916624037166. I compressed them using ImageOptim.

I built and verified locally.


odk1-src/collect-import-export.rst
```diff
@@ -1,50 +1,43 @@
...
-
-Sharing QR code
+Settings QR codes
 ------------------
 
-You can click on the |share| icon to share the QR code as an image. When you click on it, it displays a list of applications and services like *whatsapp*, *facebook*, *hangouts*, *bluetooth*, *MMS* to name a few, which can be used to share the QR code. This is useful when there are several different data collection sites and all devices have to be configured in the same way, in which case the QR code can be shared from one reference device. 
+QR codes are a type of barcode that can encode a lot of data with low risk of error. These can be used to configure Collect on a fleet of devices. The QR image displayed on the :guilabel:`Configure via QR code` screen contains all of your current General and Admin settings. When this QR code is scanned from the ODK Collect app on another device, the settings from the code are imported.
```
seadowg: Not sure we need a description of what a QR code is here? Maybe you put that in based on people being confused.

odk1-src/collect-import-export.rst
```diff
@@ -1,50 +1,43 @@
...
 
-.. |share| image:: /img/collect-settings/share-icon.*
-             :alt: Share icon for sharing the QR code. 
-             :height: 43 px
-             :width: 43 px
+.. note::
+  QR codes generated by Collect only contain settings with non-default values because of the constraints on the amount of data a QR code can hold.
```
seadowg: I'm not sure this quite conveys the effect of not having default values included. Is it that non-default values in the app won't be overriden by the QR code?

odk1-src/collect-import-export.rst
```diff
@@ -1,50 +1,43 @@
...
+Sharing a QR code
+------------------
 
-  At app launch, the settings are loaded from a different location :file:`/sdcard/odk/collect.settings`, which tracks the state of your settings at all times. 
+You can tap on the :fa:`share-alt` icon to share the QR code as an image. When you click on it, it displays a list of applications and services like *whatsapp*, *facebook*, *hangouts*, *bluetooth*, *MMS* which can be used to share the QR code. This is useful when there are several different data collection sites and all devices have to be configured in the same way, in which case the QR code can be shared from one reference device. 
 
-  Settings are saved to and loaded from different locations to avoid deleting the saved settings.
+.. warning::
+  Since the QR code encodes compressed plain text settings and may contain the admin and server passwords, you should be careful about how you share it. For example, if you print out the QR code and tape it on wall, someone could use a standard QR code scanner to get the admin password.
```
seadowg: Definitely good to have this warning twice given they might only read one of the sections 👍

odk1-src/collect-import-export.rst
lognaturel: Ah, good point. I was just blindly improving what was there before.

odk1-src/collect-import-export.rst
lognaturel: Yes, you're right. I've tried to improve that. The note is now redundant with the text above it but I think that's ok because it's quite an important point.

PR #1194, opened 2020-03-12
Update references to the Collect root directory in preparation for the scoped storage migration

lognaturel:
Users of Collect from the Play Store will need to move their files to a scoped directory by August 2020. The migration will start being available in Collect v1.26.0. This PR replaces all references to `/sdcard/odk/`.

I searched the `odk1-src` directory for `/sdcard` and `/odk` and made replacements as appropriate. I decided to leave the .settings file documentation alone because it will get removed by #1192.


lognaturel:
Thanks, @grzesiek2010! I'll merge now since it's accurate even before the release.

PR #1201, opened 2020-03-27
Add Central SMTP auth details

### LINKED ISSUE #1200: Document how to add SMTP username/password to Central's email infrastructure
Most custom-mail servers will require a username and password, so we should document now to do that. Central uses [nodemailer](https://nodemailer.com/smtp/), so the typical settings are below.
```
"transportOpts": {
	"host": "smtp.sendgrid.net",
	"port": 587,
        "secure": false,
	"auth": {
	  "user": "apikey",
	  "pass": "blah-blah-blah"
	}
}
```
secure is for TLS and should be true if the port is 465, false for other ports. 

Glad to send in a PR if @issa-tseng is OK with this.

yanokwa:
Closes #1200

* Add SMTP auth settings
* Makes sure code blocks line up under the text it refers to.
* Add JSON metadata. No user-visible change.

You can confirm this all works:
Screen Shot 2020-03-27 at 1 46 27 PM

### DIFF
--- a/odk1-src/central-install-digital-ocean.rst
+++ b/odk1-src/central-install-digital-ocean.rst
@@ -205,9 +205,20 @@ Using a Custom Mail Server
 ODK Central ships with a basic EXIM server bundled to forward mail out to the internet. To use your own custom mail server:
 
 1. Ensure you have an SMTP relay server visible to your Central server network host.
-2. Edit the file ``files/service/config.json.template``:
-   * Under ``email``, then ``transportOpts``, you'll find settings for ``host`` and ``port``.
-   * These correspond with the network hostname and the TCP port, respectively.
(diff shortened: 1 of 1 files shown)


issa-tseng:
but these instructions will always apply authentication right? this will probably fail if the smtp server does not require credentials?

yanokwa:
Good catch, @issa-tseng. My guess is that most SMTP servers require auth. I've added https://github.com/opendatakit/docs/pull/1201/commits/f0272aa1604f5723384ea0ffb14e39948dc0045e to say that auth section should be removed if no auth is needed.

issa-tseng:
my guess is the opposite; i've mostly seen firewall checks for stuff like this. either way, looks good.

PR #1218, opened 2020-04-23
Smaller and white logo

yanokwa:
Screen Shot 2020-04-23 at 3 48 58 PM

Tried to match the size to the forum logo. I eyeballed it.

Changed the docs color to white to match the default look of Sphinx.


lognaturel:
To be honest, I rather like a bigger logo but 🤷🏻‍♀️

PR #1247, opened 2020-08-31
update documentation on date function

terencehonles:
### What is included in this PR?

Updates documentation on `date` function to match that it also accepts a string in ISO-8601 based on the comment by @yanokwa in https://forum.getodk.org/t/constraint-on-date/6355/2

### DIFF
--- a/odk1-src/form-operators-functions.rst
+++ b/odk1-src/form-operators-functions.rst
@@ -677,7 +677,7 @@ Converting dates and time
 
 .. function:: date(days)
 
-  Converts an integer representing a number of :arg:`days` from January 1, 1970 (the `Unix Epoch`_) to a standard date value.
+  Converts an integer representing a number of :arg:`days` from January 1, 1970 (the `Unix Epoch`_) or a string in ISO 8601 format to a standard date value.
 
   .. _Unix Epoch: https://en.wikipedia.org/wiki/Unix_time
     


terencehonles:
There also seems to be no documentation on the `date-time` function. Originally I had thought you must use `date(decimal-date-time('2000-01-01'))` based on a usage I found here https://github.com/getodk/collect/blob/8e7f360e587f3d5b90dd573471a84ead46af9d2d/collect_app/src/androidTest/assets/forms/validate.xml

specifically:
```xml
<bind calculate="decimal-date-time(concat( /validate/year ,&quot;-01-01&quot;))" nodeset="/validate/result" type="string"/>
```

but a coworker of mine found you could call the `date` function directly with a string. I had also found the following:

https://github.com/getodk/xforms-spec/blob/gh-pages/_sections/30-bindings.md#xpath-functions

but the `date` function doesn't specify what "Converts to a string in the ....date format." means. Is it ISO 8601? Should that be updated too?

lognaturel:
Thanks for this and sorry about the radio silence -- there are a few things to unpack here!

> found you could call the date function directly with a string

First, what are you trying to do with the date? I'm not even sure you need to call `date()` on it at all. Have you tried without? The internal date type gets serialized to a ISO 8601 string. It's possible we have a bug somewhere but it should be possible to use an ISO 8601 date string anywhere a date goes and have it automatically convert. 

> no documentation on the date-time function

As you said, this should be like `date` but include the `time` component. Similarly, it shouldn't really be necessary (or basically called implicitly on usage of a string as a dateTime). It would be good to see an example of the usage you're trying to make of it

terencehonles:
This would have to be answered by one of my coworkers, I believe we were doing some date math and I'm not sure if they tried without calling this function. I do believe one of the cases was `date(decimal-date-time('2000-01-01'))` but I'm not sure if I have the time to dig it up.

lognaturel:
> I believe we were doing some date math

More details would be much appreciated! It's possible that there's a bug lurking somewhere in there so it would be great to see if we can hunt it down. For now I'll close this PR as the usage it describes shouldn't be necessary.

PR #1251, opened 2020-09-07
Update ODK Central install docs on db with SSL

florianm:
See https://github.com/getodk/docs/issues/1250
Add the magic incantation to allow using a custom db with SSL.
Minor edit: database placeholder uses "my-db-name" in line with postgres terminology (and node-postgres docs).

closes https://github.com/getodk/docs/issues/1250

Updated wording in one paragraph in the docs.

None.

Merge.

https://github.com/getodk/central/issues/96

---
This change is [Reviewable](https://reviewable.io/reviews/getodk/docs/1251)

### DIFF
--- a/odk1-src/central-install-digital-ocean.rst
+++ b/odk1-src/central-install-digital-ocean.rst
@@ -292,6 +292,8 @@ ODK Central ships with a PostgreSQL database server. To use your own custom data
     CREATE EXTENSION IF NOT EXISTS pg_trgm;
 
 4. Edit the file ``files/service/config.json.template`` to reflect your database host, table, and authentication details.
+The key `"ssl": {"rejectUnauthorized": false}` is only required if the external database requires TLS/SSL 
+(see [node-postgres docs](https://node-postgres.com/features/ssl)), otherwise it can be omitted.
 
   .. code-block:: console
 
@@ -299,7 +301,8 @@ ODK Central ships with a PostgreSQL database server. To use your own custom data
       "host": "my-db-host",
       "user": "my-db-user",
       "password": "my-db-password",
-      "database": "my-db-table"
+      "database": "my-db-name",
+      "ssl": {"rejectUnauthorized": false}
     },
 
 5. Build and run: ``docker-compose build service``, ``docker-compose stop service``, ``docker-compose up -d service``.


lognaturel:
Thanks, @florianm, and I'm glad you were able to address your issue!

I don't think we should include this in official docs, though. It seems your database's certificate was probably self-signed or issued by an internal certificate authority so that's why you had to opt out of man-in-the-middle protection. The default should be to require a trusted certificate.

florianm:
My real world use case leading to this issue: Our IT provides infrastructure on the Azure cloud (substitute for your own provider). We use a hosted postgres database from a third party, as that db turns out to be faster than anything we can tune ourselves. Access is restricted to our internal network (outside my visibility) and likely a self signed cert. So there is no security issue being introduced as far as I can tell. 

Would a clarification/warning about dropping man in the middle protection (and when that is ok, like in our case) make this PR acceptable? Possibly a "if you run into error X, try require SSL false"?

lognaturel:
I think it's important that we make sure the default setup docs are as clear as possible without distractions. There are so many different kinds of environments that we risk boiling the ocean! That said, we're not there yet, so I think adding a note is ok. I like describing it as if you run into such and such error and you're confident using a self-signed certificate is safe, ...

florianm:
Great, I'll word it as a note!

Valid point to set a clear "golden path" and shunt the noise e.g. to an FAQ.
The R package drake labels such issues with "FAQ" and includes them into the FAQ section. That outsources the noise to GH.

lognaturel:
> The R package drake labels such issues with "FAQ" and includes them into the FAQ section

That's a great strategy! I'm proposing starting a troubleshooting section [here](https://github.com/getodk/docs/pull/1262/files#diff-446cfcd93389dd90ee7ae1d65c5b77a8) so that might be another place for this kind of content to eventually go.

lognaturel (changes requested):
Changing status so it's clear at a glance that we've agreed on some changes to make!

lognaturel:
Closing for now since this page has been restructured. I have a slight preference for this information going in [the troubleshooting section](https://docs.getodk.org/central-troubleshooting/) but a note would be fine, too.

PR #1260, opened 2020-09-23
Recommend open source recorder app

### LINKED ISSUE #1243: Recommended audio apps are out of date
https://docs.getodk.org/form-question-types/#audio-widget

Sony's app is now discontinued, so we should look at other apps to recommend.

lognaturel:
Closes #1243

https://gitlab.com/axet/android-audio-recorder/-/tree/master

I've verified that it responds to the necessary implicit intent to be launched by Collect. It supports pretty much any format someone would want, audio clipping, and various other nice things.

### DIFF
--- a/odk1-src/form-question-types.rst
+++ b/odk1-src/form-question-types.rst
@@ -1718,8 +1718,8 @@ Records audio from an external app.
 
   There are many apps available for this, including:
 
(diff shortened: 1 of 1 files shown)


seadowg (approved):
Ah this is the one on FDroid as well. Impossible to find in the Play Store but really good.

PR #1262, opened 2020-09-23
Describe limited support for Central on environments other than DO

lognaturel:
We need to be explicit that the core team has limited ability to verify Central releases on platforms other than Digital Ocean and that we generally won't be able to provide free support for custom installs.

I tried to make the message direct but still friendly. I went with "community support" but also considered "free support". I considered saying "because Central is developed by a small team" or something like that but it didn't seem all that useful.

### DIFF
--- a/odk1-src/central-troubleshooting.rst
+++ b/odk1-src/central-troubleshooting.rst
@@ -1,0 +1,21 @@
+.. _central-troubleshooting:
+
+Troubleshooting Central 
+=========================
+
+.. _email-failures:
+
+Users aren't receiving Central emails
+--------------------------------------
+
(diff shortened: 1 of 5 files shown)


odk1-src/central-troubleshooting.rst
```diff
@@ -0,0 +1,21 @@
+.. _central-troubleshooting:
+
+Troubleshooting Central 
+=========================
+
+.. _email-failures:
```
yanokwa, line 6: Nit, but I kinda like `troubleshooting-emails`

yanokwa (changes requested):
Great work, @lognaturel. Really like the troubleshooting section. I would like to see a tip after the last paragraph at Logging into ODK Central that says...

> If you find that users are not receiving emails, read :ref:`troubleshooting email<central-troubleshooting>`

yanokwa (approved):

PR #1263, opened 2020-09-25
Add background on Central backups

lognaturel:
We've gotten various questions about Central backups and where they send data (e.g. https://github.com/getodk/central/issues/167). This PR clarifies that managing external backups is a possibility and may be a better option for a lot of folks.

### DIFF
--- a/odk1-src/central-backup.rst
+++ b/odk1-src/central-backup.rst
@@ -3,6 +3,13 @@
 Backing Up Central
 ==================
 
+Having a data backup strategy is a critical part of running a web service like ODK Central. Backups should go to a system in a different physical location from where Central is installed in order to prevent data loss across a broad range of scenarios. Many cloud providers offer backup strategies that run automatically. If you are an experienced system administrator, you may want to set up your own backups of the PostgreSQL database. If you don't already have a server-wide backup system in place and don't want to set up your own database backup, Central provides a managed backup system to Google Drive.
+
+.. _central-managed-backups:
+
+Managed backups
+---------------
+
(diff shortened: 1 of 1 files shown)


odk1-src/central-backup.rst
```diff
@@ -3,6 +3,13 @@
 Backing Up Central
 ==================
 
+Having a data backup strategy is a critical part of running a web service like ODK Central. Backups should go to a system in a different physical location from where Central is installed in order to prevent data loss across a broad range of scenarios. Many cloud providers offer backup strategies that run automatically. If you are an experienced system administrator, you may want to set up your own backups of the PostgreSQL database. If you don't already have a server-wide backup system in place and don't want to set up your own database backup, Central provides a managed backup system to Google Drive.
```
yanokwa, line 6: "of the PostgreSQL database that contains all Central's data."

Basically, we need a lightweight way of saying that everything that is in Central is in that database. That is, there is nothing else important on the filesystem.

odk1-src/central-backup.rst
```diff
@@ -50,7 +57,11 @@ Setting up backups
 Restoring a backup
 ------------------
 
-Restoring a backup to an ODK Central instance will entirely replace all of its data with the backup. Please be very sure you are restoring to the right place with the right backup snapshot before proceeding.
+Restoring a backup to an ODK Central instance will entirely replace all of its data with the backup. Please be very sure you are restoring to the right place with the right backup snapshot before proceeding.
```
yanokwa, line 60: Too many ODK Centrals. Just Central is fine.

yanokwa (changes requested):

lognaturel:
Regarding versions, maybe what we should say is if things don't work, do `docker-compose stop`, `docker-compose build`, `docker-compose up -d`. I believe that would run migrations. @issa-tseng?

yanokwa (approved):

PR #1266, opened 2020-10-09
Docs 1257

### LINKED ISSUE #1257: Update section on select question types to reflect new minimal appearance
https://docs.getodk.org/form-question-types/#select-widget-with-modal-menu and below.

grzesiek2010:
closes #1257

The pr includes updates to select widgets according to the demo form: https://docs.google.com/spreadsheets/d/1af_Sl8A_L8_EULbhRLHVl8OclCfco09Hq2tqb9CslwQ/edit#gid=0


odk1-src/form-question-types.rst
```diff
@@ -766,102 +766,63 @@ This is especially helpful for questions with a large number of choices.
...
   :tc:`select_one {list_name}`
 appearance
-  *none*
+  *columns-pack*
 
-Select widgets support image choices.
-The images are referenced in the **choices** sheet,
-and the image files
-need to be included in the :file:`media` folder.
+When the :tc:`columns-pack` appearance is added, the app tries to accommodate as many choices in a single line as possible.
```
lognaturel, line 779: Add: if the choice labels have different lengths, they will not be in even columns.

odk1-src/form-question-types.rst
```diff
@@ -906,30 +862,27 @@ specify an :th:`appearance` of :tc:`compact-2`.
...
 appearance
-  :tc:`quickcompact`
-
-The :tc:`quickcompact` appearance
-adds :ref:`autoadvance <autoadvance>` functionality
-to the design of the :ref:`compact-image-select`.
+  :tc:`no-buttons`
 
-.. video:: /vid/form-widgets/quickcompact.mp4
+When the :tc:`no-buttons` appearance is added, the app does not display buttons but images or labels if images are missing.
```
lognaturel, line 876: "the app does not display buttons but images or labels if images are missing." -> "the app displays choice labels and/or images without the selection radio button. This is particularly useful for building a grid of images."

lognaturel (changes requested):
Thank you, looks great! I made a couple of changes to the form source:
- changed "spinner" to "select one minimal" since I don't think normal people know what a "spinner" is and we don't use one anymore, anyway!
- removed "grid" widget in favor of describing the column usage

I don't think the screenshots here need to be updated but wanted to let you know and see if you had any pushback.

odk1-src/form-question-types.rst
grzesiek2010: `And` shouldn't be used here. We display images if media files are provided or labels otherwise never both. Right?

lognaturel: Ah! Yes, I think you're right. That's not what I got from your original sentence. Something like "the app displays choices without the selection radio button. If images are specified for choices, only the images are displayed. This is particularly useful for building a grid of images."

grzesiek2010: Ok, fixed.

odk1-src/form-question-types.rst
grzesiek2010: Fixed.

lognaturel (approved):

PR #1268, opened 2020-10-09
Remove odk-x and odk1 references

yanokwa:
* Dropped all references to `odk1`
* Moved all shared-src into the src directory
* Confirmed `make build` renders site
* Confirmed that icons, footers, and contributing from shared-src is visible
* Confirmed show screenshot JS works
* Confirmed videos load


lognaturel:
Let's merge #1266 first.

lognaturel (approved):

PR #1280, opened 2020-11-19
Update docs for audio changes

seadowg:
addresses https://github.com/getodk/collect/projects/4#card-49212675

* Screenshot updated
* Description updated
* Section on quality parameter
* Section on using external apps

### DIFF
--- a/src/form-question-types.rst
+++ b/src/form-question-types.rst
@@ -1703,7 +1703,15 @@ The quality of audio recordings can be customized using the ``quality`` paramete
      - 8kHz
      - ~5MB/hour
 
+.. tip::
+
+  We'd recommend only using ``voice-only`` for one-on-one interviews in a quiet place as otherwise there might be too much detail loss. If in doubt, it's a good idea to test the different qualities out to see which one fits your use case best.
+
+  If it's a possibility that an individual question could need different qualities depending on context you can use :ref:`relevance <relevants>` to switch between them.
+
 If no ``quality`` is specified, ``normal`` is used.
+
+
 
 .. rubric:: XLSForm
 
--- a/src/form-question-types.rst
+++ b/src/form-question-types.rst
@@ -1700,7 +1700,7 @@ When built-in audio recording is enabled and recording is initiated, a recording
 
 .. warning::
 
-  Pause is only available on Android 7.0 and above. On lower Android versions, the pause button is hidden. 
+  Pause is only available on Android 7.0 and above. On lower Android versions, the pause button is hidden.
 
 At the right of the control bar are a pause button (3) and a stop button (4). When the pause button is tapped, recording is temporarily suspended and the button icon changes to a microphone. When the microphone is tapped, recording is resumed. Recording can be paused and resumed as many times as desired. When the stop button is tapped, the recording is ended and can no longer be modified.
 
@@ -1768,25 +1768,26 @@ In the parameters column, write ``quality=`` followed by the desired value.
 
  audio,voice_only_audio,Voice audio,quality=voice-only
 
+Changing audio quality during form entry
+"""""""""""""""""""""""""""""""""""""""""
+
 If it's a possibility that an individual question could need different qualities depending on context you can use :ref:`relevance <relevants>` to switch between them:
 
-  .. container:: details
+.. rubric:: XLSForm
 
-    .. rubric:: XLSForm
+.. csv-table:: survey
+  :header: type, name, label, parameters, relevance
 
-    .. csv-table:: survey
-      :header: type, name, label, parameters, relevance
+  select_one, yes_no, is_quiet, Are you currently in a quiet location with only one person speaking at a time?
 
-      select_one, yes_no, is_quiet, Are you currently in a quiet location with only one person speaking at a time?
+  audio recording_voice_only, Please record, quality='voice-only',, ${is_quiet} = 'yes'
+  audio recording_normal, Please record, quality='normal',, ${is_quiet} = 'no'
 
-      audio recording_voice_only, Please record, quality='voice-only',, ${is_quiet} = 'yes'
-      audio recording_normal, Please record, quality='normal',, ${is_quiet} = 'no'
+.. csv-table:: choices
+  :header: list_name, name, label
 
-    .. csv-table:: choices
-      :header: list_name, name, label
-
-      yes_no, yes, Yes
-      yes_no, no, No
+  yes_no, yes, Yes
+  yes_no, no, No
 
 .. _external-audio-app:
 
--- a/src/form-question-types.rst
+++ b/src/form-question-types.rst
@@ -1666,15 +1666,21 @@ appearance
 
 Records audio using the device's microphone or a connected external microphone. By default, an :ref:`external application <external-audio-app>` is used. Starting with Collect v1.29, you can also use :ref:`built-in recording <built-in-audio-recording>`.
 
-.. image:: /img/form-widgets/audio-start.*
-  :alt: The Audio form widget as displayed in the ODK Collect App on an Android phone. The question text is "What does it sound like?" There are two buttons: Record Sound and Choose Sound.
+Audio files can quickly become very big so if you record audio in your form, make sure that you carefully consider your audio quality settings. Also test making submissions to your server with the Internet conditions you expect when gathering data to make sure that you can send files of the size you expect.
+
+Android devices can make many sounds during use and these will be included in recordings. We recommend turning off sounds from button presses, camera shutters and notifications before recording.
+
+  .. container:: details
+
+    .. image:: /img/form-widgets/audio-start.*
+      :alt: The Audio form widget as displayed in the ODK Collect App on an Android phone. The question text is "What does it sound like?" There are two buttons: Record Sound and Choose Sound.
 
 .. rubric:: XLSForm
 
 .. csv-table:: survey
   :header: type, name, label
 
-  audio, sound_like, What does it sound like?
+  audio, bird_recording, What does it sound like?
 
 .. _built-in-audio-recording:
 
--- a/src/form-question-types.rst
+++ b/src/form-question-types.rst
@@ -1703,6 +1703,15 @@ The quality of audio recordings can be customized using the ``quality`` paramete
      - 8kHz
      - ~5MB/hour
 
+.. rubric:: XLSForm
+
+In the parameters column, write ``quality=`` followed by the desired value.
+
+.. csv-table:: survey
+ :header: type, name, label, parameters
+
+ audio,voice_only_audio,Voice audio,quality=voice-only
+
 Recording with an external app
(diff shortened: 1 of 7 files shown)


src/form-question-types.rst
```diff
@@ -1655,33 +1655,86 @@ In the parameters column, write ``max-pixels=`` followed by the desired maximum
...
+     - AMR
+     - 12.2kbps
+     - 8kHz
+     - ~5MB/hour
+
+.. tip::
+
+  We'd recommend only using ``voice-only`` for one-on-one interviews in a quiet place as otherwise there might be too much detail loss. If in doubt, it's a good idea to test the different qualities out to see which one fits your use case best.
+
+  If it's a possibility that an individual question could need different qualities depending on context you can use :ref:`relevance <relevants>` to switch between them.
```
lognaturel, line 1707: I think this is not going to be actionable for most people. I wonder whether for things like this we might consider adding support questions that we answer ourselves and then link to them? Alternately we could use a folder in the `getodk` Google Drive to have form definitions we link to.

src/form-question-types.rst
```diff
@@ -1655,33 +1655,86 @@ In the parameters column, write ``max-pixels=`` followed by the desired maximum
...
+ :header: type, name, label, parameters
+
+ audio,voice_only_audio,Voice audio,quality=voice-only
+
+Recording with an external app
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. warning::
+
+  Before Collect 1.29 audio questions would always use an external app.
```
lognaturel, line 1726: Is this the most useful place for this warning? Maybe it's more the `Customizing audio quality` section that needs a warning saying it's only available in 1.29+.

src/form-question-types.rst
```diff
@@ -1655,33 +1655,86 @@ In the parameters column, write ``max-pixels=`` followed by the desired maximum
...
+  :tc:`none`
 
-.. note::
+Records audio using the device's microphone.
 
-  Some Android devices provide a default application for audio recording.
-  Others do not,
-  and the user will need to install an audio recording app.
+.. image:: /img/form-widgets/audio-start.*
+  :alt: The Audio form widget as displayed in the ODK Collect App on an Android phone. The question text is "What does it sound like?" There are two buttons: Record Sound and Choose Sound.
```
lognaturel, line 1667: It's not obvious from these docs that I can record while answering questions. I think a couple of sentences on that, a link to the field-list docs and maybe even a screenshot would be really helpful.

lognaturel (commented):
I really like the structure!

src/form-question-types.rst
lognaturel: Some other things to add:
- description of behavior while recording (can't leave question screen in any way)
- behavior of pause vs. stop

src/form-question-types.rst
seadowg: Yeah that sounds better to me!

src/form-question-types.rst
seadowg: This has been discussed a few times right? Like a gallery of example forms. I think if they are docs they should probably live as part of the docs code. Should we just do an inline example here?

src/form-question-types.rst
seadowg: Yeah all sounds good! I can look at adding this in.

src/form-question-types.rst
lognaturel: > if they are docs they should probably live as part of the docs code

I see that. How about having it inline but collapsed by default? Like "show details" in https://docs.getodk.org/briefcase-install/, for example (though I wouldn't say that's a very compelling use of the feature). So you can keep exactly the text you have now as a new paragraph and then maybe "show worked example" or even just "show details" right below?

lognaturel: To be explicit, my concern is that we kind of want users to be able to skim vertically down the page and get everything that there is to get about audio recording. If there's a very long example related to an uncommon case then they may get distracted and miss important things below.

I think longer-term we should consider splitting up this very long page so topics are a bit better contained. But the collapsing of long examples seems like a fine approach for now.

seadowg: > How about having it inline but collapsed by default?

I like that idea! Will go with that.

lognaturel:
Some other things we've thought of or that have been added:
- [ ] Built-in recording is off by default in v1.29, turn it on through form design or settings update
- [ ] Add admin and general keys to JSON dict for settings
- [ ] Pausing (only available in Android 7.0+)
- [ ] Sounds from shutters, button presses will be captured and should be turned off
- [ ] If copying fails, where to find the files, mention they’re not cleaned up
- [ ] Describe using with multiple questions in a field list + sample collapsed by default

src/form-question-types.rst
lognaturel: Consolidated the rest of these bullets in https://github.com/getodk/docs/pull/1280#issuecomment-740836339

lognaturel:
I wasn't able to push to your branch, @seadowg. Not sure if that's because of the settings, because I rebased on top of master or what. https://github.com/getodk/docs/compare/master...lognaturel:audio is a branch with my updates which are reflected in the checklist above. Feel free to `cherry-pick` and keep working on it or I can finish the remaining checklist items tomorrow!

lognaturel:
I'm feeling pretty happy with where https://github.com/getodk/docs/compare/master...lognaturel:audio is at. @seadowg over to you for a final polish pass!

src/form-question-types.rst
```diff
@@ -1658,33 +1659,151 @@ In the parameters column, write ``max-pixels=`` followed by the desired maximum
...
+.. _customizing-audio-quality:
+
+Customizing audio quality
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 1.29
+
+  `ODK Collect v1.29.0 <https://github.com/getodk/collect/releases/tag/v1.29.0>`_, `pyxform` v1.3.0, `XLSForm Online` v2.3.0
+
+The quality of audio recordings can be customized using the ``quality`` parameter. If a ``quality`` is specified, the built-in recorder is always used, regardless of Collect settings. If no ``quality`` is specified and Collect is :ref:`configured to use the built-in recorder <use-external-app-for-audio-recording>`, ``normal`` is used. The available quality values are:
```
seadowg, line 1728: "If a `quality` is specified, the built-in recorder is always used, regardless of Collect settings" isn't strictly true but it does feel like explaining the `external` exception might not make sense seeing as it's documented below. What do you think?

src/form-question-types.rst
```diff
@@ -1658,33 +1659,151 @@ In the parameters column, write ``max-pixels=`` followed by the desired maximum
...
+     - 32kHz
+     - ~11MB/hour
+   * - voice-only
+     - .amr
+     - AMR
+     - 12.2kbps
+     - 8kHz
+     - ~5MB/hour
+
+.. tip::
```
seadowg, line 1758: This felt right in a tip but not 100% sure I've found the right place for it.

src/form-question-types.rst
lognaturel: Yes, that sounds right to me. In this context it seems like what's most useful.

src/form-question-types.rst
```diff
@@ -1658,33 +1659,151 @@ In the parameters column, write ``max-pixels=`` followed by the desired maximum
...
-  - `RecForge II <https://play.google.com/store/apps/details?id=dje073.android.modernrecforge>`_
+.. csv-table:: survey
+  :header: type, name, label
 
-  Any app that responds to
-  `android.provider.MediaStore.Audio.Media.RECORD_SOUND_ACTION`
-  will be compatible.
+  audio, bird_recording, What does it sound like?
 
+.. tip::
```
seadowg, line 1679: This felt right in a tip but not 100% sure I've found the right place for it.

src/form-question-types.rst
```diff
@@ -1658,33 +1659,151 @@ In the parameters column, write ``max-pixels=`` followed by the desired maximum
...
+appearance
+  :tc:`none`
 
-.. note::
+Records audio using the device's microphone or a connected external microphone. By default, an :ref:`external application <external-audio-app>` is used. Starting with Collect v1.29, you can also use :ref:`built-in recording <built-in-audio-recording>`.
 
-  Some Android devices provide a default application for audio recording.
-  Others do not,
-  and the user will need to install an audio recording app.
+.. image:: /img/form-widgets/audio-start.*
```
seadowg, line 1669: I took the container out from around this screenshot as the whitespace wasn't behaving correctly when it was minimized (very little space between it and the rubric below). It's also the only widget with a minimized screenshot as far as I could see? Happy for it to go back in if there are strong feelings about it.

src/form-question-types.rst
```diff
@@ -1658,33 +1659,151 @@ In the parameters column, write ``max-pixels=`` followed by the desired maximum
...
+   * - voice-only
+     - .amr
+     - AMR
+     - 12.2kbps
+     - 8kHz
+     - ~5MB/hour
+
+.. tip::
+
+  We'd recommend only using ``voice-only`` for one-on-one interviews in a quiet place as otherwise there might be too much detail loss. ``low`` will sound compressed but speech is generally intelligible, even if multiple people are talking at once. ``normal`` is similar to typical podcast settings and will sound good on most devices. If in doubt, it's a good idea to test the different qualities out to see which one fits your use case best.
```
seadowg, line 1760: Not sure about "podcast quality" here. I'm guessing you're right that a lot of podcasts would be delivered at this kind of quality, but they'd probably be recorded with good mics/gear and then encoding. Maybe I'm overthinking?

lognaturel: I see. Maybe another take is "``normal`` will sound good on most devices and the limiting quality factor will generally be the quality of the microphone and the recording environment rather than the encoding."

seadowg: Actually I like the use of "podcast" more and more as I think about it as it feels like it doesn't overpromise around non voice recordings. We're encouraging people to try them out which feels like the most important thing. Forget I said antyhing!

src/form-question-types.rst
lognaturel: It just doesn't feel all that useful to me but let's leave it as-is for now and we can always tweak!

src/form-question-types.rst
lognaturel: Seems good to me! 👍

src/form-question-types.rst
lognaturel: Maybe there should be a tip or note saying that mic quality and recording environment are going to be limiting factors. I do think that's important to set expectations and also maybe give a hint to someone who's not getting the results they want that another device or an external mic could help.

seadowg: Ok to address that I've tried to add more emphasis to our note on testing things out. What do you think?

lognaturel (approved):

PR #1293, opened 2020-12-14
Add warning and recovery instructions for docker-compose down

lognaturel:
Central's `docker-compose` configuration uses all anonymous volumes which can result in perceived data loss and exceeding Let's Encrypt quotas for folks who use `docker-compose down` instead of following our instructions exactly. We've explored changing the `docker-compose` config but it's risky: https://github.com/getodk/central/compare/issa/move-postgres-data. This PR is an alternative that tries to preemptively warn users not to use `docker-compose down` and provides instructions for those that do anyway.

I played with some language saying we know it's an uncommon setup or that it's an accident of history but I ended up deciding it's not necessary.

I also thought about mentioning the Let's Encrypt quota issue but there's not really a whole lot a user can do to recover once they hit that state other than wait so I don't think it's worth mentioning.

I went back and forth on whether or not to provide instructions inline in the docs but I think it's happened enough times that it's helpful to have.

### DIFF
--- a/src/central-install.rst
+++ b/src/central-install.rst
@@ -5,6 +5,9 @@ Installing Central
 
 Central is distributed and installed using `Docker <https://en.wikipedia.org/wiki/Docker_(software)>`_. Docker makes it possible to describe exactly how Central's different components should be configured no matter where it is installed. Don't worry if you don't know about Docker yet! We have put together step-by-step instructions for our recommended solutions below.
 
+.. warning::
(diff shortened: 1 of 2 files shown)


issa-tseng:
ln... i did this already... :(
https://github.com/getodk/docs/pull/1281/commits/e2b70b033443ea6f777dbe85eb85428532830c9c

lognaturel:
Whoops, well, not a big deal, it was just adapting the instructions from the forum. I think the important thing is to have a preemptive warning as well and we can talk about where that should go.

issa-tseng:
yeah we can merge that part of your pr into the other one

lognaturel:
Sounds good!

PR #1297, opened 2020-12-16
Corrected greater than less than condition example

### LINKED ISSUE #1296: Form logic: "Validating and restricting responses" example incorrect greater/less than symbols
In the "Form logic: Validating and restricting responses" section there is an example constraint:

```
. < 20 and . > 200
```

The documentation asserts that that example will evaluate to True if the response is between 20 and 200. This is incorrect. This example will instead evaluate universally to False. I suggest changing the example above to:

```
. > 20 and . < 200
```

I'll open a Pull Request momentarily.

johngodlee:
closes #1296 

Previously the example erroneously suggested that `. < 20 and . > 200` would evaluate to True  if the value was between 20 and 200. In actuality, this would evaluate universally to False. I switched around the direction of the greater/less than symbols.

### DIFF
--- a/src/form-logic.rst
+++ b/src/form-logic.rst
@@ -436,7 +436,7 @@ For example:
 :tc:`. >= 18`
   True if response is greater than or equal to 18.
 
-:tc:`. < 20 and . > 200`
+:tc:`. > 20 and . < 200`
   True if the response is between 20 and 200.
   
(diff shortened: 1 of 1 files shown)


lognaturel (approved):
Good catch, thanks so much! Looks like it was an upside down day when the original expression went in. 🙃🙃🙃

PR #1300, opened 2020-12-23
Central v1.1

lognaturel:
Completes #1281

I can't push to @issa-tseng's branch because of gitlfs things I don't fully understand.

I ran the new image through ImageOptim. If I had more time I'd make a better screenshot but I think the important thing is to have most of the functionality shown.

I ran this locally.

### DIFF
--- a/src/central-backup.rst
+++ b/src/central-backup.rst
@@ -56,6 +56,15 @@ Setting up backups
 
   You can verify your Google Drive usage `on the Drive storage page <https://drive.google.com/settings/storage>`_. You may want to periodically remove older backups to free up space.
 
+.. _central-backup-immediate:
+
+Performing an immediate backup
+------------------------------
+
+It is possible to immediately download a backup of your database to your own computer. As of Central v1.1, you will still need to have managed backups configured. Once you do, you will see a button :guilabel:`Download Backup Now` next to the :guilabel:`Terminate` button near the top.
+
+Clicking this button will perform an immediate backup and download the result to your computer. This process can take some time, and it is normal for data to download quite slowly for many minutes before it gets faster. Take care in using this feature particularly if you have a lot of data and traffic, as performing a backup while a lot of data is being saved to the database can cause a lot of slowdown.
+
 .. _central-backup-restore:
 
 Restoring a backup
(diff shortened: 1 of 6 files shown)


src/central-backup.rst
```diff
@@ -56,6 +56,17 @@ Setting up backups
...
+.. _central-backup-immediate:
+
+Performing an immediate backup
+------------------------------
+
+It is possible to immediately download a backup of your database to your own computer. As of Central v1.1, you will still need to have managed backups configured in order to access this functionality in the web management interface. Once you do, you will see a button :guilabel:`Download Backup Now` next to the :guilabel:`Terminate` button near the top.
+
+Clicking this button will perform an immediate backup and download the result to your computer. This process can take some time, and it is normal for data to download quite slowly for many minutes before it gets faster. Take care in using this feature particularly if you have a lot of data and traffic, as performing a backup while a lot of data is being saved to the database can cause a lot of slowdown.
+
+You can also use `the API <https://docs.getodk.org/central-backup/#performing-an-immediate-backup>`_ to download a backup without configuring managed backups.
```
matthew-white, line 68: It looks like this link points back to this section of the docs rather than to the API docs. (I think the way this link was placed in the forum announcement may have been confusing.)

src/central-submissions.rst
```diff
@@ -72,13 +78,19 @@ To find the Form submissions page, first find the form in the Form listings page
 
    .. image:: /img/central-submissions/listing.png
 
-The table preview you see here will show you the first ten fields of your survey and their results, with the latest submissions shown closest to the top. Any downloadable files will appear with a green download link you can use to directly download that media attachment. The submission's instance ID will always be shown at the right side of this preview table.
+The table preview you see here will at first show you the first ten fields of your survey and their results, with the latest submissions shown closest to the top. Any downloadable files will appear with a green download link you can use to directly download that media attachment. The submission's instance ID will always be shown at the right side of this preview table.
+
+If your form has more than ten fields, you can show more columns by accessing the :guilabel:`Columns shown` dropdown and checking the columns you wish to see. While the Columns shown pane is open, you can use your browser's search feature (usually Ctrl+F or Cmd+F) to search for particular column name if you have many.
```
matthew-white, line 83: We ended up adding a search field to the dropdown itself, so the user actually shouldn't need to use Ctrl/Cmd+F.

matthew-white (commented):
I read the new docs here and compared them to #1281. This is looking good! 🚀

src/central-submissions.rst
lognaturel: I did go back and forth on this. I think it's a useful tip that I wouldn't probably have thought to try. I'll change it to "the search box or your browser's search feature".

matthew-white (approved):

yanokwa (approved):

PR #1304, opened 2021-01-05
Improve swap commands

yanokwa:
* I prefer Ubuntu's instructions because seems likely to be a better source of truth
* Add swap to fstab! Not sure how we missed this.
* Reducing the swappiness of swap so it's only used when absolutely necessary. Default is 60.

Screen Shot 2021-01-07 at 12 08 53

### DIFF
--- a/src/central-install-digital-ocean.rst
+++ b/src/central-install-digital-ocean.rst
@@ -225,16 +225,27 @@ If you have installed Central on a 1GB droplet, you may encounter problems expor
 
 In this case, the first thing you can try is to add a swap file. We **do not** recommend adding swap unless you are struggling with attachment exports, and if you can afford it, upgrading to a 2GB machine will yield much better results than adding swap. But if you just need your export to work for now, this can be an effective workaround.
 
-Log into your server so you have a console prompt, and run these commands, adapted from `this article <https://linuxize.com/post/create-a-linux-swap-file/>`_:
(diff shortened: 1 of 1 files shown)


src/central-install-digital-ocean.rst
```diff
@@ -225,16 +225,22 @@ If you have installed Central on a 1GB droplet, you may encounter problems expor
...
 
  fallocate -l 1G /swap
- dd if=/dev/zero of=/swap bs=1024 count=1048576
+ dd if=/dev/zero of=/swap bs=1k count=1024k
  chmod 600 /swap
  mkswap /swap
  swapon /swap
+ sysctl vm.swappiness=10
 
+Finally, run ``nano /etc/fstab`` and add the following to the end of the file.
```
lognaturel, line 239: I think it's helpful to say "To make sure swap is there after a reboot, ..."

src/central-install-digital-ocean.rst
```diff
@@ -225,16 +225,22 @@ If you have installed Central on a 1GB droplet, you may encounter problems expor
...
 
 .. code-block:: console
 
  fallocate -l 1G /swap
- dd if=/dev/zero of=/swap bs=1024 count=1048576
+ dd if=/dev/zero of=/swap bs=1k count=1024k
  chmod 600 /swap
  mkswap /swap
  swapon /swap
+ sysctl vm.swappiness=10
```
lognaturel, line 237: According to your source doc this will not persist between reboots. You'd need to edit `/etc/sysctl.conf` for that.

lognaturel (approved):

PR #1308, opened 2021-01-15
Describe lightweight timestamping

### LINKED ISSUE #1292: Add section on lightweight timestamping using trigger
When a specific form value changes, a `calculation` of `now()` can be triggered.

lognaturel:
closes #1292

The primary goal of this PR is to add a section on light weight timestamping. I also added some general context on the `trigger` column and made light improvements to the audit docs.

I'm not entirely satisfied with where the `Triggering calculations on value change` section ends up. It's certainly not as important as constraints, etc. Ideally we'd have a page for calculates that would progress into more advanced uses such as this. Given the current monolithic form logic page, I think this is reasonable.

### DIFF
--- a/src/form-logic.rst
+++ b/src/form-logic.rst
@@ -14,6 +14,8 @@
   Onwuachi
   sophia
   timestamp
+  timestamping
+  dateTime
 
 ***********
 Form Logic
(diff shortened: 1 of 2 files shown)


src/form-logic.rst
```diff
@@ -410,7 +412,51 @@ In the example above, when a participant is selected, his or her phone number is
...
+:ref:`Calculations <calculations>` are recomputed any time one of the values in its expression changes. For example, if your form includes the calculation `${q1} + ${q2}`, it will be recomputed any time either of the values for `${q1}` or `${q2}` changes.
+
+Calculations can also be triggered when a value not involved in the calculation changes. This uses the same ``trigger`` column as  :ref:`defaults from form data <defaults-from-form-data>`. It is particularly useful for triggering calculations that involve values not in the form like random numbers or time.
+
+.. _lightweight_timestamping:
+
+Lightweight timestamping
+-------------------------------
+
+Knowing how long an enumerator spent answering a question can help with quality control and training. ODK Collect provides an :doc:`audit log <form-audit-log>` that contains rich information about how a form was answered. However, it can be unnecessarily complex for answering questions about a small number of questions. In that case, you may prefer to collect a timestamp when a particular question's value changes. This is similar to the ``start`` and ``end`` timestamp :ref:`metadata types <metadata>`.
```
yanokwa, line 434: s/a form was answered/an enumerator navigated a form/. 
* Match what the form-audit-log says. "How a form was answered" reads weird to me.

>However, it can be unnecessarily complex if all you need is lightweight timestamping of a few questions (similar to the ``start`` and ``end`` timestamp :ref:`metadata types <metadata>`.)
* Too much roundabout text. Just say the thing!
* There are too many "questions" in the text and overloading of "question."
* I don't love "Collect a timestamp" when you say "capture" below. Also the apps name is Collect, which makes it weird.

yanokwa (changes requested):
Reads really well. I thought it'd be super dry, but I think it's compelling enough to share on social media. Nicely done!

PR #1309, opened 2021-01-16
Move and deprecate Aggregate-specific content

yanokwa:


src/form-update.rst
```diff
@@ -31,9 +31,8 @@
 Updating forms in Aggregate
 ============================
 
-.. note::
-
-  This page is only about ODK Aggregate. ODK Central allows any kind of form update that doesn't change the type of a field. See :ref:`central-forms-updates`.
```
lognaturel, line 36: I think the second sentence is helpful and should be maintained. It can give folks landing on this page more of a reason to switch.

lognaturel (approved):
@yanokwa doesn't want to keep the second sentence and it's not important enough for me to fight about.

PR #1325, opened 2021-02-13
Update to demo server

yanokwa:
Describe new and old ways of getting version.

Screen Shot 2021-02-17 at 08 48 49


lognaturel:
I think we should say something like "If your Central server has a question mark icon in the upper right, click on it and then click on `Version`. If your Central server does not..." Hopefully servers without this functionality will be phased out soon.

lognaturel (approved):

PR #1327, opened 2021-02-17
Add background audio documentation

lognaturel:
To be shipped with Collect v1.30

I recommend looking at the two commits separately. There's a little bit of redundancy with the foreground audio section but I couldn't find smooth ways to link between them. This feels ok and I think it's fine if the wording ends up deviating a little bit.

Any screenshot needed? I've gone back and forth on adding one of the top bar.

TODO:
- [x] Basic sample form
- [ ] Examples of introduction screens. I'd like to do one that's just informational, one that asks for consent and one that allows the respondent to pick topics they're comfortable discussing while being recorded. I'm hesitating between attaching XLSX files and putting the form defs inline but collapsed by default. I'd love thoughts on that or other options!

### DIFF
--- a/src/form-question-types.rst
+++ b/src/form-question-types.rst
@@ -2247,6 +2247,70 @@ Collects a signature from the user.
 
   image,signature_widget,Signature widget,signature,image type with signature appearance
 
+
+.. _field-list:
+
+Grouping multiple widgets on the same screen
+------------------------------------------------
+
+type
+  :tc:`begin_group`
+appearance
+  :tc:`field-list`
+
+The :tc:`field-list` appearance attribute, applied to a group of widgets, displays them all on a single screen.
+
+.. warning::
+
+  Relevance, constraint and calculation evaluation within the same screen is supported in Collect v1.22 and later.
+
+.. warning::
+
+  Displaying :ref:`repeats` on the same screen (inside a :tc:`field-list` group) is not supported.
+
+.. seealso::
+
+  :ref:`groups` and :ref:`repeats`.
+
+.. _select-grid:
+
+Grid of selects on the same screen
+------------------------------------
+
+If you have multiple select questions with the same choices, it can be helpful to group them on one screen.
+
+.. image:: /img/form-widgets/select-grid.*
+  :alt: A field-list group of questions, as displayed in the ODK Collect app on an Android phone. A grid of questions representing underlying conditions are displayed. For eacn condition, there are radio buttons to indicate 'Yes' or 'No'.
+
+
+To do this, put your select questions in a :tc:`field-list` group and use the following :th:`appearance` attributes:
+
+:tc:`label`
+  Only the option labels are displayed, without checkboxes. This is used for the top row with the 'Yes' and 'No' options in the example above.
+:tc:`list-nolabel`
+  Only checkboxes or radio buttons are displayed, without their labels. This is used for the question rows in the example above.
+:tc:`list`
+  The labels are displayed along with checkboxes for multi-select questions and radio buttons for single-select questions. You could use this instead of having a :tc:`label` row to keep the option labels closer to the checkboxes or radio buttons.
+
+.. rubric:: XLSForm
+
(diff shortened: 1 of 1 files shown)


src/form-question-types.rst
```diff
@@ -2399,4 +2400,44 @@ storing the values for later use.
...
+.. rubric:: XLSForm
+
+.. csv-table:: survey
+  :header: type, name
+
+  background-audio, my_recording
+
+When a form includes a question of type :tc:`backgroud-audio`, audio is recorded while the form is open and attached to the form submission as a single audio file. These recordings can be used for quality assurance, training, transcription, and more. Use background recording instead of an :ref:`audio question <customizing-audio-quality>` when you want to make sure to record everything that happens during form filling.
+
+Before adding background audio recording to your form, make sure that you have a plan for following local laws around audio recording. We generally recommend including a note at the beginning of the form to remind data collectors and participants that they are being recorded and to describe the purpose of the recording. Depending on the context, you may be required to ask for the consent of every person in speaking range of the microphone. <Examples>
```
seadowg, line 2427: This and the next paragraph feel like they might suit being a "tip" right? I feel like this whole section is quite long, and it'd be good to break up the opinionated/advisory stuff from the core info.

src/form-question-types.rst
```diff
@@ -2399,4 +2400,44 @@ storing the values for later use.
...
+
+It can be helpful to combine background audio recording with :doc:`audit logging <form-audit-log>` to have more context while listening to the recording. Recording starts at the ``form start`` and ``form resume`` events and stops at the ``form exit`` event. You can use the difference between the ``form start`` time and the start time of an event to identify how far into the background recording a certain event happened.
+
+There is an override available to data collectors when recording might compromise safety or when consent to record can't be obtained but it's still important to capture some data. When the audio recording option is unchecked, audio recording ends immediately and any previously-recorded audio is deleted. Recording can't be resumed during the current form-filling session. The setting persists across form filling sessions. If :doc:`audit logging <form-audit-log>` is enabled, a ``background audio disabled`` event will be logged if a data collector toggles off recording and a ``background audio enabled`` event will be logged if a data collector toggles it back on.
```
seadowg, line 2441: This could maybe live in a sub-heading ("Enumerator override" for instance).

src/form-question-types.rst
```diff
@@ -2399,4 +2400,44 @@ storing the values for later use.
...
+
+In some rare cases such as the device running out of space, the recording may complete successfully but not be attached to the form. If this happens, the recording may be available in the ``recordings`` folder of the :ref:`Collect directory <collect-directory>`. This folder is never cleared so consider emptying it yourself once you have retrieved its files.
```
seadowg, line 2443: Maybe this could be a warning? Or we could pull it out as a subsection of the recording question section and link to it here.

seadowg:
> Any screenshot needed? I've gone back and forth on adding one of the top bar.

Yeah, not sure. I think you've done a good job of explaining it here. There's not a lot of UI to give examples of, so we probably don't need them.

> I'm hesitating between attaching XLSX files and putting the form defs inline but collapsed by default. I'd love thoughts on that or other options!

Do you mean linking to an example or having the form defs inline? I think for this it'd be fine to have it inline seeing as it's just one line.

seadowg (changes requested):

lognaturel:
> seeing as it's just one line.

I'm talking about the intro screen examples. Those are going to be big. I'm now thinking they're likely more usable as XLSX.

seadowg:
@lognaturel gah sorry yeah that makes sense. Yes I agree it'd be better as an example XLSX.

lognaturel:
Thanks! I added some headings as suggested and did a little bit of reorganization that I think make things flow better.

I propose we go with this for now and file an issue to add introduction examples and make other improvements.

seadowg:
@lognaturel agreed. I think the headings really help!

seadowg (approved):

PR #1319, opened 2021-02-10
Added documentation about new external binary widgets

grzesiek2010:
closes #1318 

#### What is included in this PR?
Added documentation about new external binary widgets.

PR #1332, opened 2021-02-27
Document receiving files from external apps

### LINKED ISSUE #1318: Add documentation about new external binary widgets
We are in the middle of implementing new external binary widgets:
- external image widget
- external audio widget
- external video widget
- external arbitrary file widget

those widgets should be available in v1.30 so we need to update the documentation.

lognaturel:
Closes #1318, replaces #1319

Converts images to git LFS, removes images that don't add a lot, documents how apps can send back binary files.


src/launch-apps-from-collect.rst
```diff
@@ -107,24 +144,10 @@ XLSForm
...
 
-``ex:android.intent.action.DIAL(uri_data='tel:5555555')``
-  Launches a phone dialer with the number ``5555555`` as the number to dial.
+Since Collect v1.30.0, it is possible to populate binary fields (:ref:`image <external-app-image-widget>`, :ref:`video <external-app-video-widget>`, :ref:`audio <external-app-audio-widget>` or :ref:`file <external-app-file-widget>`) in field lists. An app that returns one or more binary files to Collect as part of a field list must provide a `content URI <https://developer.android.com/guide/topics/providers/content-provider-basics#ContentURIs>`_ as the value for each ``extra`` that corresponds to a binary question to populate. Additionally, the external application must specify `ClipData items <https://developer.android.com/reference/android/content/ClipData.Item>`_ for each URI it returns. Your app can then grant Collect temporary permissions to the files using the `Intent.FLAG_GRANT_READ_URI_PERMISSION <https://developer.android.com/reference/android/content/Intent#FLAG_GRANT_READ_URI_PERMISSION>`_ intent flag. See :ref:`a similar code example above <external-app-design>`.
```
grzesiek2010, line 151: Links here are to external widgets but in such an intent group we can use normal binary widgets and also normal binary widgets are used as a sample in the table above so it might be a bit misleading.

grzesiek2010, line 151: Links here are to external widgets but in such an intent group we can use normal binary widgets and also normal binary widgets are used as a sample in the table above so it might be a bit misleading.

grzesiek2010 (approved):
Just one thing but generally it looks fine.

src/launch-apps-from-collect.rst
lognaturel: Oh yes, thank you!

PR #1338, opened 2021-04-06
Improve Document

john7325-1925487:
closes #

addresses #

As a new user, I found the ODK-X document to be quite confusing during the setup process. I think a great way to tackle this is by adding pictures or screenshots in certain areas to guide us through the process. For example, the portion of the document that went over the setting up of the adb.


lognaturel:
Thanks for getting involved, @john7325-1925487! The ODK-X documentation is now at https://github.com/odk-x/docs.

PR #1347, opened 2021-05-07
Provide general introduction to XLSForm, add starter template

### LINKED ISSUE #1261: Document the XLSForm settings sheet
- [x] An explicit `form_id` should always be set
- [x] An explicit `form_title` should always be set
- [x] An explicit `version` should always be set
- [x] `instance_name` can be really useful if a specific record ever needs to be identified again on the device
  - [x] link to [instance map](https://docs.getodk.org/collect-form-map/#form-instance-map-markers)
  - [ ] link to Central docs on submission page

lognaturel:
Closes #1261 

The high-level goal of this PR is to document the `settings` worksheet in XLSForm. The immediate need this satisfies is to have something useful to link to from [Central submission details page docs](https://github.com/getodk/docs/pull/1346#discussion_r625222771).

I also took the opportunity to take care of various things we've been wanting to do for some time:
- I linked to an XLSForm template. I'm pretty confident in the columns I included but if it seems controversial, we could pull it for now and have some more discussion. I don't think we want to change the column order once we publish.
- Cut out a bunch of things from the encryption page. It needs to be written but this is a quick move towards that.
- Add a high-level introduction to forms that is aimed at project designers, not developers.

### DIFF
--- a/src/collect-settings.rst
+++ b/src/collect-settings.rst
@@ -224,6 +224,8 @@ Specifies how blank forms should be updated:
   When enabled, forms are sent immediately when they are finalized, if the device can connect to the internet. If an internet connection is not available at the time of finalization, your finalized forms will be queued to send as soon as connectivity is established.
   You can specify whether to send over WiFi, cellular data, or both.
 
+.. delete-after-send:
+
 :guilabel:`Delete after send`
 """"""""""""""""""""""""""""""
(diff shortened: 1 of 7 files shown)


yanokwa:
I've put my improvements at https://github.com/yanokwa/odk-docs/tree/1347-yaw

PR #1364, opened 2021-06-14
Fix audio quality parameter example

### LINKED ISSUE #1363: Incorrect use of audio parameters in example
The ["survey" table](https://docs.getodk.org/form-question-types/#id110) under the heading "Changing audio quality during form entry" shows an invalid method of specifying the audio quality.    

incorrect | correct
--- | ---
quality='voice-only' | quality=voice-only
quality='normal' | quality=normal

If the quotes are included, an error with the message of `Invalid value for quality.` is thrown.

joshuaberetta:
closes #1363

Removed quotes from audio quality parameters resulting in invalid survey.


lognaturel (approved):
Thank you!

PR #1382, opened 2021-08-11
Update collect-forms.rst

yanokwa:
Sorry, but this bothers me.

### DIFF
--- a/src/collect-forms.rst
+++ b/src/collect-forms.rst
@@ -17,7 +17,7 @@ In order to fill out forms with survey participants, you must first load blank f
 
 .. _in-app-get-blank-forms:
 
-Loading Forms from a server
+Loading forms from a server
 -----------------------------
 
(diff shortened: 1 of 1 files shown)


src/collect-forms.rst
```diff
@@ -17,7 +17,7 @@ In order to fill out forms with survey participants, you must first load blank f
 
 .. _in-app-get-blank-forms:
 
-Loading Forms from a server
+Loading forms from a server
```
seadowg, line 20: One quick thought: should we actually always be capitalizing ODK concepts (forms, projects, instances etc) to highlight that we're not just using the normal word? This is something that seems to be really inconsistently done in the docs, but we could just add a decision about it (arbitrary or not) to the style guide?

yanokwa: Sigh. Yes, we should capitalize and be consistent.

yanokwa:
Per the review feedback, I've closed this PR, opened a new one at https://github.com/getodk/docs/pull/1384, and filed an issue at https://github.com/getodk/docs/issues/1383.

PR #1397, opened 2021-09-21
Default to built-in audio recorder, remove RecForge recommendation

yanokwa:
We have a customer who has had crashing issues with RecForge. See https://forum.getodk.org/t/issues-with-audio-recording/35116/3 for more. 

I have confirmed that RecForge crashes on v2021.2.4. It's a very ad-heavy app so rather than track down the issue, I think we should remove it. I have confirm that axet's call record works great on my Pixel 3a running Android 11.

And while I was there, I think we should recommend people use the built-in audio recorder, so I've moved some of the language around. 

Things to consider for review:
- I'm not sure why we didn't do this originally. Maybe our audio recorder is terrible for older versions of Android? I don'
- I tried with a few other apps (Smart Recorder, Audio Recorder - no ads) but none of them showed up on the dialog Collect presents to choose a recorder.
- How quality interacts with the external app recording setting in Collect can be tricky. Make sure I've described that correctly!

### DIFF
--- a/src/form-question-types.rst
+++ b/src/form-question-types.rst
@@ -1667,7 +1667,7 @@ In the parameters column, write ``max-pixels=`` followed by the desired maximum 
 
 .. _audio:
 
-Audio widget
+Audio widgets
 ----------------
 
 .. contents::
@@ -1681,7 +1681,11 @@ type
 appearance
   :tc:`none`
 
-Records audio using the device's microphone or a connected external microphone. By default, an :ref:`external application <external-audio-app>` is used. Starting with Collect v1.29, you can also use :ref:`built-in recording <built-in-audio-recording>`.
+Records audio using the device's microphone or a connected external microphone. By default, an :ref:`external application <external-audio-app>` is used. 
+
+.. tip::
+
+  Instead of the default audio widget, we recommend you use the :ref:`built-in audio recorder <built-in-audio-recording>` because you can customize audio quality and record while filling out other questions. Built-in recording is available in Collect v1.29 or later.
 
 .. image:: /img/form-widgets/audio-start.*
   :alt: The Audio form widget as displayed in the ODK Collect App on an Android phone. The question text is "What does it sound like?" There are two buttons: Record Sound and Choose Sound.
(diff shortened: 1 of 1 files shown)


lognaturel:
Typically we describe whatever the default behavior is first. I think maybe instead of changing the order we should consider having an introduction that says we typically recommend built-in recording and describes situations when external apps may be desired (external mic, more control over file type and compression). 

Perhaps we should also consider switching the Collect default. 

I’m ok going with this if you still prefer it over what I’ve described.

seadowg (commented):
> Perhaps we should also consider switching the Collect default.

We definitely talked about doing this eventually, but looking at analytics it does look like we're still seeing more usage of external than internal at the moment in terms of inclusion in forms - in the last 30 days we've seen [67 forms use internal](https://console.firebase.google.com/u/0/project/api-project-322300403941/analytics/app/android:org.odk.collect.android/events/~2Freport~2FAudioRecordingInternal%3Ft%3D1632329938491&fpn%3D322300403941&swu%3D1&sgu%3D1&sus%3Dupgraded&params%3D_u..pageSize%253D25%2526_r..eventId%253DAudioRecordingInternal%2526_u.dateOption%253Dlast30Days&cs%3Dapp.m.events.detail&r%3Ddefault&g%3D1) and [148 use external](https://console.firebase.google.com/u/0/project/api-project-322300403941/analytics/app/android:org.odk.collect.android/events/~2Freport~2FAudioRecordingExternal%3Ft%3D1632329928182&fpn%3D322300403941&swu%3D1&sgu%3D1&sus%3Dupgraded&params%3D_u..pageSize%253D25%2526_r..eventId%253DAudioRecordingExternal%2526_u.dateOption%253Dlast30Days&cs%3Dapp.m.events.detail&r%3Ddefault&g%3D1). We don't really have insight on how many of the forms/enumerators using external actually need it, and it's also not a huge amount of forms in play. I think we can probably do it, but we should give people a heads-up. 

For the moment I'd advocate for keeping the order of the docs here as it is, removing the RecForge recommendation, adding a tip that we recommend switching to internal as @lognaturel suggests and then announcing that we're going to change the default in v2021.4.

yanokwa:
We worked the PR. Changes...

- Title should be audio widgets since there are multiple-ish
- Moved information about audio-recording into a tip so it's more obvious. Also point out the value of using it.
- Added syntax for built-in so people can link directly to this section.
- Added a form design example for the built-in recorder to make it easy for folks to copy and paste and to link directly to.
- Added syntax for external so people can link directly to this section (and as we deprecate the default, this will be helpful).
- Moved Customizing audio quality down the hierarchy under built-in recorder.

I'd love another review, @seadowg and @lognaturel.

src/form-question-types.rst
```diff
@@ -1681,7 +1681,11 @@ type
 appearance
   :tc:`none`
 
-Records audio using the device's microphone or a connected external microphone. By default, an :ref:`external application <external-audio-app>` is used. Starting with Collect v1.29, you can also use :ref:`built-in recording <built-in-audio-recording>`.
+Records audio using the device's microphone or a connected external microphone. By default, an :ref:`external application <external-audio-app>` is used. 
+
+.. tip::
+
+  Instead of the default audio widget, we recommend you use the :ref:`built-in audio recorder <built-in-audio-recording>` because you can customize audio quality and record while filling out other questions or using another app. Built-in recording is available in Collect v1.29 or later.
```
seadowg, line 1688: Just as a side note: this messes with my mental model a bit. I'd think of there being one audio widget that can behave in different ways, rather than there being multiple audio widgets. It does look like this way of describing it is consistent with the geo widgets though, so I think that's on me.

seadowg: I think we can drop the "or using another app" here. That'd be possible with an external recorder as well.

seadowg (approved):
Looks great! Just one wee tweak. This should be pretty simple to adapt when we change the default behaviour as well.

src/form-question-types.rst
yanokwa: Agree that widgets are a weird term! We were young and foolish back then. File an issue with your suggested terminology. Field types? Entry elements? ODK experiences?

Issue #1365, opened 2021-06-16
Update documentation for Collect projects and first launch (v2021.2)


lognaturel:
- [x] Project name/icon/color can be configured from admin settings
- [x] Revise collect-settings
- [x] Demo project can be reconfigured or deleted
- [x] App version is at the bottom of the screen
- [x] First launch screen appears when there are no projects
- [x] QR code can be scanned or imported from image (useful when all staff are remote)
- [x] Icon and color for the current project are at top right and tapping it makes it possible to configure the current project or switch projects
- [x] App bar title is the project name
- [x] QR code has new `projects` object with new keys [here](https://github.com/getodk/collect/pull/4607/files#diff-e07bc4a275dc23c5fc4f81390c4498debea636dbc1d9c9158861d45ecf78048e)
- [x] Users are prompted to confirm when they enter in project connection details that match existing project(s) (getodk/collect#4697)

seadowg:
Looking through the docs there is **A LOT** that needs to be revised. @lognaturel's list here is a good way to check that we've covered everything, but my process here is just to run throught the Collect docs in order and try and update/change things for projects.

lognaturel:
I'm working on the last few items here.

PR #1406, opened 2021-10-20
Update QR code docs

seadowg:
Work towards #1365

Initial pass at re-writing/structuring the QR docs. I've tried to move away from the focus on the QR code screen itself and instead describe settings QR codes and then try and cover all the different use cases for them. There aren't currently any screenshots, but I couldn't convince myself of places they might be needed - would be good to get feedback on places they might help.


src/collect-import-export.rst
```diff
@@ -1,58 +1,56 @@
...
-Importing settings from an image saved on your device
-------------------------------------------------------
-
-You can import settings from a QR code saved on your device by clicking on the :fa:`ellipsis-v` icon which is located to the right of the share icon. Then select the :guilabel:`Import QRCode` option.
-
 Making your own QR code
 ---------------------------
 
-The contents of a settings QR code is a JSON object with two objects for general and admin settings. The JSON structure is the following:
+The contents of a settings QR code is a JSON object with objects for unprotected ("general") and protected ("admin"), as well as the project details themselves. The JSON structure is the following:
```
lognaturel, line 38: the word "settings" is missing before the first comma.

lognaturel (changes requested):
Sorry, I completely missed this! Looks good, just one word missing.

lognaturel (approved):

PR #1412, opened 2021-11-03
or_other should not be used due no language or choice_filter support

yanokwa:
or_other doesn't have [language support](https://github.com/XLSForm/pyxform/issues/218) or [choice_filter](https://github.com/XLSForm/pyxform/issues/17) support and people should know that. We should probably deprecate it at some point.


lognaturel (approved):
Feels odd to me to document something we don't recommend using but I guess some forms already have it so best to give folks who come across it an explanation.

PR #1420, opened 2022-01-08
Fix pdf links

### LINKED ISSUE #1417: Link to download "Contributing to ODK Docs" documentation as PDF fails
At the bottom of https://docs.getodk.org/contributing/, there is a link to "Download this documentation as a PDF." When the link is clicked, no file is downloaded.

jennifermklein:
closes #1417

I noticed that the link on the [Contributing to ODK Docs](https://docs.getodk.org/contributing/) page to Download this documentation as a PDF did not work and would yield a "Failed - No file" message in the download bar. I updated the link in the footer.html file to include the entire link an incorrect path isn't inadvertently used.

Confirm correct link is being used and correct document is being accessed.

### DIFF
--- a/src/_templates/footer.html
+++ b/src/_templates/footer.html
@@ -11,7 +11,7 @@
   {% endif %}
 
   <hr/>
-<p class="footerText"><a href="_downloads/ODK-Documentation.pdf" download>Download this documentation as a PDF.</a></p>
+<p class="footerText"><a href="https://docs.getodk.org/_downloads/ODK-Documentation.pdf" download>Download this documentation as a PDF.</a></p>
 <p class="footerText">If you still need help, you can ask support questions in the <a href="https://forum.getodk.org/">ODK Forum</a>.</p>
   <div role="contentinfo">
     <p>


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3
+FROM python:3.9.7
```
lognaturel, line 1: Thanks for separating out changes into little pull requests! Now that the version change has been merged, you can `git rebase upstream/master` and then `git push --force` and this commit should no longer be visible here.

src/_templates/footer.html
```diff
@@ -11,7 +11,7 @@
   {% endif %}
 
   <hr/>
-<p class="footerText"><a href="_downloads/ODK-Documentation.pdf" download>Download this documentation as a PDF.</a></p>
+<p class="footerText"><a href="https://docs.getodk.org/_downloads/ODK-Documentation.pdf" download>Download this documentation as a PDF.</a></p>
```
lognaturel, line 14: This works for the published version! But it will give a false sense of the world with a set of docs hosted somewhere else or locally. Ideally this could dynamically use the base url.

yanokwa:
I think it's better we flip things around here. That is, merge the fix and file an issue to reduce the use of absolute paths. I'll do that now..

Aviii24x7:
Didn't get the answer~!!

PR #1426, opened 2022-02-07
Hide Aggregate from sidebar because it is end-of-life

yanokwa:
Aggregate has reached end-of-life so we should remove it from the side-bar and save some space. 

With this PR, you don't see the Aggregate content unless you know the link. 
Screen Shot 2022-02-07 at 10 36 16

And when you link to it, then you see some of the sidebar hierarchy. 
Screen Shot 2022-02-07 at 10 36 46

We should eventually remove Aggregate content entirely from the docs, but this seems like a reasonable step.


lognaturel (approved):

yanokwa:
Note that I do not use `exclude_patterns` to exclude `aggregate-*.rst` in `src/conf.py` because then every internal link that references Aggregate docs will break.

PR #1427, opened 2022-02-07
Add up-to-date training resources

yanokwa:
My initial goal is to provide information that is ODK-centric (e.g., not Aggregate or Survey123 or KoBo) and is up-to-date. I think it's important to summarize each link and provide the media type and date and so I've done that.

Screen Shot 2022-02-07 at 12 06 08

I cheat a bit with xlsforms.org date, but I think I can be forgiven.

I'm very open to changing any of this.

### DIFF
--- a/src/training.rst
+++ b/src/training.rst
@@ -1,0 +1,18 @@
+Training Resources
+==================
+
+Below are resources for learning more about ODK. Email support@getodk.org if you'd like to add a link to this list.
+
+.. warning::
+
+ ODK has frequent improvements and so the resources below quickly fall out of date. We recommend you start with the resources that are most recent and check them against ODK's authoritative and up-to-date docs.
+
+#. `XLSForm form building guide <https://xlsform.org/>`_. Great as a tutorial or reference guide. Updated Jan 2022.
+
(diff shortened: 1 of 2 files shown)


src/training.rst
```diff
@@ -0,0 +1,20 @@
+.. |date| date:: %b %Y
+
+Training Resources
+==================
+
+Below are resources for learning more about ODK. Email support@getodk.org if you'd like to add a link to this list.
+
+.. warning::
+
+ ODK has frequent improvements and so these resources quickly fall out of date. We recommend you start the resources that are most recent.
```
lognaturel, line 10: I think you should add something that's like "The docs here are authoritative. You can use these resources to get a different perspective but check them against the docs"

"We recommend you start the resources that are most recent." probably missing a with

src/training.rst
```diff
@@ -0,0 +1,20 @@
...
+Training Resources
+==================
+
+Below are resources for learning more about ODK. Email support@getodk.org if you'd like to add a link to this list.
+
+.. warning::
+
+ ODK has frequent improvements and so the resources below quickly fall out of date. We recommend you start with the resources that are most recent and check them against ODK's authoritative and up-to-date docs.
+
+#. `XLSForm form building guide <https://xlsform.org/>`_ from the ODK team. Frequently updated and great as a tutorial or reference guide. Updated |date|.
```
lognaturel, line 12: “From the ODK team” doesn’t seem right. It only sees light updates.

lognaturel (approved):

PR #1444, opened 2022-03-09
Add tip about increasing Node memory

yanokwa:


lognaturel (approved):
This is not really just about upgrades, is it? Can’t this help with out of memory issues with exports?

yanokwa:
Reworked all the things to make increasing memory allocation a more generic thing
Screen Shot 2022-03-08 at 23 13 59

And once we have that, we can point to that in the troubleshooting section.
Screen Shot 2022-03-08 at 23 13 42

lognaturel (approved):

PR #1451, opened 2022-03-24
Skip spellchecking on Apple Silicon because enchant requires Intel chip

yanokwa:
enchant requires an Intel chip (https://pyenchant.github.io/pyenchant/install.html?highlight=silicon#on-macos) and I'd rather not install the Intel version of brew just for this. I also don't love using the Docker build.

This PR skips spellcheck when on Apple Silicon.


lognaturel (approved):
Fine by me. Spellcheck uses the local dict which is not the same everywhere, anyway, so CI has to have the last word.

PR #1454, opened 2022-04-06
Deprecate Briefcase

yanokwa:
Same thing we did with Aggregate
* Add a message to the top of all pages.
* Remove from table of contents
* Add orphan tag so build completes

All this ensures that the pages are hidden, but links work. There will still need to be some clean up of mentions of Briefcase and Aggregate.

### DIFF
--- a/src/briefcase-api.rst
+++ b/src/briefcase-api.rst
@@ -1,5 +1,10 @@
+:orphan:
+
 Briefcase Aggregate API
 =========================
+
+.. warning::
+  ODK Briefcase is no longer being updated. If you're using Briefcase for CSV exports, data decryption, or automation, use :doc:`ODK Central <central-intro>` instead.
 
 .. _introduction:
 
--- a/src/briefcase-install.rst
+++ b/src/briefcase-install.rst
@@ -1,5 +1,10 @@
+:orphan:
+
 Setting Up ODK Briefcase
 ===================================
+
+.. warning::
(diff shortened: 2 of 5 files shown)


src/briefcase-api.rst
```diff
@@ -1,6 +1,11 @@
+:orphan:
+
 Briefcase Aggregate API
 =========================
 
+.. warning::
+  ODK Briefcase is no longer being updated. Please use :doc:`ODK Central <central-intro>` instead.
```
lognaturel, line 7: Recommending Central feels odd. It makes sense for folks who were using Aggregate + Briefcase but not do folks running entirely offline. I don’t have a great recommendation yet but wanted to point that out.

lognaturel (approved):

src/briefcase-api.rst
libcthorne: @yanokwa Is there an alternative to the ODK Briefcase CLI out there? We are using this to automate pulling and decryption of forms

yanokwa: What backend are you using to pull the data from?

libcthorne: @yanokwa We're using ona.io which I believe provides the same interface as ODK Aggregate (https://github.com/onaio/onadata/blob/4db58e41494f0addc7ea0e97f6c7b8dac78cd2b5/setup.py#L7)

yanokwa: Briefcase has not had an update for 2 years and it's still working. We don't have any reason to believe it will stop working. The only real change here is that we are being explicit that it won't ever get an update.

If you are using Ona, talk to their support team about what they recommend. My guess is that they have an API that you can use.

PR #1524, opened 2022-12-06
Tweak datasets structure

yanokwa:
I think quick start shouldn't have a long title because it's nicer as a link.

Testing seems more like a sub-section to datasets and entities in v2022.3

Sections after building sub-sections seemed like they should be sub-sections, so I've done that.

I've also tried to standardize case for Datasets and Entities in titles. Form maybe should be title case, but I don't feel strongly about it.

We could probably drop "Creating forms to use Datasets and Entities" as a section and bring "Building a form that creates" and "Building a form that uses" up a section.

### DIFF
--- a/docs/central-datasets.rst
+++ b/docs/central-datasets.rst
@@ -108,9 +108,7 @@ Writing forms to use Central Datasets and Entities
 
 Central does a lot of work to help you manage Entity data, but at least for now the only way to ask it to do so is to create a form that describes how.
 
-In the following section, we describe how to author forms that create new Entities, also known as registration forms.
-
-After that, we discuss forms that use Entities, also known as follow-up forms.
+In the following section, we describe how to author forms that create new Entities, also known as registration forms. After that, we discuss forms that use Entities, also known as follow-up forms.
 
 .. _central-datasets-registration-forms:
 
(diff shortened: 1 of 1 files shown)


docs/central-datasets.rst
```diff
@@ -201,5 +199,5 @@ Your ``trees`` Dataset can now be attached to any Form using ``select_one_from_f
 
 You can see the above example as a working spreadsheet `here <https://docs.google.com/spreadsheets/d/12oJZDpJ8RxtmNopfqNKp3RWMsf4O3MWACYOTub_yZaQ/edit#gid=0>`_.
 
-The same Dataset can be used in many different Forms. The concepts and patterns described in the data collector workflows documentation (https://docs.getodk.org/data-collector-workflows) and the datasets documentation (https://docs.getodk.org/form-datasets/) apply to server-managed Datasets as well.
+The same Dataset can be used in many different Forms. The concepts and patterns described in the `data collector workflows documentation <https://docs.getodk.org/data-collector-workflows>`_ and the `datasets documentation <https://docs.getodk.org/form-datasets>`_ apply to server-managed Datasets as well.
```
lognaturel, line 202: Use links to the sections instead

lognaturel (approved):

PR #1525, opened 2022-12-08
There's only form submissions. Also this is shorter.

yanokwa:


lognaturel:
This would break any links to the section and doesn't feel worth it to me.

lognaturel:
@yanokwa points out that this is top-level so it doesn't matter.

lognaturel (approved):

PR #1533, opened 2022-12-20
Add direct link to export options

lognaturel:
Updates the section on submission exports to match new language. Adds a section header so we can directly link to export options. Crosslinks with form updates.


yanokwa (approved):
LGTM. Read over the content and confirmed it rendered.

PR #1540, opened 2023-01-11
Document upstream SSL

yanokwa:


yanokwa:
@delcroip Does this documentation sufficiently cover the configuration you have?

lognaturel (approved):
I haven't tried this but it seems reasonable. Let's merge it and see if we get questions/feedback.

PR #1544, opened 2023-01-13
Rename central-datasets to central-entities

yanokwa:


matthew-white:
This might be on the radar already, but since we link to central-datasets from Frontend, we would need to change central-datasets to central-entities in Frontend as well.

s3_website.yml
```diff
@@ -31,6 +31,7 @@ redirects:
   briefcase-guide/index.html: briefcase-intro
   briefcase-vs-aggregate/index.html: briefcase-intro
   briefcase-and-aggregate/index.html: briefcase-intro
+  central-datasets/index.html: central-entities
```
yanokwa, line 34: @matthew-white Does this redirect address your concern?

matthew-white: Ah I didn't notice that, excellent. ✨

lognaturel (approved):

PR #1547, opened 2023-01-19
Remove warning that is not or no longer true

lognaturel:
The error that the warning mentions definitely does not occur. There's a small chance that this is a change in behavior but I'm not sure.

One surprising thing about that example is that if a is 2 and b is blank, a_plus_b is blank. It might be expected to be 2 (and in fact, that might be the expected W3C XForms behavior, I'm not sure, JR's type coercion is not great).

### DIFF
--- a/docs/form-logic.rst
+++ b/docs/form-logic.rst
@@ -647,39 +647,6 @@ Complex example
 
 .. include:: incl/form-examples/constraint-on-selected.rst
 
-.. warning::
-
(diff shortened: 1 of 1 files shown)


yanokwa (approved):
I was shown this test in Collect and the docs aren't correct. Hélène also tested in Enketo and confirms both give NaN unless both values are provided.

I've also confirmed that https://docs.getodk.org/form-logic/#empty-values-in-math exists and warns people about adding things to null values.

PR #1548, opened 2023-01-20
Remove media:: prefix for image, audio, video

lognaturel:
The `media::` prefix is really just a `pyxform` implementation detail. It leads to issues like https://github.com/XLSForm/pyxform/issues/601 so better to remove, especially since https://xlsform.org/en/#media doesn't use it.

I did `grep -r "media::" .` from the root and verified that no `rst` docs have it.

### DIFF
--- a/docs/form-language.rst
+++ b/docs/form-language.rst
@@ -8,7 +8,9 @@ add columns of user-facing content with language-specific columns. All columns r
 
  - ``label``
  - ``hint``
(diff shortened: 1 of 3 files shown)


yanokwa (approved):
Confirmed that no more mentions of "media::" exist in the docs. Also confirmed it's not used in the form template in the docs or in all widgets.

PR #1552, opened 2023-01-26
Document styling markers

### LINKED ISSUE #1488: Document styling markers in select one from map widget
In ODK collect we are working on styling markers displayed on a map in [select one from map widget](https://docs.getodk.org/form-question-types/#select-one-from-map-widget). It will be possible to specify color and symbol: https://github.com/getodk/collect/pull/5280
This probably should be mentioned in the docs: https://docs.getodk.org/form-question-types/#select-one-from-map-widget

grzesiek2010:
closes #1488

It adds a note in https://docs.getodk.org/form-datasets/#building-selects-from-geojson-files which looks like this:
image

### DIFF
--- a/docs/form-datasets.rst
+++ b/docs/form-datasets.rst
@@ -105,6 +105,12 @@ GeoJSON files referenced in forms can have any number of ``features`` and any nu
 
 All properties are displayed by :ref:`select from map <select-from-map>` questions and can be :ref:`referenced by any part of the form <referencing-values-in-datasets>`.
 
+.. note::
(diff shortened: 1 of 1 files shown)


grzesiek2010:
Maybe we should also update the screenshot from https://docs.getodk.org/form-question-types/#select-one-from-map-widget so that it's well visible that styling markers is supported... what do you think?

lognaturel (approved):

lognaturel:
I don't feel strongly about adding/updating a screenshot! I think this is a great note so I've merged it and you can see if you want to add/change the screenshot.

Issue #1381, opened 2021-08-09
Central can't always be downgraded after upgrade


matthew-white:
We've seen that it's not always possible to downgrade Central after upgrading. That means that if you have trouble upgrading, it's not guaranteed that you'll be able to downgrade to a version that did work. I'm not sure there's a way to change this (certainly not an easy way). However, I think it'd be useful to document this fact (maybe at https://docs.getodk.org/central-upgrade/), because it can be surprising. I vaguely remember discussing this issue before, but I don't think I see anything about it in the docs or in the issues here.

This issue comes up for Backend and for Frontend.

For Backend, I think the main issue is database migrations. I don't believe we expose a way through Docker to roll back database migrations. Perhaps part of the reason why is that it's not always possible to undo a migration: some migrations are partially destructive. There are also some migrations that technically could be undone, but we haven't written the code to do so, usually because doing so would be fairly complex.

For Frontend, I think the main issue is local storage. A new version of Frontend may set an item in local storage that previous versions did not. If you upgrade to the new version, then take an action that sets the item, then downgrade, then take an action that in the new version would set the item, then upgrade again, the item might not reflect the actual state, causing issues in the new version. (This came up when we were working on v1.2, which sets `sessionExpires` in local storage: if you upgrade to v1.2 and log in; downgrade to v1.1 and log out; then upgrade to v1.2 again; then you won't be able to log in, because v1.2 thinks you're logged in even though you're not.)

PR #1557, opened 2023-01-31
Update Central upgrade instruction structure to make commands copiable + draw attention to relevant version upgrade notes

lognaturel:
Goals:
- Make it easy to copy shell commands
- Improve the experience for people who upgrade every time
- Make it clear that instructions for every version between the current and target upgrade version are relevant
- Match https://docs.getodk.org/central-troubleshooting/#database-reset-after-running-docker-command which @yanokwa recently restructured

Todo:
- Add general instructions for performing maintenance: make a backup, announce a maintenance period, etc

@yanokwa, @alxndrsn, I'm interested in your early thoughts. I think this structure will make it easier to make the postgres upgrade instructions visible.

### DIFF
--- a/docs/central-upgrade.rst
+++ b/docs/central-upgrade.rst
@@ -3,52 +3,90 @@
 Upgrading Central
 =================
 
-We release new versions of Central regularly. You do not have to upgrade to the latest version immediately, but we generally recommend that you do so to get access to the newest features, bug fixes, and security updates.
+We release new versions of Central regularly. We generally recommend that you upgrade as soon as you can to get access to the newest features, bug fixes, and security updates. It is not generally possible to downgrade and we do not guarantee that versions prior to the latest release can still be installed.
 
-.. admonition:: Note
+Start by reviewing upgrade notes for all versions between your current version and the one you are upgrading to.
 
-  To see your version number, click on the question mark icon in the upper right section of your Central menu bar, then click :guilabel:`Version`. If you don't see the question mark, you can also see the version number by adding ``version.txt`` to the root URL (e.g., `demo.getodk.cloud/version.txt <https://demo.getodk.cloud/version.txt>`_).
+* Central v1.5: no special instructions.
+* :ref:`Central v1.4 <central-upgrade-1.4>`: see notes if you have tens of thousands of submissions.
+* :ref:`Central v1.3 <central-upgrade-1.3>`: see notes to ensure you have the correct version of ``docker-compose``.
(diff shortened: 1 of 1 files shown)


alxndrsn:
The scope of this PR seems beyond what's described.  Some of the changes look 100% acceptable and could be merged immediately.

Making instructions copyable sounds like a great idea!

lognaturel:
@matthew-white this intersects with docs you're writing so you may want to take a quick look to see whether you like the proposed structure.

docs/central-upgrade.rst
```diff
@@ -3,52 +3,90 @@
...
 =================
 
-We release new versions of Central regularly. You do not have to upgrade to the latest version immediately, but we generally recommend that you do so to get access to the newest features, bug fixes, and security updates.
+We release new versions of Central regularly. We generally recommend that you upgrade as soon as you can to get access to the newest features, bug fixes, and security updates. It is not generally possible to downgrade and we do not guarantee that versions prior to the latest release can still be installed.
 
-.. admonition:: Note
+Start by reviewing upgrade notes for all versions between your current version and the one you are upgrading to.
 
-  To see your version number, click on the question mark icon in the upper right section of your Central menu bar, then click :guilabel:`Version`. If you don't see the question mark, you can also see the version number by adding ``version.txt`` to the root URL (e.g., `demo.getodk.cloud/version.txt <https://demo.getodk.cloud/version.txt>`_).
+* Central v1.5: no special instructions.
```
matthew-white, line 10: Why mention v1.5 in particular?

matthew-white (commented):
The new structure looks great to me! I like that we're making it very clear when there are version-specific instructions. Also, I think this PR closes #1381.

docs/central-upgrade.rst
yanokwa: People on the forum sometimes ask if there are upgrade instructions for a version if we don't explicitly document that there are none.

matthew-white: That makes sense to me. I guess I'm wondering why we'd mention v1.5 and not other versions, for example, v2022.3 or v1.1. Could we instead add a generic note along the lines of "Most versions of Central do not have special upgrade instructions"?

matthew-white: (Also, definitely don't want to block merging / no strong feelings, mostly just wanted to throw out the thought.)

lognaturel:
Thanks, all. Don't hesitate to make edits in follow-on PRs.

PR #1560, opened 2023-02-01
Update upgrade instructions for v1.5.3 and v2022.3

### LINKED ISSUE #1503: Add upgrade instructions for v1.5.3
If a user has [disabled or customized Sentry](https://docs.getodk.org/central-install-digital-ocean/#disabling-or-customizing-sentry), then they will need to follow special upgrade instructions in order to upgrade to v1.5.3 or later. Those instructions are listed in the [forum post](https://forum.getodk.org/t/odk-central-v1-5/37882/4), but we should add them to the docs.

matthew-white:
Closes #1503.

### DIFF
--- a/docs/central-upgrade.rst
+++ b/docs/central-upgrade.rst
@@ -7,7 +7,9 @@ We release new versions of Central regularly. We recommend that you upgrade as s
 
 Start by reviewing upgrade notes for all versions between your current version and the one you are upgrading to. You can skip upgrades and directly install the latest version as long as you make sure to follow all relevant upgrade instructions.
 
-* Central v1.5, v2022.3: no upgrade notes
+* :ref:`Central v2022.3 <central-upgrade-2022.3>`: update your NGINX configuration if you have disabled or customized Sentry
+* :ref:`Central v1.5.3 <central-upgrade-1.5.3>`: fix errors with ``git pull`` if you have disabled or customized Sentry
+* Central v1.5.0: no upgrade notes
 * :ref:`Central v1.4 <central-upgrade-1.4>`: optionally add memory if you have tens of thousands of submissions
 * :ref:`Central v1.3 <central-upgrade-1.3>`: ensure you have the correct version of ``docker-compose``
 * :ref:`Central v1.2 <central-upgrade-1.2>`: fix errors with ``git pull``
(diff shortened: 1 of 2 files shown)


docs/central-upgrade.rst
```diff
@@ -7,7 +7,9 @@ We release new versions of Central regularly. We recommend that you upgrade as s
 
 Start by reviewing upgrade notes for all versions between your current version and the one you are upgrading to. You can skip upgrades and directly install the latest version as long as you make sure to follow all relevant upgrade instructions.
 
-* Central v1.5, v2022.3: no upgrade notes
+* :ref:`Central v2022.3 <central-upgrade-2022.3>`: update your NGINX configuration if you have disabled or customized Sentry
+* :ref:`Central v1.5.3 <central-upgrade-1.5.3>`: fix errors with ``git pull`` if you have disabled or customized Sentry
+* Central v1.5.0: no upgrade notes
```
matthew-white, line 12: It seems potentially confusing to mention patch versions like v1.5.0 here but nowhere else. But I'm not sure how else to approach it. I don't think we want to list every patch version without upgrade notes.

docs/central-upgrade.rst
```diff
@@ -74,6 +77,46 @@ You'll be asked to confirm the removal of all dangling images. Agree by typing t
...
+.. _central-upgrade-2022.3:
+
+Upgrading to Central v2022.3
+-----------------------------
+
+In v2022.3, we added Content Security Policy reporting. If you have disabled or customized Sentry, then you will need to modify ``files/nginx/odk.conf.template``. See our documentation about :ref:`configuring Sentry <central-install-digital-ocean-sentry>` to learn more about the specific changes that you need to make.
+
+.. _central-upgrade-1.5.3:
+
+Upgrading to Central v1.5.3
```
matthew-white, line 89: This section was adapted from the forum post: https://forum.getodk.org/t/odk-central-v1-5/37882/4

docs/central-upgrade.rst
lognaturel: I think we should have a single section per version and include special instructions for patches there. Once the patches are out, it wouldn't make sense for someone to upgrade to the .0. And in fact it's not easy to do so. I'll follow up with a proposal.

lognaturel (approved):

Issue #1561, opened 2023-02-02
Update section about Central web user roles


matthew-white:
The section about Central web user roles lists the actions that each role allows. However, I think some actions are missing. For example:

- Set up usage information reporting
- Actions related to datasets. Most importantly perhaps, Data Collectors can only read a dataset's entities indirectly, that is, only if the dataset is attached to a form.

PR #1563, opened 2023-02-02
Update docs after sunset of Google Drive backups

### LINKED ISSUE #1535: Remove Google Drive backups
See getodk/central#323. I think we should probably continue to document how to restore from a backup .zip file, since that functionality isn't going away. However, we should otherwise remove docs about Google Drive backups.

matthew-white:
Closes #1535.


docs/central-backup.rst
```diff
@@ -7,94 +7,43 @@ Having a data backup strategy is a critical part of running a web service like O
...
 
-If you don't already have a full system backup in place and don't want to set up your own database backup, Central provides a managed backup system to Google Drive.
+If you don't already have a full system backup in place and don't want to set up your own database backup, Central provides an API endpoint to download a backup of the database.
 
-.. _central-managed-backups:
+.. _central-direct-backups:
 
-Managed backups / Direct backups via API
-----------------------------------------
+Direct Backups via API
```
matthew-white, line 16: This section remains, but I removed a couple of other sections on this page. In some cases, I added language from those removed sections to this section.

docs/central-backup.rst
```diff
@@ -7,94 +7,43 @@ Having a data backup strategy is a critical part of running a web service like O
...
 
-#. Once you have the file on your local computer, you will have to transfer that backup snapshot file to your ODK Central server. If you don't know how to do this, and you used our DigitalOcean installation guide, please see `their instructions <https://www.digitalocean.com/community/tutorials/how-to-use-sftp-to-securely-transfer-files-with-a-remote-server>`_ on how to transfer a file to a Droplet.
+1. Find the backup file on your local computer. You will have to transfer that backup snapshot to your ODK Central server. If you don't know how to do this, and you used our DigitalOcean installation guide, please see `their instructions <https://www.digitalocean.com/community/tutorials/how-to-use-sftp-to-securely-transfer-files-with-a-remote-server>`_ on how to transfer a file to a Droplet.
```
matthew-white, line 46: Quick RST question: does the first one need to be `1.`, or can it be `#.`? I tried `#.`, and it seemed fine, but I didn't want to diverge from the existing pattern.

docs/central-backup.rst
```diff
@@ -7,94 +7,43 @@ Having a data backup strategy is a critical part of running a web service like O
...
 
-  If you only use web forms for previews or for making submissions directly from Central, managed backups are sufficient. You can regenerate previews by uploading the same form with a new :ref:`form version <central-forms-updates>`. 
+  If you only use Web Forms for previews or for making Submissions directly from Central, Direct Backups are sufficient. You can regenerate previews by uploading the same Form with a new :ref:`Form version <central-forms-updates>`.
 
-ODK Central features an optional off-site backup system that backs up all Central data (but as detailed above, not Web Form configurations).
+The Central API offers an HTTP endpoint to perform an immediate backup of the database and download the result to your computer. We call this type of backup a Direct Backup and recommend it only for specific use cases. For each Direct Backup, we extract all your data (including user accounts, Forms, and Submissions), then encrypt it with an optional passphrase. The encrypted backup will be returned in the HTTP response. As detailed above, the backup will not include Web Form configurations.
```
matthew-white, line 25: "encrypt it with an optional passphrase" is probably not the most clear. I wanted to convey:

- The result is always encrypted.
- You may specify a passphrase.
- But you don't have to specify a passphrase.

My instinct is that we shouldn't delve too much into encryption and the passphrase here. What I didn't want to imply is that encryption without a passphrase provides any additional security. Users who are interested in Direct Backups will navigate to the API docs, which offer more details on these subjects.

docs/central-users.rst
```diff
@@ -80,12 +80,6 @@ By default, Central roles are configured to allow the following:
 +------------------+---------------+-----------------+----------------+----------------+
 | Delete           | x             |                 |                |                |
 +------------------+---------------+-----------------+----------------+----------------+
-| **Server Configuration**                                                             |
-+------------------+---------------+-----------------+----------------+----------------+
-| Setup Backups    | x             |                 |                |                |
```
matthew-white, line 85: Setting up Google Drive backups was removed in v1.5.3.

docs/central-users.rst
```diff
@@ -80,12 +80,6 @@ By default, Central roles are configured to allow the following:
 +------------------+---------------+-----------------+----------------+----------------+
 | Delete           | x             |                 |                |                |
 +------------------+---------------+-----------------+----------------+----------------+
-| **Server Configuration**                                                             |
```
matthew-white, line 83: I removed the entire "Server Configuration" section, but there is one server configuration action that is still available: setting up usage information reporting. I've filed the related issue #1561.

docs/central-intro.rst
```diff
@@ -33,7 +33,6 @@ Here are some of the major features we support today:
    - Connectable to analysis and dashboard applications like Excel, Power BI, or R over OData
    - Server-managed datasets for registration/follow-up workflows
  - Integrated checklist-based help system
- - Optional encrypted off-site data backups to Google Drive
```
matthew-white, line 36: We also include this line in ["What's coming in Central"](https://forum.getodk.org/t/whats-coming-in-central/19677) in the forum. @yanokwa, would you up for removing that line?

docs/central-backup.rst
```diff
@@ -7,94 +7,43 @@ Having a data backup strategy is a critical part of running a web service like O
 
 If you are an experienced system administrator, you may want to set up your own backups of the PostgreSQL database that contains all of Central's data. One strategy for doing this is to :ref:`configure a separate database server <central-install-digital-ocean-custom-db>` and back up that database server.
 
-You will additionally need to have a backup of Enketo data to be able to restore existing Web Form links. The strict minimum required to do this are Enketo's Redis store and the keys generated in the Enketo configuration. **In general, we recommend making a full system backup.**
+You will additionally need to have a backup of Enketo data to be able to restore existing Web Form links. The strict minimums required to do this are Enketo's Redis store and the keys generated in the Enketo configuration. **In general, we recommend making a full system backup.**
```
matthew-white, line 10: I thought "minimum &hellip; are" was confusing, but I also don't love "minimums &hellip; are". A rephrase might be better, something like: "At a minimum, you should back up Enketo's Redis store and the keys generated in the Enketo configuration."

docs/central-backup.rst
lognaturel: #. is better but it doesn’t deal well with nested code blocks, notes, or other directives as far as I have been able to figure out. We should use auto numbering where we can!

lognaturel (approved):

PR #1564, opened 2023-02-03
Minor upgrade updates

lognaturel:
Minor follow-up to #1560 and #1563

I don't know what "recommend it only for specific use cases" referred to. At this point I think it's clear that we prefer full machine backups and managed database backups but that this is the only other alternative.


yanokwa (approved):
LGTM. I added one commit with a tweak. If it looks good, merge.

PR #1565, opened 2023-02-03
Change section language

lognaturel:
I've repeatedly had the experience of searching for Central docs that end up being in this "setting up" section that is now much broader. I think we should either use a broader term in the TOC or split off "Setting Up". "Setting Up" may have been chosen to parallel Collect but they're not the same -- Central requires ongoing maintenance whereas Collect does not.

I started with "Administrating ODK Central" but that is so long that it wraps.

I tried "Setup and Administration" which I like but it doesn't fit with the other headers.

I considered "Managing ODK Central" and I don't have a strong preference between that and "Running". "Operating" might also work.


ktuite:
"Running ODK Central" makes sense to me as someone who'd be looking for how to make central "go" and keep it going.

ktuite (approved):

yanokwa:
"Managing Central" feels a lot better to me because it better matches the role of "server management". [DHIS2](https://docs.dhis2.org/en/home.html) calls it management.

yanokwa:
Why don't we split it up into 2 sections: Installing and Managing?

lognaturel:
Have split "setup" into "install" and "manage". At some point it would be nice to add some general server administration advice to "manage" but I think this is a good improvement we can do first.

PR #1566, opened 2023-02-04
Add v2023.1 upgrade notes

lognaturel:
Maybe the `danger` admonition will be controversial. I really want this one to stand out.

Screen Shot 2023-02-04 at 9 07 07 AM


yanokwa (approved):
More important to get this message out first rather than get the message perfect.

PR #1573, opened 2023-02-09
Add special upgrade instructions for v2023.2 (postgres upgrade!)

alxndrsn:


lognaturel:
I see the space script is now in the repo. That means the pull and submodule upgrade happen before running the script. If there's not enough space, now you've checked out code that you can't build. That's probably ok but wanted to mention that it's different from previous instructions which had the space check first.

There needs to be a section on what to do if you have an external database. That's one of the things we really want QA on.

I was imagining that it'd be like the other version sections where there are some special instructions but the bulk of the steps are the standard instructions. Did that feel too disjoint?

alxndrsn:
> There needs to be a section on what to do if you have an external database. That's one of the things we really want QA on.

Added.

> I was imagining that it'd be like the other version sections where there are some special instructions but the bulk of the steps are the standard instructions. Did that feel too disjoint?

Once we're happy with the instructions, we can see if they can be "refactored" to share with the standard instructions.

alxndrsn:
> There needs to be a section on what to do if you have an external database. That's one of the things we really want QA on.

Added.

> I was imagining that it'd be like the other version sections where there are some special instructions but the bulk of the steps are the standard instructions. Did that feel too disjoint?

Once we're happy with the instructions, we can see if they can be "refactored" to share with the standard instructions.

lognaturel:
I was wondering why the code snippets were so big and I realized I messed up some of the formatting. Consider cherry picking https://github.com/getodk/docs/commit/1bc911618e0e589b2f77cbbf5a8c786eaa745a2f

lognaturel:
Replaced by #1577.

PR #1574, opened 2023-02-10
Add section on naming repeats to help with navigation

lognaturel:
Closes https://github.com/getodk/docs/issues/376

Not really sure what to name the little location summary at the top of a question. I went with form location summary but happy for other ideas.

Ran images through imageoptim.


yanokwa (changes requested):
The side by side images are confusing because there's no border on the image and they kind of look like the same thing if you aren't paying attention.

I would put one image, then some text, then the other image. Alternatively, add a caption or header to each.

lognaturel:
Yeah, you're right. Can we merge this and file a follow-on issue? I'll get it next time I do a mini doc sprint. But I think it's useful information to have out in the mean time.

yanokwa (approved):

PR #1577, opened 2023-02-21
Document Central v2023.2 upgrade to Postgres 14

lognaturel:
Alternative to #1573 

TODO:
- [x] Have folks touch ./files/allow-postgres14-upgrade even if they use a custom db
- [x] Warning that there is a lot to download and this will take a long time on slow connections

I rebased the commits from #1573  on master to get sidebar updates (clean rebase). Then I separated out default and custom db instructions into tabs and added instructions on how to determine which one is being used.

I tried to identify points at which something could go wrong and provide some guidance.

The part I'm least happy with is the part about dealing with the `docker-compose.yml` conflicts in the custom db case. Unfortunately I don't think we can assume that everyone following these instructions would know how to resolve merge conflicts. We've seen organizations have high turnover and folks with very little IT experience suddenly becoming admins. I want to make sure folks like that have the best shot possible at getting this right.

Screen Shot 2023-02-20 at 5 16 14 PM

[Archive.zip](https://github.com/getodk/docs/files/10788705/Archive.zip)

@alxndrsn feel free to push to this branch and/or cherry pick my commits onto yours.


alxndrsn:
For some reason I don't get a finger mouse-pointer :point_up_2: when mouse is hovering over these tabs.  Tested in both Firefox & Chrome.  Works fine on the example at https://sphinx-tabs.readthedocs.io/en/latest/.

docs/central-upgrade.rst
```diff
@@ -76,14 +77,170 @@ You'll be asked to confirm the removal of all dangling images. Agree by typing t
...
+.. tabs::
+
+  .. tab:: Default database
+
+    .. warning::
+      Before starting:
+
+      #. :doc:`Back up your server <central-backup>`.
+      #. Make sure you have some time available in case something goes wrong (we recommend at least 2 hours). You may want to announce a maintenance window.
+      #. Review upgrade notes for all versions between your current version and the version you are upgrading to.
```
ktuite, line 114: In a step below, it says "Log into the web interface and do some quick spot checks. For example, verify that submission counts and latest submission dates look right and try a data export." Would have liked something to tell me to make a note of these numbers before hand.

docs/central-upgrade.rst
```diff
@@ -76,14 +77,170 @@ You'll be asked to confirm the removal of all dangling images. Agree by typing t
...
+      #. Read the instructions at the top of this section carefully and **make sure you are actually using the default database configuration**. Following these instructions with a custom database setup could result in perceived data loss.
+
+    #. Log into your server. If you used our :doc:`DigitalOcean installation steps <central-install-digital-ocean>`, we suggest reviewing the section :ref:`central-install-digital-ocean-build` as a reminder, or if you can't remember your password to start at the top of that section to reset your password.
```
ktuite, line 117: Making the main action of each step bold would help me follow the steps. When I hadn't gone through any of the steps, I wanted to know more about what would happen in each step and what its purpose was.

> **Log into your server.** If you used...

docs/central-upgrade.rst
```diff
@@ -76,14 +77,170 @@ You'll be asked to confirm the removal of all dangling images. Agree by typing t
...
+
+    #. Restart the server and verify that everything works as expected.
+
+       .. code-block:: bash
+
+            docker-compose up -d
+
+    #. Log into the web interface and do some quick spot checks. For example, verify that submission counts and latest submission dates look right and try a data export.
+
+    #. The upgrade process performs a copy and leaves the old database intact. The following command will show you how much space this old database takes.
```
ktuite, line 186: Here, the steps switch from "You're done! You did it! Hopefully eerything looks good!" to more... optional/less time critical cleanup. I think this should be a separate section so you can know when you're done with the hardest part.

ktuite (commented):
I've successfully gone through the basic upgrade steps! 🎉🎉🎉 I kind of want to try again and see what happens if I don't follow the rules, don't touch the right file, etc. 

Left some comments about how the docs could give a little bit more context.

I went ahead and made these changes in a branch somewhere. I could do a PR on the PR, or just add them to this branch, but I wanted someone to see them first. They're here in this commit: https://github.com/getodk/docs/commit/66a5a478e05a159da813583bc10e7fdf0408bf9c

lognaturel:
Thanks, @ktuite, I have incorporated your commit!

yanokwa:
We should add the upgrade to Docker Engine/Compose. docker-compose reaches end-of-life in June 23. This step should come first. 

Sometimes we put the steps people need to follow in the docs, but I'd rather point to the source in case people have some weird setup. The Docker instructions are very thorough.

### upgrade docker 

Check to see if you have Docker Engine v23.x and Docker Compose v2.16.x or greater.
```
$ docker --version
Docker version 23.0.1, build a5ee5b1

$ docker compose version
Docker Compose version v2.16.0
```
If you are using old versions, follow the instructions at https://docs.docker.com/engine/install/ubuntu/ to upgrade. 

Once you are running the latest version, remove `docker-compose`. 
```
rm -f `which docker-compose`
```

yanokwa:
Upgrading the nginx container is simple. It might be easiest to put it as part of the postgres migration? After you run the stop, you can run `docker-compose rm -f nginx` then continue with the build, up, etc.

docs/central-upgrade.rst
```diff
@@ -76,14 +77,179 @@ You'll be asked to confirm the removal of all dangling images. Agree by typing t
...
+
+       *If you don't have enough space,* **stop here** and resume when you have increased the disk space available. You may achieve this by clearing out data you don't need (e.g., logs) or by increasing the total disk space available (e.g., by :ref:`adding external storage <central-install-digital-ocean-external-storage>`).
+
+    #. **Create a file to prove that you're carefully reading these instructions.** This is required to continue.
+
+       .. code-block:: bash
+
+          touch ./files/allow-postgres14-upgrade
+
+    #. **Stop Central.** Central needs to be stopped in order to make a clean copy of the database.
```
yanokwa, line 155: You should be able to build, then stop. That will minimize your downtime. That's what we've done historically, but this should be tested.

lognaturel: I'm now wondering whether this is a good idea. That means things can happen between backup and stop. Maybe our standard recommendation should be to stop everything, back up, then do what needs to be done. Folks who are more advanced can do whatever they want.

docs/central-install-digital-ocean.rst
```diff
@@ -296,21 +296,21 @@ Central ships with a basic EXIM server bundled to forward mail out to the intern
...
+         "secure": false,
+         "auth": {
+           "user": "my-smtp-user",
+           "pass": "my-smtp-password"
+         }
        }
      }
-   }
 
 3. Build and run: ``docker-compose build service``, ``docker-compose stop service``, ``docker-compose up -d service``.
```
ktuite, line 315: I'm seeing `docker-compose` in the digital ocean install instructions, e.g. in Advanced Configuration Options and earlier in Changing Server Settings.

docs/central-install-digital-ocean.rst
```diff
@@ -320,40 +320,40 @@ Using a Custom Database Server
...
-  .. code-block:: console
+   .. code-block:: json
 
-    "database": {
-      "host": "my-db-host",
-      "user": "my-db-user",
-      "password": "my-db-password",
-      "database": "my-db-table"
-    },
+      "database": {
```
ktuite, line 345: I haven't read through the part where the config changes but I think this and some of the other nearby instructions are going to be different...

ktuite (commented):
Was skimming the install on digital ocean section and the other files with minor changes. Looking again at the 2023.2 upgrade instructions next.

docs/central-upgrade.rst
```diff
@@ -25,65 +26,351 @@ Upgrade notes
...
+   
+       #. **Upgrade your database server**. We recommend using the latest point release of PostgreSQL 14 that is available.
+
+       #. **Create a file to prove that you're carefully reading these instructions.** This is required to continue.
+   
+          .. code-block:: console
+   
+            $ touch ./files/allow-postgres14-upgrade
+   
+       #. Follow the :ref:`standard upgrade instructions <central-upgrade-steps>`.
```
ktuite, line 357: This could totally stay as a link to the one place the instructions are written, but I was _thinking_ what if there was a little summary of the instructions below:

```
cd central (they've probably already done this)
git pull
git submodule update -i
docker compose build
docker image prune
docker compose up -d
```

docs/central-upgrade.rst
```diff
@@ -25,65 +26,351 @@ Upgrade notes
...
+
+   .. code-block:: console
+
+     $ rm -f `which docker-compose`
+
+#. **Migrate configuration customizations.** This will simplify future Central upgrades. First, check what files have been customized:
+
+   .. code-block:: console
+
+     $ git status
```
ktuite, line 131: Was just wondering what some of these instructions would look like if they did the `git pull` step first. How different would cleaning up the config files be? In other words, how important is it that you migrate the custom config options *before* pulling the new code changes?

ktuite (commented):
Some thoughts from reading the docs w/o trying things yet!

docs/central-upgrade.rst
```diff
@@ -25,65 +26,340 @@ Upgrade notes
...
+
+     #. See what changes have been made:
+
+        .. code-block:: console
+
+          $ git diff docker-compose.yml
+
+        You will see additions in green with ``+`` prefixes.
+
+        You can ignore any changes related to a custom database because those have been addressed by migrating ``files/service/config.json.template``. Copy any other changes to a scratch file on your local computer. You will use this to copy your custom values into the new format.
```
yanokwa, line 199: I'd put a new line before Copy.

docs/central-upgrade.rst
lognaturel: I have a hard time making sense of how to fix the merge conflicts myself, even after doing it a few times. I was not able to come up with a good way to explain it.

docs/central-upgrade.rst
```diff
@@ -25,65 +26,340 @@ Upgrade notes
...
+
+     $ cd central
+
+#. **Upgrade Docker if needed.** Check to see if you have Docker Engine v23.x and Docker Compose v2.16.x or greater:
+
+   .. code-block:: console
+
+     $ docker --version && docker compose version
+
+   If you are using old versions, follow the instructions at https://docs.docker.com/engine/install/ubuntu/ to upgrade.
```
yanokwa, line 119: If you are using old versions, follow the instructions to install [Docker Engine](https://docs.docker.com/engine/install/ubuntu) (not Desktop).

We link to instructions for Ubuntu because that's the OS we recommend. You'll need to setup the apt repository and install the latest version of Docker Engine.

docs/central-install-digital-ocean.rst
lognaturel: Should now be fixed.

docs/central-install-digital-ocean.rst
lognaturel: Addressed.

docs/central-upgrade.rst
```diff
@@ -25,65 +26,351 @@ Upgrade notes
...
+          $ git checkout -- docker-compose.yml
+
+   .. dropdown:: ``files/enketo/config.json.template`` or any others
+     :icon: file-code
+
+     #. Stash changes so they can be applied after the upgrade. These are considered advanced customizations and you may need to resolve merge conflicts when you re-apply them.
+
+        .. code-block:: console
+
+          $ git stash
```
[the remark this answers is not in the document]
lognaturel: It's only for files that are not the two special ones mentioned above. The most likely is `files/enketo/config.json.template` but `files/nginx/odk.conf.template` is another possibility and there could be others. Most folks will NOT need to deal with this which is why I want to bury it.

docs/central-upgrade.rst
lognaturel: The standard instructions are concise and copy-able enough now that I'd rather send them there so they can see any unexpected output step by step.

docs/central-upgrade.rst
lognaturel: To be clear, the process I've documented is what I ended up finding the easiest for myself.

docs/central-upgrade.rst
```diff
@@ -25,65 +26,342 @@ Upgrade notes
...
+   If you see files listed with a `modified:` prefix, follow instructions for each of them:
+
+   .. dropdown:: ``files/service/config.json.template``
+     :icon: file-code
+
+     #. See what changes have been made:
+
+        .. code-block:: console
+
+          $ git diff files/service/config.json.template
```
ktuite, line 144: As I'm going through the steps, I feel very nervous about this step of copying values into another file and deleting my changes to this file long before I know if it will actually work. I think there should be a recommendation to make a copy/backup of this file to refer to just in case.

ktuite: It is also awkward to check my diff and then enter new values into `.env`. I also wish I had the `.env.template` already.

docs/central-upgrade.rst
```diff
@@ -25,65 +26,342 @@ Upgrade notes
...
+          EMAIL_IGNORE_TLS=true-or-false
+          EMAIL_SECURE=true-or-false
+          EMAIL_USER=my-email-user
+          EMAIL_PASSWORD=my-email-password
+
+        .. note::
+
+          ``EMAIL_IGNORE_TLS`` should generally be set to ``false``. ``EMAIL_SECURE`` should generally be set to ``true`` if you use port 465 and to ``false`` for other ports.
+
+     #. Discard all ``files/service/config.json.template`` customizations. Make sure you have correctly copied all of them into ``.env``. You may want to keep the copy on your local computer until you have verified that everything works.
```
ktuite, line 182: Oh there's the message about making a backup, right after it said to get rid of it 😅 "You may want to keep the copy on your local computer until you have verified that everything works."

ktuite: "Make sure you have correctly copied all of them into" ... 
It's unclear how to copy other customizations into .env and which ones will be supported. You could say "Refer to [.env.template](https://github.com/getodk/central/blob/next/.env.template)  (but in master, not next) to see which things will be configured now in .env.

ktuite: `email.serviceAccount` --> `EMAIL_NO_REPLY` ?

docs/central-upgrade.rst
ktuite: They way I finally wrapped my head around this section:

1. Make backups for files/service/config.json.template and docker-compose.yml
2. Reset the git contents of those files
3. Things from those two files specifically are moving to `.env` (i'd prefer if i could put them back into .env after i git pull)
4. git stash any other changes you have so you can apply those changes after the upgrade

docs/central-troubleshooting.rst
```diff
@@ -46,21 +46,21 @@ You may run into a "Could not connect with Server" error message when previewing
 
 To resolve this problem, first identify your upstream DNS servers. Run ``cat /run/systemd/resolve/resolv.conf`` to see your current list of nameservers with their IP addresses. They will look like this:
 
-  .. code-block:: console
+.. code-block:: console
```
ktuite, line 49: Above in this file, it says you can run `docker logs --tail 100 central_service_1` but the docker containers have hyphens now `docker logs --tail 100 central-service-1`

ktuite (commented):
More comments!

docs/central-upgrade.rst
lognaturel: Cool! That looks about like what we have now but we've broken it down a bit more? Most folks will only have changes to either `files/service/config.json.template` or  `docker-compose.yml`, hence the separation. I think copying to the local machine is much easier because of having multiple windows at once.

> i'd prefer if i could put them back into .env after i git pull

Why?

docs/central-troubleshooting.rst
```diff
@@ -110,57 +110,57 @@ Database reset after running Docker command
...
 
 4. Stop the Docker containers to prepare for restoration.
 
-  .. code-block:: console
+   .. code-block:: console
 
-    cd ~/central
-    docker-compose stop
+     $ cd ~/central
+     $ docker-compose stop
```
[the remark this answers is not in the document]
lognaturel: This stuff is only going to be possible in versions before v2023.2 so it feels more consistent to leave the `docker-compose` that was documented at that time.

docs/central-troubleshooting.rst
```diff
@@ -110,57 +110,57 @@ Database reset after running Docker command
...
 
-  .. code-block:: console
+   .. code-block:: console
 
-    cd ~/central
-    docker-compose build
-    docker-compose up -d
+     $ cd ~/central
+     $ docker-compose build
+     $ docker-compose up -d
```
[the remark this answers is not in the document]
lognaturel: Same as above.

docs/central-troubleshooting.rst
lognaturel: Ah yes!

docs/central-upgrade.rst
lognaturel: > Oh there's the message about making a backup

Ah! I think you may have missed the part about copying the diff lines into a file on your local computer? Maybe that needs to be clarified?

> email.serviceAccount --> EMAIL_NO_REPLY ?

Indeed, good catch!

docs/central-upgrade.rst
ktuite: I did copy from the blocks, I just felt like I was doing this kind of blindly, not knowing where I'd end up.

I did go seek out the template, and seeing it made me think, "Oh, that's what i'm aiming for, that's not scary". I also used it to copy over the sentry values.

docs/central-upgrade.rst
ktuite: It's true, I did not read the sentence that said "Copy those to a scratch file on your local computer. You will use this to copy your custom values into the new format."  Though it is what I ended up doing.

lognaturel: Oops! Ok, I changed the step to make it about copying since that's the important action. How does that look?

docs/central-upgrade.rst
lognaturel: I think maybe this is because Cloud customizes everything? For most people only one or maybe two of these blocks will be relevant. Most common will be just email, I think.

docs/central-upgrade.rst
```diff
@@ -170,6 +170,10 @@ This is *critical infrastructure upgrade*. In particular, it upgrades the includ
 
           If your password has special characters in it, you will need to put single quotes ( ``'`` and ``'``) around the password. Values without special characters do not need quotes around them.
 
+        .. note::
+
+          If your database uses an SSL connection, add ``DB_SSL=true`` to ``.env``. If you don't need an SSL connection, omit that variable. If you enable SSL, the ``rejectUnauthorized`` option is also set to ``true`` which allows self-signed certificates. Make sure you have appropriately restricted access to your external database.
```
matthew-white, line 175: `false` instead of `true` for `rejectUnauthorized` I think.

lognaturel: 😬

ktuite (approved):
Yay!

Issue #1579, opened 2023-02-21
Improve images in section on naming repeats for better navigation


lognaturel:
https://github.com/getodk/docs/pull/1574

PR #1580, opened 2023-02-21
Add paragraphs and image spacing

yanokwa:
Follow up on #1574.

docs


lognaturel:
You're happy with that? I filed #1579 and was thinking I'd show the whole phone after all but if you're ok with this we can close the issue.

lognaturel (approved):

PR #1596, opened 2023-04-05
Updated 'Form metadata settings'

grzesiek2010:
I've updated [Form metadata settings](http://localhost:8000/collect-settings/#form-metadata-settings) according to the current state of ODK Collect. It looks like [Metadata](http://localhost:8000/form-question-types/#metadata) does not require any changes.


lognaturel (approved):
Thank you!

PR #1598, opened 2023-04-11
Remove Google Drive/Sheets

yanokwa:
I made sure everything rendered properly. I also search for Drive, Sheets, Google to find candidates for removal/edits.

### DIFF
--- a/docs/collect-import-export.rst
+++ b/docs/collect-import-export.rst
@@ -144,14 +144,12 @@ Here are the keys for all settings and the set of values they can take:
     "general" : {
 
       // Server
-      "protocol": {"odk_default", "google_sheets"},
+      "protocol": {"odk_default"},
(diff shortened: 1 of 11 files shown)


docs/collect-import-export.rst
```diff
@@ -144,14 +144,12 @@ Here are the keys for all settings and the set of values they can take:
     "general" : {
 
       // Server
-      "protocol": {"odk_default", "google_sheets"},
+      "protocol": {"odk_default"},
```
yanokwa, line 147: Will we still have protocol going forward? And will it be an array.

lognaturel: We will leave it for backwards-compatibility I imagine. Maybe we can remove it from docs, though.

lognaturel (approved):

PR #1599, opened 2023-04-11
Deprecate briefcase

yanokwa:
Made sure everything renders well

### DIFF
--- a/docs/collect-connect.rst
+++ b/docs/collect-connect.rst
@@ -46,4 +46,3 @@ Other options
 
   collect-connect-google
   Transferring blank and completed forms directly with adb <collect-adb>
-  Using ODK Briefcase  <briefcase-using>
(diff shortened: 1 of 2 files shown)


lognaturel:
I thought we'd made an explicit choice to keep this in because it's helpful? We may not maintain Briefcase anymore but it still works well in many cases.

yanokwa:
I think it's become way less relevant for Collect. Most times when we need to get data off a device, we have folks zip up the SD card. It's been years since I asked anyone to use Briefcase to get submissions off a device. For that reason, I think this PR should be merged.

lognaturel:
> when we need to get data off a device, we have folks zip up the SD card. It's been years since I asked anyone to use Briefcase to get submissions off a device.

But Briefcase was never used for transferring from Collect to a computer! It's used for either exporting a CSV or submitting to a server. It feels to me like the entirely offline use cases still exist.

PR #1601, opened 2023-04-18
Embed Central API docs by rendering its OpenAPI spec

### LINKED ISSUE #1600: Embed API docs
It would be nice if API documentation is also inside docs.getodk instead of a separate endpoint

sadiqkhoja:
closes #1600 

- Python module to parse openAPI specification and generate RST files for each resource in API spec

#### How it works?

- Before sphinx build, new `openapi` module is executed
- It parses `central.yml` file and extracts relevant data
- Passes the data to mustache templates and the template engine generates RST files

#### TODO:

- [x] Write unit tests
- [x] Review and validate all endpoints, for some we have missing request/response example and schema
- [x] May be break __main__.py into multiple files


lognaturel:
I've done a quick first pass over this and plan to go deeper next week.

I think some of the high level decisions you made are:
- Use OpenAPI for the API definition moving forward. That seems non-controversial now that it has "won" as a standard.
- Maintain the Central API docs in the docs repo rather than alongside the backend code. I think that's a bit unfortunate and adds friction but it's probably pragmatic. Did you consider alternatives? Maybe CI in backend could push changes to the API docs to this repo? I find it nice when docs changes are with corresponding code changes but now that I think about it, maybe that hasn't been the typical process anyway.
- Write custom code to go from OpenAPI to Sphinx. I think you said none of the existing Sphinx plugins you looked at were complete enough for our needs, is that right?
- Create separate pages for different API sections instead of one massive page. There are tradeoffs: because search is not very good it may be hard to find things.
- Anything else?

What kind of feedback do you think I should aim for? My current thinking is that quality and maintainability of the conversion script is not all that important. I'd probably do a read through the rendered docs looking for anything unexpected. Anything else?

sadiqkhoja:
> Maybe CI in backend could push changes to the API docs to this repo?

That would be ideal. I can do that as low priority task

> quality and maintainability of the conversion script is not all that important

Great, but if you can give some feedback on that will be nice, that would be my learning. I may incorporate them later as low priority

> I'd probably do a read through the rendered docs looking for anything unexpected

That would be great. One known issue: MD tables are not rendered correctly in API description e.g. /central-api-odata-endpoints/#data-document

lognaturel (approved):

sadiqkhoja:
I have few things to complete today, won't get chance to fix this. Looks like issue related to directory change

PR #1613, opened 2023-05-18
Add pull option to docker compose build

lognaturel:
As of https://github.com/getodk/central/pull/428, we pin all base images to minor versions. The goal is for hosters to get patches at time of upgrade.

I considered adding `pull` for all the sections in https://docs.getodk.org/central-install-digital-ocean/#advanced-configuration-options and https://docs.getodk.org/central-troubleshooting/ The nice thing with that is there would be more opportunities to get point releases. But all of those are rare so it doesn't seem worth possibly adding a confusing setep.

<S>
My understanding is that this is what the `--pull` option for `build` does and that `docker compose pull` would not update base images. https://github.com/docker/compose/issues/2742#issuecomment-180077368
</s>

### DIFF
--- a/docs/central-upgrade.rst
+++ b/docs/central-upgrade.rst
@@ -54,15 +54,16 @@ Upgrade steps
 
   $ git submodule update -i
 
-4. **Build** from the latest code you just fetched.
+4. **Build** from the latest code you just fetched. The ``pull`` options pull updates to all Docker images.
 
 .. code-block:: console
 
-  $ docker compose build
+  $ docker compose pull
+  $ docker compose build --pull
 
 .. note::
 
-  If you run into problems with this step, try stopping Central (``docker compose stop``) and then retry ``docker compose build``.
+  If you run into problems with this step, try stopping Central (``docker compose stop``) and then retry ``docker compose build --pull``.
 
 5. **Clean up unused Docker images**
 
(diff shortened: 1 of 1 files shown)


yanokwa:
In my testing, `docker compose build --pull` didn't update anything, but `docker compose pull` seemed to update a lot.

I think in Upgrade steps, Step 4, should be..
```
docker compose pull
docker compose build
```

This is better than what we have now, and we don't have to put pulling everywhere. I sure would like to understand the difference between the commands though.

lognaturel:
Maybe `docker-compose build --pull --no-cache` would be the same as `docker compose pull` followed by build?

yanokwa:
Bard says...

`docker compose pull`: Pulls the images for the services defined in the compose file.
- You might want to use `docker compose pull `if you are working on a development machine and you want to make sure that you are always using the latest versions of the images for the services that you are developing.

`docker compose build --pull`: Builds the images for the services defined in the compose file and also pulls any newer versions of the images that are already present on the local machine.
- You might want to use `docker compose build --pull` if you are deploying a production application and you want to make sure that you are using the latest versions of the images for the services that are running in production.

yanokwa:
https://www.reddit.com/r/docker/comments/13lz580/comment/jkssd7y/ has more information.

I think we need to do both?
* `docker compose pull` to get update images referenced in the `docker-compose.yml` file.
* `docker compose build --pull` to get update images used in any `Dockerfiles`

docs/central-upgrade.rst
```diff
@@ -54,15 +54,16 @@ Upgrade steps
 
   $ git submodule update -i
 
-4. **Build** from the latest code you just fetched.
+4. **Build** from the latest code you just fetched. The ``pull`` options pull updates to all Docker images.
```
yanokwa, line 57: The ``pull`` option ensures all Docker images are up-to-date.

yanokwa (approved):

PR #1619, opened 2023-05-31
add venv instructions to readme

issa-tseng:
since it seems to be a requirement right now


lognaturel (approved):

lognaturel:
I think @yanokwa and @matthew-white do without but it never works for me!

PR #1630, opened 2023-07-13
Exclude prompts from copy button

lognaturel:
Currently, when you try to copy a CLI example using the copy button, the `$` prompt comes with it. 💢 

With this configuration update, it won't. 🎉 

I've tried this locally.

Docs I followed: https://sphinx-copybutton.readthedocs.io/en/latest/use.html


yanokwa (approved):
LGTM

PR #1640, opened 2023-08-29
Document user access to entities

matthew-white:
This PR is intended to complete the docs portion of getodk/central#462. I'm starting out with a couple of commits that update the table of user roles, then @lognaturel will take a look at other text on the same page.

Working on this change made me remember that access to entities works pretty differently from access to other resources. That's because Data Collectors' and App Users' access to entities is mediated by forms and submissions. For example, by default those users can't list entities, but if a form attachment to which they have access is linked to an entity list, _then_ they'll be able to list its entities. I tried to distinguish between different types of access, but I'm not sure I found the right language. I'll leave a couple of line comments about specifics.

Relatedly, some of the new actions are longer, so I increased the width of the Action column from 16 characters to 32 characters. I don't think that number has any user-facing effects on its own, but it allowed me to add longer actions. In one case, I wrapped an action over two lines. I think the table still isn't overly wide:


docs/central-users.rst
```diff
@@ -19,67 +19,94 @@ Central features Role-based User permissioning. In the current release of Centra
...
++----------------------------------+---------------+-----------------+----------------+----------------+
+| Create from Submission           | x             | x               |                | x              |
++----------------------------------+---------------+-----------------+----------------+----------------+
+| Create Directly                  | x             | x               |                |                |
++----------------------------------+---------------+-----------------+----------------+----------------+
+| View in Linked Forms             | x             | x               | x              | x              |
++----------------------------------+---------------+-----------------+----------------+----------------+
+| | View & Download Anywhere,      | x             | x               | x              |                |
+| | Not Just in Linked Forms       |               |                 |                |                |
++----------------------------------+---------------+-----------------+----------------+----------------+
```
matthew-white, line 79: You can see that this text is slightly misaligned in the screenshot in the description. It's like it's missing left padding. I tried another RST syntax for multiple lines: instead of the second `|`, a blank line between the multiple lines. But the first line ended up wrapping awkwardly.

```rst
+----------------------------------+---------------+-----------------+----------------+----------------+
| View & Download Anywhere,        | x             | x               | x              |                |
|                                  |               |                 |                |                |
| Not Just in Linked Forms         |               |                 |                |                |
+----------------------------------+---------------+-----------------+----------------+----------------+
```

docs/central-users.rst
```diff
@@ -19,67 +19,94 @@ Central features Role-based User permissioning. In the current release of Centra
...
+| Edit & Review                    | x             | x               |                |                |
++----------------------------------+---------------+-----------------+----------------+----------------+
+| Comment                          | x             | x               | x              |                |
++----------------------------------+---------------+-----------------+----------------+----------------+
+| View & Download                  | x             | x               | x              |                |
++----------------------------------+---------------+-----------------+----------------+----------------+
+| OData Access                     | x             | x               | x              |                |
++----------------------------------+---------------+-----------------+----------------+----------------+
+| **Project Entity Lists**                                                                             |
```
matthew-white, line 59: I personally don't think the "Project" prefix adds much, but I wanted to be consistent.

docs/central-users.rst
```diff
@@ -19,67 +19,94 @@ Central features Role-based User permissioning. In the current release of Centra
...
+| **Project Entities**                                                                                 |
++----------------------------------+---------------+-----------------+----------------+----------------+
+| Create from Submission           | x             | x               |                | x              |
++----------------------------------+---------------+-----------------+----------------+----------------+
+| Create Directly                  | x             | x               |                |                |
++----------------------------------+---------------+-----------------+----------------+----------------+
+| View in Linked Forms             | x             | x               | x              | x              |
++----------------------------------+---------------+-----------------+----------------+----------------+
+| | View & Download Anywhere,      | x             | x               | x              |                |
+| | Not Just in Linked Forms       |               |                 |                |                |
```
matthew-white, line 78: Do actions need to be title-case? I feel like it doesn't work particularly well for longer actions like this one.

matthew-white, line 78: I know it's not the form that's linked to the entity list, it's the form attachment, but I thought it'd be confusing to say "Form Attachment" here: more explanation would be required.

docs/central-users.rst
```diff
@@ -19,67 +19,94 @@ Central features Role-based User permissioning. In the current release of Centra
...
++----------------------------------+---------------+-----------------+----------------+----------------+
+| Update Settings                  | x             | x               |                |                |
++----------------------------------+---------------+-----------------+----------------+----------------+
+| **Project Entities**                                                                                 |
++----------------------------------+---------------+-----------------+----------------+----------------+
+| Create from Submission           | x             | x               |                | x              |
++----------------------------------+---------------+-----------------+----------------+----------------+
+| Create Directly                  | x             | x               |                |                |
++----------------------------------+---------------+-----------------+----------------+----------------+
+| View in Linked Forms             | x             | x               | x              | x              |
```
matthew-white, line 75: Here, I list "view" before "update", but for submissions, it's "view" that comes last. I thought "view" should come earlier here, because the two types of view actions contrast in a similar way as the two types of create actions directly above.

docs/central-users.rst
```diff
@@ -19,67 +19,94 @@ Central features Role-based User permissioning. In the current release of Centra
...
++----------------------------------+---------------+-----------------+----------------+----------------+
+| | View & Download Anywhere,      | x             | x               | x              |                |
+| | Not Just in Linked Forms       |               |                 |                |                |
++----------------------------------+---------------+-----------------+----------------+----------------+
+| OData Access                     | x             | x               | x              |                |
++----------------------------------+---------------+-----------------+----------------+----------------+
+| Update Data                      | x             | x               |                |                |
```
matthew-white, line 82: We use the term "edit" above for submissions, but in Frontend, we only talk about "updating" entities.

matthew-white, line 82: For .5, we'll need to update this to account for entity updates via submission.

lognaturel:
Here's an alternative take I'd love your thoughts on: https://github.com/getodk/docs/compare/master...lognaturel:docs:entities-roles

In particular:
Screen Shot 2023-08-31 at 12 02 42 PM

Screen Shot 2023-08-31 at 12 02 49 PM

The additional rows about actions that could be taken through Forms or Submissions felt confusing to me. It feels easier to me to convey the idea that if Forms are defined to do stuff related to Entities, then that stuff can be done by anyone who can fill the form.

I thought about adding information about how the form definition can filter what users can see but that they can always get the full document from their client but it didn't feel like the right place for that level of detail.

docs/central-users.rst
lognaturel: I don't think they should be. I removed it in a commit.

matthew-white:
I haven't taken a full look at the branch yet, but I like what's in the screenshot above! I agree that the note box and the notes below "Project Entity Lists" and "Project Entities" are clearer than the additional action rows. I also like that the note box mentions app users.

For the note below "Project Entity Lists", I think the action that the note pertains to is "View and download" under "Project Entities", not any of the actions under "Project Entity Lists". I see how it makes sense this way though, since it's the entity list that's attached to the form. That said, I think it's slightly confusing that the first note says that Data Collectors get read-only access to entity lists, but then the second note says that they can create entities. There's not actually a contradiction, but I think "-only" opens up the potential for confusion. As an alternative for the first note, what do you think about "Data Collectors can view Entity Lists attached to Forms"?

> It's not possible to push to other people's docs branches because of something about gitlfs. Feel free to either cherry-pick commits you like or tell me to open another PR!

That's an unexpected result of gitlfs! Since you've continued making progress, and your branch includes the commits here, I think we should close this PR and open a new one based on your branch.

lognaturel:
Replaced by https://github.com/getodk/docs/pull/1645

matthew-white:
> It's not possible to push to other people's docs branches because of something about gitlfs.

I just noticed that I didn't have "Allow edits by maintainers" checked. Maybe that was the reason for the access issue? I thought that option was on by default, I'll try to look out for it in the future.

PR #1647, opened 2023-09-12
Fix warnings and make sure OpenAPI is always built

lognaturel:
Fixes a few warnings I'd never noticed before.

The last commit actually runs the OpenAPI to RST conversion that https://github.com/getodk/docs/pull/1601 omitted. I completely forgot to come back to review how the conversion was triggered. It was only added to the autobuild make rule so it didn't happen when HTML was built for serving.


Makefile
```diff
@@ -14,11 +14,8 @@ help:
 
 .PHONY: help Makefile
 
-clean:
```
lognaturel, line 17: Just use the catch-all for this.

docs/conf.py
```diff
@@ -109,3 +110,6 @@
 
 # Exclude all prompt characters generated by pygments and all console output in copy button
 copybutton_exclude = '.linenos, .gp, .go'
+
+# Convert Central OpenAPI spec to RST
+openapi
```
lognaturel, line 115: It would be more expected to call a function here. But Python modules and imports are so convoluted I don't want to mess with it right now.

yanokwa (approved):
LGTM

lognaturel:
@sadiqkhoja the OpenAPI conversion was definitely not running for `make dirhtml` which is what is run by CI. So I was convinced [650d13e](https://github.com/getodk/docs/pull/1647/commits/650d13e00d184617f960f6b594570f85cf0572c0) was the fix but it still looks like the conversion isn't run at all!

PR #1651, opened 2023-09-13
Replace Dataset with Entity List

lognaturel:
Do not merge until Central v2023.4 and pyxform 12.1.2 are released!

Closes https://github.com/getodk/central/issues/466

I made a few cosmetic changes as I went, I don't think any should be controversial! @matthew-white and @ktuite, I'll let you two decide who most wants to review this. @ktuite at some point we should talk about whether we should consider updating your video. It's still very relevant! But the change in terminology is a bit distracting.

### DIFF
--- a/docs/central-entities.rst
+++ b/docs/central-entities.rst
@@ -28,7 +28,7 @@ Next, a new feature. You can now tell Central that some of your incoming Submiss
 
 .. image:: /img/central-entities/intro-diagram.png
 
-Using this functionality, you don't have to create and upload Entity Lists yourself. When Submissions arrive and are approved, Central will create new Entities that will appear in follow-up Forms as soon as those Form updates are received.
+Using this functionality, you don't have to create and upload Entity Lists yourself. When Central receives submissions (or they are approved, see :ref:`the settings section <central-entities-settings>`), new Entities are automatically created. As soon as users of the Follow-up Form(s) get Form updates, they can interact with the new Entities.
 
 If you are interested in seeing how Entities can fit into your workflow right away, we recommend following the Quick Start guide below, where you will upload a tree registration Form and a tree follow-up Form we have created already and see how trees are created by one Form and appear in another.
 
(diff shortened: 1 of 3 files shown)


docs/central-entities.rst
```diff
@@ -159,21 +166,19 @@ Test your Form to make sure it works and collects the data that you need.
 
 .. _central-entities-registration-forms-destination:
 
-Specify the Dataset the Form should save Entities to
-""""""""""""""""""""""""""""""""""""""""""""""""""""
+Specify the Entity List the Form should save Entities to
+""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
 
-Add a new ``entities`` sheet to your XLSForm. This is where you will specify your Dataset’s name, under ``dataset``.
+Add a new ``entities`` sheet to your XLSForm. This is where you will specify your Entity List name, under ``list_name``.
```
lognaturel, line 172: @ktuite I'm specifically interested in your thoughts on `list_name` since you've made educational materials for all this. Do you think it's helpful or harmful that it's the same as the column header to specify the list that a choice is in from the `choices` sheet?

I thought it was helpful but I'm questioning that a little bit because the usage is different. On the `choices` tab, we're explicitly stating the list that a concrete item belongs in. Here, we're generally declaring the list that abstract entities will be created in. Maybe that's similar enough?

matthew-white:
I think @ktuite will be taking a look at this PR. Thank you, @ktuite!

> and pyxform 12.1.2

Are we planning to ship a pyxform change with Central .4? It might be nice for what's in the docs to match the pyxform that we ship with Central.

lognaturel:
> Are we planning to ship a pyxform change with Central .4?

Yes and I just reopened the corresponding issue. Realized that as I worked on these docs yesterday.

docs/central-entities.rst
ktuite: i think `list_name` is okay. the entities tab has such a different purpose and shape than the choices tab that i think it's ok for the column header to be the same but have slightly different meanings within the different context. this happens with `label`, too. having some other word that is slightly different yet also isn't a perfect fit seems like it would be more confusing

docs/central-entities.rst
```diff
@@ -65,67 +75,64 @@ In this Quick Start, you will upload two Forms we have already authored for you:
 First, let's prepare the Forms for use with Central.
 
 1. Create a new Project, if you can. It can be easier to see what's going on without other Forms in the way.
-2. Upload the `Tree registration Form <https://docs.google.com/spreadsheets/d/1xboXBJhIUlhs0wlblCxcQ3DB5Ubpx2AxLDuaXh_JYyw/edit#gid=2050654322>`_. Notice how it has seen from the Form definition that the Form creates Entities in a Dataset, and populates three Entity Properties.
-3. Publish the Form. As the dialog says, this action creates the ``trees`` Dataset within the Project you created.
+2. Upload the `Tree registration Form <https://docs.google.com/spreadsheets/d/1xboXBJhIUlhs0wlblCxcQ3DB5Ubpx2AxLDuaXh_JYyw/edit#gid=2050654322>`_. Notice how Central knows from the Form definition that the Form creates Entities in an Entity List, and populates three Entity Properties.
```
ktuite, line 78: We should update "dataset" --> "list_name" in this demo form.

ktuite (approved):
I'm liking this dataset -> "entity list" change more and more because now I can tell that pre-entity-era "datasets", like "form datasets" are different. 

In `central-intro.rst`, should this line be updated or clarified?

```
* Server-managed datasets for registration/follow-up workflows"
```

Maybe "server-managed datasets (entity lists)..."?

lognaturel:
> In central-intro.rst, should this line be updated or clarified?

Good catch! Fixed.

docs/central-entities.rst
lognaturel: Yes, yes, thank you! I'll set myself a reminder to do that when we merge.

PR #1653, opened 2023-09-15
Fix build by fixing openapi import

ktuite:
I was running into this error when I tried to build, and when I looked at the differences between this branch and an old version of `master` I had that did build, I figured out it was coming from the change to add the openapi rendering. 

```
$ make autobuild
sphinx-autobuild -b dirhtml "docs" "docs/_build/html"
[sphinx-autobuild] > sphinx-build -b dirhtml /Users/ktuite/Desktop/code/odk/docs/docs /Users/ktuite/Desktop/code/odk/docs/docs/_build/html
Running Sphinx v5.3.0

Configuration error:
There is a syntax error in your configuration file: invalid syntax (spec_processor.py, line 270)

Command exited with exit code: 2
The server will continue serving the build folder, but the contents being served are no longer in sync with the documentation sources. Please fix the cause of the error above or press Ctrl+C to stop the server.
```

`import openapi` added to `requirements.txt`
In `docs/conf.py`, import of `openapi` moved to above the line that looks for extensions directory

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -16,9 +16,9 @@
 import platform
 import os
 import sys
+import openapi
 sys.path.insert(0, os.path.abspath('../extensions/'))
 import sphinx_video
-import openapi
 
 # -- Project information -----------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


ktuite:
I heard this is the wrong `openapi` and the one mentioned in the conf.py file is meant to be an internal/custom one. And that going from python3.8->3.10+ should fix my problem.

PR #1654, opened 2023-09-15
add autogenerated page to ignore list for sphinx-autobuild

sadiqkhoja:

This fixes infinite watch loop for sphinx-autobuild, which looks for changes in `docs` directory and custom openApi extension also add/updates files in the same directory so we want to ignore watching autogenerated pages

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ help:
 .PHONY: help Makefile
 
 autobuild:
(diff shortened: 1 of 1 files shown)


lognaturel:
I think you're also going to have to add `|_build` to avoid the same for our custom video extension. Sorry, I removed that when trying to fix the other make directives and didn't notice it was very much important!

lognaturel (approved):

PR #1655, opened 2023-09-16
Improve DKIM support

yanokwa:
This key needs to read by the Debian-exim user in the mail container. 

Currently, the key is mounted as [read-only](https://github.com/getodk/central/blob/master/docker-compose.yml#L33). And it inherits whatever permissions it has on the host, but there's no easy way to chgrp the file as part of mounting. 

~~The likely best way is to do so in the entrypoint.sh of the container ([example](https://github.com/ix-ai/smtp/blob/master/entrypoint.sh#L54C8-L54C19)), but I don't want to make upstream changes. I also don't want to have some command file in the docker-compose.yml~~

I've decided to be the change, and I've patched this upstream at https://github.com/ix-ai/smtp/pull/27.

### DIFF
--- a/docs/central-upgrade.rst
+++ b/docs/central-upgrade.rst
@@ -60,8 +60,7 @@ Upgrade steps
 
 .. code-block:: console
 
-  $ docker compose pull
-  $ docker compose build --pull
+  $ docker compose pull && docker compose build --pull
 
 .. note::
 
@@ -84,7 +83,7 @@ You'll be asked to confirm the removal of all dangling images. Agree by typing t
 
 .. code-block:: console
 
-  $ docker compose up -d
+  $ docker compose stop && docker compose up -d
 
 .. _version-specific-instructions:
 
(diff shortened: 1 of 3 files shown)


docs/central-upgrade.rst
```diff
@@ -90,6 +91,62 @@ You'll be asked to confirm the removal of all dangling images. Agree by typing t
...
+
+     .. note:: After the upgrade, consider deleting the now unused DKIM folder and its contents.
+
+        .. code-block:: console
+
+         $ cd central
+
+        .. code-block:: console
+
+         $ rm -r central/files/dkim
```
lognaturel, line 148: Should be `files/dkim`, no central prefix.

docs/central-upgrade.rst
```diff
@@ -90,6 +91,62 @@ You'll be asked to confirm the removal of all dangling images. Agree by typing t
...
+   
+  .. tab:: Default mail server
+     .. tip:: While enabling DKIM on the default mail server will improve email delivery, we strongly recommend you use a :ref:`custom mail server <central-install-digital-ocean-custom-mail>` instead.
+
+ 
+     #. **Copy any existing DKIM files to a new location**.
+
+        .. code-block:: console
+
+         $ cd central
```
lognaturel, line 119: Should these all be `~/central`?

lognaturel (approved):

PR #1657, opened 2023-09-21
Add docs about enabling Single Sign-on in Central

### LINKED ISSUE #1633: Single Sign-on with OpenID Connect
This issue is to document new support for SSO (getodk/central#449).

- [x] Admins can change their own email address. If they change it to an email address that is not also associated with an identity managed by the IdP, they could lose access to the server. They would then have to turn off SSO, reprovision an admin account that matches the email address of an account managed by the IdP.

matthew-white:
Closes #1633.


docs/central-install-digital-ocean.rst
```diff
@@ -565,6 +565,104 @@ DKIM is a security trust protocol which is used to help verify mail server ident
...
+     OIDC_CLIENT_ID=
+     OIDC_CLIENT_SECRET=
+
+#. Build and restart all containers.
+
+   .. code-block:: console
+
+     $ docker compose build && docker compose stop && docker compose up -d
+
+Using Google as an Identity Provider
```
matthew-white, line 654: We discussed providing some guidance around how to set up an IdP for Google and maybe Azure. @lognaturel, would you be up for adding those instructions? I also think those instructions could come in a follow-up PR after release.

docs/central-install-digital-ocean.rst
```diff
@@ -565,6 +565,104 @@ DKIM is a security trust protocol which is used to help verify mail server ident
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+TODO Setting up the identity provider
+
+TODO Google Chrome profiles
+
+Using Azure as an Identity Provider
```
matthew-white, line 661: The brainstorming doc includes notes about Azure that I don't fully understand and didn't add here.

docs/central-install-digital-ocean.rst
```diff
@@ -565,6 +565,104 @@ DKIM is a security trust protocol which is used to help verify mail server ident
...
+Enabling SSO will not log out users who are already logged in. Users who are already logged in will be required to log in through the identity provider only after being logged out.
+
+Disabling SSO
+~~~~~~~~~~~~~
+
+It is possible to disable SSO by following the steps below. If there were users before SSO was enabled (if SSO was disabled, then enabled, then disabled again), users will be able to log into Central using their same password from before SSO was enabled. You can :ref:`reset users' passwords <central-users-web-reset-password>` after disabling SSO.
+
+To disable SSO:
+
+#. In ``.env``, set ``OIDC_ENABLED`` to ``false``. Clear ``OIDC_ISSUER_URL``, ``OIDC_CLIENT_ID``, and ``OIDC_CLIENT_SECRET``.
```
matthew-white, line 631: It's not strictly necessary to clear the latter three variables. We probably won't clear them on our dev servers. Our dev servers are unusual in that we will often toggle SSO on and off. However, that won't be the case for most users. I think it's easiest just to tell users to clear these variables rather than that they may clear them, but aren't required to. If they don't need to keep these values on the server, it's probably best to clear them, as some of them are sensitive.

docs/central-install-digital-ocean.rst
```diff
@@ -565,6 +565,104 @@ DKIM is a security trust protocol which is used to help verify mail server ident
...
+     OIDC_ENABLED=true
+     OIDC_ISSUER_URL=my-issuer-url
+     OIDC_CLIENT_ID=my-client-id
+     OIDC_CLIENT_SECRET=my-client-secret
+
+#. Build and restart all containers.
+
+   .. code-block:: console
+
+     $ docker compose build && docker compose stop && docker compose up -d
```
matthew-white, line 604: I noticed that #1655 added `docker compose stop` in various instructions, so I included it here.

docs/central-install-digital-ocean.rst
```diff
@@ -565,6 +565,104 @@ DKIM is a security trust protocol which is used to help verify mail server ident
...
+   .. code-block:: console
+
+     $ nano .env
+
+   .. code-block:: bash
+
+     OIDC_ENABLED=true
+     OIDC_ISSUER_URL=my-issuer-url
+     OIDC_CLIENT_ID=my-client-id
+     OIDC_CLIENT_SECRET=my-client-secret
```
matthew-white, line 598: According to the spec, the client ID and client secret are allowed to contain the characters `"` or `\`. However, our implementation does not support that. See https://github.com/getodk/central/pull/458#discussion_r1287901854. That said, I don't think it's useful to document that fact at this point. Alex mentioned that he hasn't seen either character in practice. Even if either character were used, I think the typical result would be a clear failure, because the Backend config file wouldn't be valid JSON.

docs/central-install-digital-ocean.rst
```diff
@@ -565,6 +565,104 @@ DKIM is a security trust protocol which is used to help verify mail server ident
...
+
+Two Accounts: Central and the Identity Provider
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When you enable SSO, users will use their account on the identity provider to log into Central. However, users will still have a Central account that is separate from the identity provider. A Central account is not automatically created for each account on the identity provider. Instead, a Central administrator will need to create a Central account for each user of the identity provider who should be allowed to log into Central.
+
+Central users will be able to change their display name shown in Central and to choose a different name from what is shown in the identity provider. However, because Central identifies users by their email address, most users will not be allowed to change their email address. Only a Central Administrator will be able to change the email address associated with a Central account. That will be needed if a user's email address changes within the identity provider. In that case, an Administrator will need to manually change the user's email address within Central to match their new address in the identity provider.
```
matthew-white, line 611: Email change is one of two special cases mentioned in the brainstorming doc. I didn't add the other case ("Email addresses change hands") because I'm not completely sure how to describe it.

docs/central-install-digital-ocean.rst
```diff
@@ -565,6 +565,104 @@ DKIM is a security trust protocol which is used to help verify mail server ident
...
+
+Logout is not centralized, which means that when a user logs out of Central, that will not log them out of the identity provider. Conversely, when a user logs out of the identity provider, that will not log them out of Central. If a user logs out of Central, then goes to log back in, they may find that login is nearly instantaneous if they are still logged into the identity provider. That is, they may find that they are not required to log into the identity provider again in order to log into Central.
+
+Enabling SSO in an Existing Installation
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+It is possible to enable SSO for an existing Central installation, even if the installation has existing users. Follow the instructions above to enable SSO. Because Central identifies users by their email address, the email address associated with a Central account must match the address in the identity provider. If the email address does not match, the user will not be able to log in.
```
matthew-white, line 620: I think this is all true. We've been able to toggle SSO on and off on dev servers. There was related discussion at https://github.com/getodk/central-backend/pull/910#discussion_r1269624813 that wasn't fully resolved. However, I understand the flow better now than when I left those comments.

docs/central-install-digital-ocean.rst
```diff
@@ -565,6 +565,104 @@ DKIM is a security trust protocol which is used to help verify mail server ident
...
+   .. code-block:: console
+
+     $ docker compose build && docker compose stop && docker compose up -d
+
+Using Google as an Identity Provider
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+TODO Setting up the identity provider
+
+TODO Google Chrome profiles
```
matthew-white, line 659: This TODO is related to the discussion here: https://github.com/getodk/central/issues/449#issuecomment-1720768753. It was also discussed in Slack. I added some notes to the brainstorming doc drawing from Slack:

> If you want to go between Central accounts when using Chrome and Google as the IdP you have to use separate browser profiles. Weirdness around when the browser uses the same identity provider.

I didn't completely follow that discussion at the time. @lognaturel, would you mind adding docs about this case?

docs/central-install-digital-ocean.rst
```diff
@@ -565,6 +565,104 @@ DKIM is a security trust protocol which is used to help verify mail server ident
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+It is possible to enable SSO for an existing Central installation, even if the installation has existing users. Because Central identifies users by their email address, the address associated with each Central account must match the address of the corresponding account on the identity provider. If the email address does not match, the user will not be able to log in.
+
+Enabling SSO will not log out users who are already logged in. Users who are already logged into Central will not be required to log into the identity provider until they are logged out of Central.
+
+Disabling SSO
+~~~~~~~~~~~~~
+
+It is possible to disable SSO by following the steps below. If there were users before SSO was enabled (if SSO was disabled, then enabled, then disabled again), users will be able to log into Central using their same password from before SSO was enabled. You can :ref:`reset users' passwords <central-users-web-reset-password>` after disabling SSO.
```
matthew-white, line 627: Rather than disabling SSO, could they use the command line to create a new Central administrator with the correct email address?

matthew-white (commented):
@lognaturel, this is now ready for you to take a look!

docs/central-install-digital-ocean.rst
```diff
@@ -565,6 +565,104 @@ DKIM is a security trust protocol which is used to help verify mail server ident
...
+
+Central is not compatible with every identity provider. Central uses the OpenID Connect (OIDC) protocol. When SSO is enabled in Central, Central does not manage passwords, but it still identifies users using their email address. Central assumes that the identity provider verifies email addresses, requiring users to prove ownership of the email address they specify. If that is not the case, then do not enable SSO in Central.
+
+If SSO is enabled, some features in Central will stop working. OData feeds will not be accessible in tools like Power BI or Excel. The API will not be available, which means that pyODK and ruODK will not be able to fetch data from Central. In the future, we hope to provide a way to access OData and the API when SSO is enabled.
```
lognaturel, line 579: I think we've tried to move away from "in the future" types of statements. Could we instead say "Currently, OData feeds..."

docs/central-install-digital-ocean.rst
```diff
@@ -565,6 +565,104 @@ DKIM is a security trust protocol which is used to help verify mail server ident
...
+
+To enable SSO in Central, you will need to provide information from your identity provider, specifically the issuer URL, client ID, and client secret. If you use Google Workspace or Azure Active Directory, you can follow the instructions below to find this information.
```
lognaturel, line 581: Let's add these sections as part of a separate PR after release as you suggest below.

lognaturel (commented):
Let's make adding the provider-specific instructions a separate issue and pull that out of here.

Other than that, looks good!

There was a note on the criteria about what happens when the OIDC env vars are invalid. I think it's ok not to document that.

docs/central-install-digital-ocean.rst
lognaturel: Yes, true. You could change that if you want.

lognaturel:
The more I think about this, the less I like it being in the Digital Ocean install instructions. Mostly this is just helpful information about how Single Sign-On works that any user would benefit from. How about making it its own page? `central-sso` can be the slug, "Single Sign-On in Central" can be the title, and I'd put it in "Managing ODK Central" because it's an admin consideration.

lognaturel:
Continued in #1661

PR #1661, opened 2023-09-25
Add docs about enabling Single Sign-on in Central

lognaturel:
Replaces https://github.com/getodk/docs/pull/1657, closes https://github.com/getodk/docs/issues/1633

Looking at it again, I noticed a key piece of information was missing: the redirect URL! I have added that.

I also added links to different identity providers' instructions. I believe they are all clear enough that we don't need to maintain our own. We can decide to do more depending on user feedback.

In https://github.com/getodk/docs/pull/1657#issuecomment-1732098592 I mentioned splitting this off into its own page. I played around with it but I realized that it's a separate question that applies to a lot of existing resources. I decided not to address it now.


lognaturel:
@ktuite I think this just needs a last sanity check to make sure all the info is there and that there aren't any glaring mistakes. Thanks!

ktuite (approved):
Looks good to me! 

Regarding the comment about moving it to a different SSO section, I think that would be nice to do someday. I don't know how many people are going to be looking for SSO in the docs right away, but it feels buried right now since it doesn't show up in the sidebar index... but maybe that's fair -- it is a customization option that is at a similar level to setting up custom certificates, databases, etc.

lognaturel:
> it is a customization option that is at a similar level to setting up custom certificates, databases, etc.

Right! I go back and forth between wanting to break all of these out and feeling like it's nice to have a more linear path through. It really depends on whether you're coming in for the first time or looking to update an existing server.

PR #1663, opened 2023-09-26
remove autogenerated rst files

sadiqkhoja:
Not sure how these files were added back by https://github.com/getodk/docs/pull/1645 even though these are git-ignored


lognaturel:
Fascinating! That makes sense -- I didn't have the .gitignore locally in that branch so the changes were pushed. Oof, that's something to be really careful about if using gitignore to protect credentials, for example.

lognaturel (approved):

PR #1665, opened 2023-09-28
Compress top level TOC, improve cli code blocks

### LINKED ISSUE #1649: Remove contributing to ODK docs and glossary
I question the usefulness of these docs. Leaving them in place means it's something that we have to keep up-to-date and is a visual distraction for users.

lognaturel:
Closes  #1649

The primary goal of this PR is to reduce the length of the TOC visible in the sidebar. The form design section in particular had gotten so long that it felt very disorganized to me. It also meant that when on a page low in the TOC, related TOC elements would always be off screen.

We also ended up with some pages that felt in the wrong place or didn't make sense at all (JavaRosa). Things should feel more organized now.

I'm doing this as a precursor to bringing some more Enketo web form docs in.

While doing that, I noticed that many CLI code blocks used the console type which is less attractive than `bash` so I changed it. I also fixed some stale Apiary links.


yanokwa:
Renames
* Integrating With Collect -> ODK Collect API (I'd also rename the file)
* ODK XForms spec -> ODK XForms Spec
* Launching External Apps -> External Apps (I'd also rename the file)
* Form Building -> Form Design
* training.rst -> additional-resources.rst

Remove Contributing and Glossary from top level

yanokwa (approved):
LGTM

PR #1666, opened 2023-09-29
Update URLs of API docs in user docs

sadiqkhoja:

Removed references to apiary

### DIFF
--- a/docs/central-backup.rst
+++ b/docs/central-backup.rst
@@ -18,7 +18,7 @@ Direct Backups via API
 
 .. warning::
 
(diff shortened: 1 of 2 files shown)


matthew-white:
Ah this is making me realize that we need to make this change in Frontend as well. There are a couple of links to Apiary. @sadiqkhoja, would you be up for making that change? I'd be happy to as well.

lognaturel (commented):
I snuck that in as a commit in https://github.com/getodk/docs/pull/1665!

Let's use relative links like I did there.

PR #1667, opened 2023-09-29
Replaced default search with Google

sadiqkhoja:
Default search is full of issues: https://github.com/sphinx-doc/sphinx/labels/html%20search

I think Google search is our best bet.

I have created a rst page to show the results but didn't add it in `toc` so build throws `checking consistency...  getodk-docs/docs/google-search.rst: WARNING: document isn't included in any toctree`. Is there any better way? or is it ok to ignore this warning. (if we go with this route then we need to add this page to ignore list of robot.txt)

Google provides many layouts, I am using "result only", which gives us ability to use custom search box and show results in a separate page. The reason to choose this approach is that default search-box of Google doesn't match our theme, especially it contains search button that takes significant space. And I am trying to avoid hacky customization.

Alternative approach is to use "overlay" layout, which shows search result in a modal box. That way we won't need a separate search rst page.

 nb: I have created "Programmable Search Engine" under my getodk account to generate `cx`. Is that okay or another account should be used?
https://programmablesearchengine.google.com/u/4/controlpanel/all

- Google search result doesn't respect dark/light switch - need to write css for that

image

### DIFF
--- a/docs/_templates/sidebar/search.html
+++ b/docs/_templates/sidebar/search.html
@@ -1,0 +1,4 @@
+<form class="sidebar-search-container" method="get" action="../google-search/" role="search">
+   <input class="sidebar-search" placeholder="Search" name="q" aria-label="Search" autocomplete="off">
+ </form>
+<div>
--- a/docs/google-search.rst
+++ b/docs/google-search.rst
@@ -1,0 +1,9 @@
+Search Results
+================
+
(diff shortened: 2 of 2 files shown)


sadiqkhoja:
Google search is full of ads in free version and paid plan is $5 per 1K searches, which is too expansive. Need to do something to the sphinx search

PR #1680, opened 2023-10-23
Add XLSForm tutorial

lognaturel:
Goal is for this to be as concise as possible while exposing someone who follows it to the major building blocks. We don't want to explain anything at this point, we just want someone to go through the process.


lognaturel:
Replaced by #1687

PR #1686, opened 2023-10-25
Add pyODK link

yanokwa:
I thought it'd be helpful to surface pyODK. Maybe it should be under the ODK Central API docs?


lognaturel (approved):
I initially wanted this to be linked from a section about API wrappers but it's a thing we maintain so I think it makes sense to surface more prominently after all.

PR #1692, opened 2023-10-26
Read API Spec location from ENV variable

sadiqkhoja:


alxndrsn:
Used in https://github.com/getodk/central-backend/pull/1039

PR #1694, opened 2023-11-02
Update Collect screens

lognaturel:
Updates Collect screenshots in project and form management pages.

I tried to keep the structural and content changes minimal but I did want to at least bring the existing content up-to-date.


alyblenkin (approved):
I think the new images look great! I would love to change the red circles and arrows because it always makes me think something is wrong when I see red. That being said, they are high-contrast and grab your attention.

PR #1700, opened 2023-11-13
Reduce width of device screenshots

lognaturel:
Reducing device screenshot width to 365px from 400px means two can go side-by-side with a screen width of 1500px. Prior to this change, screenshots would always appear vertically stacked on most laptop screens.

### DIFF
--- a/docs/_static/css/custom.css
+++ b/docs/_static/css/custom.css
@@ -28,7 +28,7 @@
 }
 
 .device-screen-vertical {
-    width: 400px;
+    width: 365px;
 }
 
 table.schema-table-wrap {


alyblenkin:
@lognaturel Is there a way for me to see this change visually? I also don't need to because it makes sense!

lognaturel:
I just realized that you can get the artifacts from the build the same way you do in Collect! You should be able to "Show all checks", "Details", and then click on the artifacts tab. In this case, https://output.circle-artifacts.com/output/job/ae813766-c860-4d02-90ab-77876aac1109/artifacts/0/docs/_build/dirhtml/collect-filling-forms/index.html is a good page to see the change on.

lognaturel:
I wonder whether I can make this responsive... should be possible, I think! Let me take a look.

alyblenkin:
> I just realized that you can get the artifacts from the build like in Collect! You should be able to "Show all checks", "Details", and then click on the artifacts tab. In this case, https://output.circle-artifacts.com/output/job/ae813766-c860-4d02-90ab-77876aac1109/artifacts/0/docs/_build/dirhtml/collect-filling-forms/index.html is a good page to see the change on.

Amazing! I would not have guessed that link. I tried the first few and thought I couldn't view it the same way as Collect, but this is great news. And showing the images side-by-side looks great.

alyblenkin (approved):

lognaturel:
I briefly tried to think of a good way to make the size responsive but I don't have a good solution off the top of my head so let's merge this initial improvement and come back to it. My ideal would be for images to be 49% of the container width until a certain threshold when they become 100% and stack.

PR #1701, opened 2023-11-13
Improve wording around finalization

lognaturel:
@alyblenkin mentioned that "mark as finalized" doesn't make as much sense now with the Finalize/Send button.


lognaturel:
Thanks @alyblenkin!

PR #1729, opened 2024-01-11
Add Entities tutorial

lognaturel:
This is a detailed tutorial in the style of https://docs.getodk.org/xlsform-first-form/ for building an entity-based workflow involving two forms and one entity list. It shows entity creation and updates. This is the example developed in this video: https://m.youtube.com/watch?v=919SIU41UQA

[See it built](https://output.circle-artifacts.com/output/job/dee502f9-256b-49a6-9361-7a5e4fb417b6/artifacts/0/docs/_build/dirhtml/tutorial-community-problems/index.html)

Questions for review:
- Does it need more visuals/screenshots?
- Is it detailed enough to be followed by someone who hasn't created an entity-based form before?
- How is the level of explanation? Too much for a tutorial? Not enough to be transferable/useful?


alyblenkin:
@lognaturel the video doesn't seem to be working. Does it work for you @yanokwa?

lognaturel:
I posted the Studio link. Updated it to the public one, does that work?

alyblenkin:
> Does it need more visuals/screenshots?

We could add more visuals to the doc, but I don't think it is missing anything right now. When you started talking about the workflow, I thought we could have a little diagram so it is more visual, but this is not a dealbreaker. And we could always add this later!

> Is it detailed enough to be followed by someone who hasn't created an entity-based form before?

There is enough detail, but the next step is for me to try it! I was going to do this next week so I could finish up a few other things with the website and scheduling interviews, but I can pause that if you think we need to do this first. 

> How is the level of explanation? Too much for a tutorial? Not enough to be transferable/valuable?

The video intro is helpful. The workflow is easy to understand and relatable. I like how you frame this behind-the-scenes video and that it will be long and you'll make mistakes. It lowers the barrier and makes it more approachable. It will likely make someone feel less intimidated to test Entities themselves. The intro and framing in the doc look good too. I don't think we need more explanation up front. We should add a link to this video, but maybe that's your plan!

Other notes:
- I only watched some of it as you said, but the first point of confusion for me was the titles of the resources because it was hard to see how they were connected. Initially, I thought the doc was about reporting problems in the ODK community rather than the tutorial scenario. Of course, this becomes very clear when you start reading and see the tutorial pill, but it might be good to link the titles more so I know they are connected. People might have multiple tabs open or return to them later and need help understanding the connection between them. Maybe something like "Entities: building a form to report & resolve problems" or maybe "Incident report & resolution with Entities" as we had for the cover image.
- Is the idea that we would have a shorter tutorial following the docs exactly without the longer explanation (just cut down your video?). Now that I watched the end bit, you're asking for feedback!
- Do you think we should link the tutorial to the YouTube description?It depends on the sequence of when we release them.?

lognaturel:
> I thought we could have a little diagram so it is more visual

Yes, please! If you have a chance to do that before Tues, let's include it, otherwise let's add it later.

> We should add a link to this video

You think so? It feels a little rough for that to me but maybe we could put it somewhere in the text with an explanation.

> titles of the resources

Yes, good point, I'll think more about unifying that. I think we should probably drop the "incident" word since it's not used inside any of the resources.

> that we would have a shorter tutorial

I was thinking we can put out the text and the behind the scenes video on Tuesday and say we could do a more polished video if there's demand. I'm not sure how helpful it would really be. I also think it depends on your experience when trying it out from text.

> Do you think we should link the tutorial to the YouTube description?

Yes, definitely. Let's get both out on Tuesday.

alyblenkin:
> Yes, please! If you have a chance to do that before Tues, let's include it, otherwise let's add it later.

No problem, I can create that diagram today. Do you still want me to test it as well?

> We should add a link to this video
> You think so? It feels a little rough for that to me but maybe we could put it somewhere in the text with an explanation.

I think it's OK if it's a little rough. I think it would be helpful for discoverability. We could put it at the end of the doc and make it clear that it's a behind-the-scenes video. 

Future thinking: it would be really cool for people could indicate if the found it helpful or not, like GitHub.
Screenshot 2024-01-15 at 9 10 05 AM

alyblenkin:
First draft! Let me know what you think Entity updates workflow

alyblenkin:
I tested the tutorial, but I got a little confused. I made some notes in our 1:1 doc to come back to tomorrow. I'm sure I'm missing one small thing that's not allowing me to test the entire workflow.

lognaturel:
@alyblenkin and @yanokwa this is ready for final review.

alyblenkin:
@lognaturel Looks good - I like the new additions. 

Did you decide not to call it "Entities updates tutorial"?

lognaturel:
I should have flagged that. In the sidebar, I think "Entities tutorial" is the most helpful at the moment given the current organization. In general, the tutorial is not very specifically focused on updates, it's really about building an end-to-end workflow supported by Entities. Maybe I should still title the page "Entity updates tutorial".

The reason I had originally called the document `Capture and resolve a community's problems with Entities` was with the thought that we might have several workflows written up this way. But of course it's a mouthful.

I will think a bit more about it and also consult @yanokwa

alyblenkin:
> I should have flagged that. In the sidebar, I think "Entities tutorial" is the most helpful at the moment given the current organization. In general, the tutorial is not very specifically focused on updates, it's really about building an end-to-end workflow supported by Entities. Maybe I should still title the page "Entity updates tutorial".

Ah yes, that makes sense! That's actually why I used the word 'workflow' in the graphic as well. I also just spoke with a user who is very keen to see more resources like this and how entities work with different types of workflows, so I could see a series being very helpful to folks.

yanokwa (approved):

PR #1730, opened 2024-01-16
Central build sometimes fails on 1 GB machines

yanokwa:
See https://forum.getodk.org/t/odk-central-installation-error/44637 as an example.


lognaturel:
This is clearly an issue and it seems like the latest release has made it even worse. Most servers don't need more than 1Gb to *run*, though, only to build. I'd like to see us call that out specifically. Do you want to modify accordingly or do you want me to do a take, @yanokwa?

PR #1743, opened 2024-02-08
Add rating feature to page footer

lognaturel:
Screenshot 2024-02-08 at 10 13 54 AM

@alyblenkin helped me get a reasonable first pass for styling, we'll come back for more polish.


yanokwa (approved):
I don't love that it pops up another page, but I'm willing to give it a shot for 30 days. If we have useful feedback, great. If not, we should kill it.

PR #1759, opened 2024-03-22
Use correct image

yanokwa:
Sigh


lognaturel (approved):
Bad reviewer! Sorry. I noticed that and my brain decided you were changing both rather than making a mistake.

PR #1761, opened 2024-03-24
Fix response json example for http 200 on entity deletion

spwoodcock:

- For the [Deleting an Entity](https://docs.getodk.org/central-api-entity-management/#deleting-an-entity) section in the Central API docs, the response example for a HTTP 200 is incorrect.
- The docs state that it is a JSON response, with a string value {'message': 'success'}:

image

- In fact, on testing, the actual response is a JSON with bool value `{'success': True}`.
- I updated the docs to reflect this.

#### Context

- I wasn't sure if the OpenAPI docs are auto-generated somehow, but I made a PR anyway to update the yaml manually.
- Feel free to close this PR and update the upstream if that is the case.

### DIFF
--- a/docs/_static/api-spec/central.yaml
+++ b/docs/_static/api-spec/central.yaml
@@ -10290,7 +10290,14 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/Success'
+                required:
+                - success
+                type: object
+                properties:
+                  success:
+                    type: boolean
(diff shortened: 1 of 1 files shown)


lognaturel:
Thanks! Indeed the API docs are managed in the Central backend repo. I've made that clearer in this readme and made the fix at https://github.com/getodk/central-backend/pull/1121

PR #1764, opened 2024-03-29
Removed the page that describes ODK Build

grzesiek2010:
ODK Build has been shut down so it doesn't seem to make sense to keep its documentation.


yanokwa:
I think we should leave the docs for another year? People might be self-host Build and it might be helpful? It's not hurting anyone.

PR #1775, opened 2024-04-08
Update date

yanokwa:
First date of publication

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -23,7 +23,7 @@ import openapi
 # -- Project information -----------------------------------------------------
 
 project = 'ODK'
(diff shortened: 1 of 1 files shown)


lognaturel:
What are the goals for this change? I see from https://webmasters.stackexchange.com/a/8249 and other sources that this doesn't matter legally. But for better or worse, I use it as a sign of "freshness." If the docs start saying copyright 2017, to me that would imply the haven't been changed since 2017. I think we should consider "2017-present" or "2017-2024" if the goal is to give a sense of longevity.

yanokwa:
Goal is to not to ever change it again, but I guess you are right. Let's keep it at 2024.

PR #1774, opened 2024-04-05
French translation of entities tutorial

mathieubossaert:
closes #1763 

#### What is included in this PR?

It includes translation of the tutorial, googlesheets forms ans screenshots.
Also include the working syntax of the choice-filter with current() function (https://github.com/getodk/docs/issues/1772) 
The french version was tested, fixed,  and approved today by 10 french geomaticians from "Conservatoires d'espaces naturels" ;-)
Thanks a lot for the clear original version !

PR #1779, opened 2024-04-09
French entities tutorial

lognaturel:
Replaces #1774 

@mathieubossaert, I wasn't able to push to your branch for reasons I don't totally understand so I branched off of it.

The biggest things I did were:
- ignore the file for spell check
- switch the Google Sheets links to ones in the ODK directory for documentation
- add the social image

Thanks again, this is fantastic! Let me know if you have any last few comments and then we can merge.

[Here it is rendered](https://output.circle-artifacts.com/output/job/4da88bcf-d1cd-468b-9899-52c33ce3cb57/artifacts/0/docs/_build/dirhtml/translations/fr/tutoriel-signalement-problemes/index.html)

@alyblenkin I did the social image at https://www.figma.com/file/WQtlBqXL6ghsEpqTHtQJTN/Untitled?type=design&node-id=1%3A100&mode=design&t=4AIqKDc7CAuf5rum-1 if you want to make any tweaks. French is much more verbose but I think I made it work.


mathieubossaert:
Everything seems to be fine 🙂

lognaturel:
@yanokwa I'm tagging you for review and merge. The most controversial thing here is the way we introduced translations. I chose to do it in an informal way for now.

Sphinx does have a way to manage translations: https://www.sphinx-doc.org/en/master/usage/advanced/intl.html It doesn't feel worth changing our build pipeline for at this moment. I think that even if we do some more translation, it will likely be one-offs like this for now and these are easier to manage as full documents than paragraph by paragraph. Once we have a process for more translations in place, we could revisit this approach.

lognaturel:
@yanokwa Please don't squash so @mathieubossaert's commits are in!

yanokwa (approved):

PR #1798, opened 2024-05-20
Add Introduction to Entities

lognaturel:
This adds a one-stop-shop document describing Entities conceptually and providing references to other parts of the documentation.

It's organized as Q&A to hopefully make it relatively easy to scan. I considered collapsing the questions and having users expand the ones they want to read but that makes it harder to search the page which I think can be helpful with a document like this.

The questions reflect questions from the forum, Entity office hours, and other conversations with users.

For now, let's introduce this as an orphan page so @yanokwa can do another round of review and we can share with Insiders.

Then, I think it should go at the top of the left table of contents:
Screenshot 2024-05-20 at 3 50 21 PM

That feels like the best spot to me: it's general content that's important to understanding what's possible with ODK. I don't see an obvious other alternative. I considered adding an Entities section on the same level as Collect and Form design but that feels worse to me. We could do a deeper rework and put it under a "concepts" section but that feels like more than we want to take on now.

### DIFF
--- a/docs/entities-intro.rst
+++ b/docs/entities-intro.rst
@@ -1,3 +1,5 @@
+:orphan:
+
 .. spelling:word-list::
 
     Airtable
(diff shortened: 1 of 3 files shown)


yanokwa (approved):

lindsay-stevens:
~~ haiku review ~~
in case it slipped through
take note of the small TODO
on line 82
> TODO: focus on list of last state

lognaturel:
Quite possibly the best review I have ever received, @lindsay-stevens, you've outdone yourself. 😂

Yes, we should address that. 😄

PR #1799, opened 2024-05-23
Entities intro improvements

yanokwa:
- [ ] I'm not convinced by the tree example in the flow section
- [ ] Case management section still needs work


docs/entities-intro.rst
```diff
@@ -42,45 +42,41 @@ Entities can also be helpful for sharing data that rarely changes between multip
...
 
-Even in fields where case management is common, there is often a need to support other kinds of workflows within the same tool and it can be awkward to use the word "case" in those contexts, especially when referencing concrete entities such as participants or hospitals. Our goal is to let you define Entity Lists that make sense in your context and to allow you to use and connect them in ways that best support your workflow.
+We use the word "entities" because Entities are much more flexible. They don't impose any requirement that you to open or close a case. Also, even in domains where case management is common, there is often a need to support other kinds of workflows within the same tool and it can be awkward to use the word "case" in those contexts, especially when referencing real things like trees, people, or cities. Our goal is to let you define Entity Lists that make sense in your context and to allow you to use and connect them in ways that best support your workflow.
```
lognaturel, line 66: I think "much more flexible" isn't quite right. We use the word "Entities" because it's a more generic framing.

"They don't impose any requirement that you to open or close a case." Not sure how strong this is.

lognaturel (approved):
Couple of small suggested changes but let's get it out there!

PR #1804, opened 2024-06-02
Add helpful link and fix formatting

yanokwa:


docs/form-datasets.rst
```diff
@@ -2,6 +2,9 @@
 Form Datasets
 ************************
 
+.. note::
+  "Datasets" and "Entity Lists" are synonyms. We typically use "Dataset" when communicating with developers and "Entity List" when communicating with everyone else.
```
lognaturel, line 6: This is not right. This page is generically about any kind of dataset that can be accessed from a form — choice lists, attached files, entity lists. Probably it’s due for a rework.

lognaturel (approved):

PR #1812, opened 2024-06-19
Add details about attaching Entity Lists

lognaturel:
Clarify that multiple Entity Lists can be attached to a form, clarify details around how to use select one and csv-external.

### DIFF
--- a/docs/entities-intro.rst
+++ b/docs/entities-intro.rst
@@ -188,11 +188,6 @@ First, attach the Entity List(s) that you want to access Entities from in your f
 
 You can access a specific Entity's properties using a :ref:`lookup expression <referencing-values-in-datasets>`. If you've used CSV form attachments or looked values up in a choice list before, looking values up in an Entity List works exactly the same way.
 
-You can attach as many Entity Lists to the same form as you need to support your workflow.
-
-.. image:: /img/entities-intro/multiple-entity-list-attachments.*
-  :alt: Diagram showing multiple Entity Lists attached to a single follow-up form.
-
 .. seealso::
     * :ref:`Looking up values in a list <referencing-values-in-datasets>`
     * :doc:`Community reporting tutorial <tutorial-community-reporting>`
--- a/docs/entities-intro.rst
+++ b/docs/entities-intro.rst
@@ -178,15 +178,20 @@ Mechanics
 How do I access Entities from my forms?
 ---------------------------------------
 
-First, attach the Entity List you want to access Entities from in your form definition either using `select_one_from_file` or `csv-external`.
+First, attach the Entity List(s) that you want to access Entities from in your form definition using ``select_one_from_file`` or ``csv-external``:
 
-If you want the user to be able to select an Entity from a list, you can use a :ref:`select_one_from_file <select-from-external-dataset>` question with the name of your Entity List followed by `.csv`. For example, if your Entity List is named ``trees``, you would have a ``select_one_from_file trees.csv`` question. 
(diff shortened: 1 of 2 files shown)


docs/entities-intro.rst
```diff
@@ -178,15 +178,20 @@ Mechanics
 How do I access Entities from my forms?
 ---------------------------------------
 
-First, attach the Entity List you want to access Entities from in your form definition either using `select_one_from_file` or `csv-external`.
+First, attach the Entity List(s) that you want to access Entities from in your form definition using ``select_one_from_file`` or ``csv-external``:
```
tobiasmcnulty, line 181:
```suggestion
First, attach the Entity List(s) that you want to access Entities from in your form definition using ``select_one_from_file`` or ``csv-external``, as described below. Note that you can attach multiple Entity Lists to a single form via these methods, and they do not need to be listed on the ``entities`` tab in the XLSForm (only the Entity List that is updated via ``save_to`` needs to be included on this tab).
```

tobiasmcnulty (commented):
Thanks for these updates! The changes look good to me.

Anecdotally, I searched this page for the word "multiple" when attempting to figure out how to do this, so I tried to add a sentence with that keyword. Take it with a grain of salt and feel free to edit/delete as you see fit, of course!

lognaturel:
Thank you! I also search by keyword and this is a great call. The point about only needing to reference an Entity List that's added to or updated is important enough that I took it out of parentheses.

docs/entities-intro.rst
```diff
@@ -178,15 +178,20 @@ Mechanics
...
+* If you want to look up Entities using a user-provided value such as a unique ID scanned from a barcode, entered manually, or looked up from another Entity's property, you can attach your Entity List with :ref:`csv-external <form-datasets-attaching-csv>`. For example, if your Entity List is named ``trees``, you would create a form field of type ``csv-external`` with name ``trees``.
 
-Once a specific Entity is identified, you can look up its properties using a :ref:`lookup expression <referencing-values-in-datasets>`. All of this works exactly the same way as it does with CSV form attachments.
+You can access a specific Entity's properties using a :ref:`lookup expression <referencing-values-in-datasets>`. If you've used CSV form attachments or looked values up in a choice list before, looking values up in an Entity List works exactly the same way.
+
+You can attach as many Entity Lists to the same form as you need to support your workflow.
+
+.. image:: /img/entities-intro/multiple-entity-list-attachments.*
```
yanokwa, line 193: I don't think the image adds much? Might even be confusing because it's not immediately obvious that these are different registration forms and different entity lists.

lognaturel: Fair

docs/entities-intro.rst
```diff
@@ -178,15 +178,15 @@ Mechanics
...
 
-If you want to look up Entities using a user-provided value such as a unique ID scanned from a barcode or entered manually, you can attach your Entity List with :ref:`csv-external <form-datasets-attaching-csv>`.
+* If you want to look up Entities using a user-provided value such as a unique ID scanned from a barcode, entered manually, or looked up from another Entity's property, you can attach your Entity List with :ref:`csv-external <form-datasets-attaching-csv>`. For example, if your Entity List is named ``trees``, you would create a form field of type ``csv-external`` with name ``trees``.
 
-Once a specific Entity is identified, you can look up its properties using a :ref:`lookup expression <referencing-values-in-datasets>`. All of this works exactly the same way as it does with CSV form attachments.
+You can access a specific Entity's properties using a :ref:`lookup expression <referencing-values-in-datasets>`. If you've used CSV form attachments or looked values up in a choice list before, looking values up in an Entity List works exactly the same way.
```
yanokwa, line 189: nitpick: s/looked values up/looked up values/

yanokwa (approved):
Some nit picks

PR #1826, opened 2024-07-08
Make it clear which properties can by used with which types of choices

### LINKED ISSUE #1823: Add note that marker-color and marker-symbol are only for points, not shapes or traces
https://github.com/getodk/docs/pull/1781#pullrequestreview-2015576633

grzesiek2010:
closes #1823 

Instead of adding a note (as mentioned in the issue), I've added a separate point for every type of choice (point, line, polygon). I think it makes more sense than a note but if you don't agree please let me know.


lognaturel (approved):
Good solution 👍

PR #1837, opened 2024-07-24
Consistently use search appearance rather than autocomplete

### LINKED ISSUE #1793: Consistently use search appearance rather than autocomplete
We probably avoided `search` to avoid confusion with `search()` but `search` is more intuitive than `autocomplete` and it's what we have in the XLSForm template. We can add a note that `autocomplete` is the same.

grzesiek2010:
closes #1793

As the title says it replaces occurrences of `autocomplete` with `search`.


latin-panda:
Hi @grzesiek2010, This has been sitting here for a little while, but it looks almost ready. What do you think it needs?

grzesiek2010:
> What do you think it needs?

Just a review. It was marked as ready for review 2 years ago, but nobody reviewed it. If you have time, feel free to take a look. There are some conflicts, so I’ll fix those first.

latin-panda (approved):
It looks good to me! The notes are clear

PR #1848, opened 2024-08-27
Update style of new range screenshots

lognaturel:
@grzesiek2010 refreshed screenshots for `range` in https://github.com/getodk/docs/pull/1846. This puts them in our [docs template](https://www.figma.com/design/Hbk9VpMD9KSlLXHmhT3jwv/Collect-frame-for-docs?node-id=549-9&t=9KseMozGrZ12QXcc-0). It also adds a brief description of how to use the screenshot template to the README.


grzesiek2010 (changes requested):
`range-widget-rating.png` seems to have not been updated.

lognaturel:
> range-widget-rating.png seems to have not been updated.

Whoops, fixed, thank you.

> I've also noticed that you exported the whole template (there is a background) instead of exporting the phone layer.

Yes, that's the goal! You can see it in action at https://docs.getodk.org/collect-forms/ for example. The subtle container gives more weight to the images and matches slides and other assets that we now produce. I tried to make those steps clear in the readme, let me know if you think any updates are needed.

grzesiek2010:
> Yes, that's the goal! You can see it in action at https://docs.getodk.org/collect-forms/ for example. The subtle container gives more weight to the images and matches slides and other assets that we now produce. 

Ahh, I see. But one more thing... your images have those white corners:
image but the ones you referred to don't 
https://docs.getodk.org/collect-forms/

alyblenkin:
This is a weird thing that happens sometimes because a layer might be turned off (I think the body layer might be hidden here). @lognaturel If you give me access to the file I can change it.

lognaturel:
Interesting, I thought it was an intentional update you made. I can’t give you edit access because of my account level but I can copy again from the main doc. I didn’t want to link to that one since it has so much in it! Or you can make a clean copy for us to use here if you prefer.

alyblenkin:
The template that I made in my [figma file ](https://www.figma.com/design/KFi9hIQdrRgo0rLqIC7Sg8/ODK-user-flows?node-id=3222-23559&t=jjHH1X9dZWJOSzHl-1)is still the same, if you want to just copy that over?

lognaturel:
I was going too quickly and thought @grzesiek2010 was talking about the outer rounded corner rather than the weird extra white. I unfortunately can barely see that white because I use low contrast settings.

I've fixed it by copying the original, thanks @alyblenkin. Is there a way to lock something down so that this doesn't happen? It has happened before and the only fix I found was to copy over again. I don't have any sense of what I'm doing to trigger it, sorry.

Should be fixed now.

grzesiek2010 (approved):

alyblenkin:
> I've fixed it by copying the original, thanks @alyblenkin. Is there a way to lock something down so that this doesn't happen? It has happened before and the only fix I found was to copy over again. I don't have any sense of what I'm doing to trigger it, sorry.

Yes we can lock the layer! I just did it in my file, so you could copy that one or lock the same layer in your version. I'm not sure why it was changing. Layers can act up some times.

Issue #1855, opened 2024-09-23
Update user docs for Central v2024.3


matthew-white:
- [x] Offline entities (@lognaturel TODO)
  - Info about submissions that are held for processing https://github.com/getodk/central/issues/821
- [ ] #1873 (@issa-tseng TODO)
- [x] Upgrade instructions (@lognaturel TODO)
   - If you use forms that create or update entities, you'll also need to update Collect (because the entity spec version will get bumped) - getodk/central#692
- [x] ~Entity properties with same name but different capitalization can't be created anymore~

lognaturel:
https://github.com/getodk/docs/issues/1873 is in the backlog, we can close this.

Issue #1786, opened 2024-04-30
Update user docs for Central v2024.2


matthew-white:
- [x] Changes to ops (@lognaturel TODO)
  - In the docs [here](https://docs.getodk.org/central-install-digital-ocean/#using-a-custom-ssl-certificate), we suggest rebuilding `nginx`. But @lognaturel says that with .2, `nginx` will just need to be restarted: "The logic to put the certs in the right place and everything is in the command that’s run at container start time." In general: "settings changes should not require rebuilds."
  - @lognaturel: "I think we’ll want to document for folks who use their own infrastructure that we now publish images."
  - #1858
- [x] Upgrade instructions
  - #1858
- [x] S3 configuration. ~See notes at getodk/central#700~
  - #1859

PR #1858, opened 2024-09-25
Add Central v2024.2 upgrade instructions

lognaturel:
See #1855

### DIFF
--- a/docs/central-upgrade.rst
+++ b/docs/central-upgrade.rst
@@ -99,10 +99,10 @@ Upgrading to Central v2024.2
 
 There are no required special steps related to this upgrade. However, there are some optional changes that you may want to know about or opt into.
 
-1. Configuration no longer requires rebuilding
+1. Changes to .env no longer require rebuilding
 **********************************************
 
-If you make changes to your `.env` file to configure Central as described in the :doc:`setup instructions <central-install-digital-ocean>`, you no longer have to rebuild Central for the changes to take effect. Instead, a stop and restart is now enough:
+If you make changes to your `.env` file to configure Central as described in the :doc:`setup and configuration instructions <central-install-digital-ocean>`, you no longer have to rebuild Central for the changes to take effect. Instead, a stop and restart is now enough:
 
 .. code-block:: bash
 
--- a/docs/central-upgrade.rst
+++ b/docs/central-upgrade.rst
@@ -109,10 +109,10 @@ If you make changes to your `.env` file to configure Central as described in the
    $ docker compose stop
    $ docker compose up -d
 
-2. File data can be stored in S3-compatible storage
-***************************************************
+2. Files can be stored in S3-compatible storage
+************************************************
 
-By default, Central stores all of its data in a database, including files like images attached to submissions. If you have or plan on collecting a lot of files, you may prefer to store these in separate storage to reduce load on the database and possibly reduce hosting costs. See instructions on how to opt in.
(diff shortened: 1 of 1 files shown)


matthew-white:
Just wanted to check, is this PR ready for review / to be merged?

lognaturel:
We'll want to add a link to https://github.com/getodk/docs/pull/1859/files#diff-d1fd5d1b68bb4a3d5db94b2138eaf7ab19c8f8eeee5d9c69a6745a58d5e6eed1R330

Maybe you could plan on doing that after #1859 is merged either by pushing a commit here or doing another PR?

lognaturel:
I added the link because I realized I wanted to change some wording too. The build will be broken but I think we can squash and merge after #1859 and follow up if there are any issues on `master`.

docs/central-upgrade.rst
```diff
@@ -91,6 +92,33 @@ You'll be asked to confirm the removal of all dangling images. Agree by typing t
...
 --------------------------------------
 
+.. _central-upgrade-2024.2:
+
+Upgrading to Central v2024.2
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+There are no required special steps related to this upgrade. However, there are some optional changes that you may want to know about or opt into.
+
+1. Configuration no longer requires rebuilding
```
matthew-white, line 102: Is there still some configuration outside of changes to `.env` that does require rebuilding? If so, it might be clearer for this to say "Changes to .env no longer require rebuilding".

matthew-white (approved):
Looks great. :+1: Exciting to read about these improvements!

docs/central-upgrade.rst
```diff
@@ -99,10 +99,10 @@ Upgrading to Central v2024.2
 
 There are no required special steps related to this upgrade. However, there are some optional changes that you may want to know about or opt into.
 
-1. Configuration no longer requires rebuilding
+1. Changes to .env no longer require rebuilding
```
matthew-white, line 102: I see that the spellcheck is complaining about ".env". Does it need to be enclosed in backticks? I'm happy to make that change if it'd help.

matthew-white:
Looking at the PR description and at edits to #1855, I'm realizing that you were probably using #1855 to prepare these docs. I actually intended that one to be the tracking issue for .3. I had a separate issue for .2 at #1786, but I moved some notes from #1786 to #1855 when it looked like they weren't going to be relevant to the .2 release. I just now moved some notes back from #1855 to #1786 (e.g., about S3). Sorry, my moving things around was probably a source of confusion. 😬 I don't think there's any issue, I just wanted to make a note of the two tracking issues.

One bullet that #1786 has that I think hasn't been captured yet is that users will no longer need to rebuild after configuring a custom cert. My understanding is that that's true of .2 even though it isn't just a change to `.env`, though maybe that's not right. But I think it makes sense to document that after this PR. I've copied the bullet to #1855 so we don't lose track of it.

docs/central-upgrade.rst
```diff
@@ -10,6 +10,7 @@ Start by reviewing upgrade notes for all versions between your current version a
 Upgrade notes
 -------------
 
+* :ref:`Central v2024.2 <central-upgrade-2024.2>`: configuration no longer requires rebuild, consider using S3 storage
```
matthew-white, line 13: Might be worth updating this line to match the recent changes to below?

```suggestion
* :ref:`Central v2024.2 <central-upgrade-2024.2>`: some configuration no longer requires rebuild, consider using S3 storage
```

lognaturel: I matched the section title, hopefully that seems ok.

PR #1860, opened 2024-09-25
make autobuild: use docker

alxndrsn:
Replaces requirement for a local, functioning python env with requirement for a local, functioning docker env.


lognaturel:
> Replaces requirement for a local, functioning python env with requirement for a local, functioning docker env.

Hah! Can we please make this additive rather than a replacement? Maybe `make docker-autobuild` or just `make docker`?

alxndrsn:
Is `autobuild` idiomatic?  What about having `make dev-local` and `make dev-docker`?

lognaturel:
It's from https://pypi.org/project/sphinx-autobuild/ Given it's been like this for years, it's in a lot of our fingers. We could add a `dev-local` "alias" if you want but let's keep `autobuild` as well.

alxndrsn:
:+1:

How about `make autobuild-docker`?  That should keep muscle memory for `make autobuild` working ok, but also make the docker option discoverable without needing extra documentation.

lognaturel:
Love it, thanks!

PR #1883, opened 2024-11-15
Add info to ensure `pgrowlocks` is installed for S3 on custom DBs

spwoodcock:
### Issue

- I was setting up external S3 for Central today.
- I wanted to test manual syncing using `node lib/bin/s3.js upload-pending 1`.
- The `pgrowlocks` Postgres extension was not installed in the db, causing an error.

### Solution

- Added the extension to the [Using a custom database](https://docs.getodk.org/central-install-digital-ocean/#using-a-custom-database-server) section during install.
    image

- Referenced the extension install in the [Using S3-compatible storage](https://docs.getodk.org/central-install-digital-ocean/#using-s3-compatible-storage) section, to help users debug if they encounter an error setting this up (linking to the 'Using a custom database' section).

    image


lognaturel (approved):
Thanks!

PR #1884, opened 2024-11-15
Clarify loss of Location EXIF Data when choosing images

### LINKED ISSUE #1851: Loss of Location EXIF Data When Using Android 13 Photo picker
With the introduction of Android 13, Google rolled out a new photo picker that is backported to devices running Android 11 and 12. This new picker is automatically used whenever an intent with Intent.ACTION_GET_CONTENT is called (when we attach images from a device). The new picker enhances privacy by removing sensitive metadata, including GPS location from photos, which affects apps relying on this data.
This behavior may change in the future, as many developers have reported issues and are requesting a way to retain location data. For now, we should document how the current system works.
See more: https://github.com/getodk/collect/issues/6009

grzesiek2010:
closes #1851

This PR adds a warning to clarify that sensitive EXIF data may not be accessible on newer Android devices.

### DIFF
--- a/docs/form-question-types.rst
+++ b/docs/form-question-types.rst
@@ -1682,7 +1682,7 @@ Image widgets
   Image files can be very large. We recommend always including :ref:`a maximum image size in form design <scaling-down-images>`. Also, consider making test submissions to your server with the Internet conditions you expect when gathering data to make sure that you can send files of the size you expect.
 
 .. warning::
-  As of Android 13, Google introduced a new photo picker, backported to devices running Android 11 and 12. This new picker is automatically used when users click the ``Choose Image`` button. The picker enhances privacy by removing sensitive metadata, including GPS location, from selected images. If you rely on metadata like GPS location, please note that this data may no longer be available.
+  Sensitive metadata, such as GPS location, is typically no longer included in images selected via :guilabel:`Choose Image` on Android 11 and later.
 
 .. _default-image-widget:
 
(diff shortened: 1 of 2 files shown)


docs/form-question-types.rst
```diff
@@ -1681,6 +1681,9 @@ Image widgets
 .. tip::
   Image files can be very large. We recommend always including :ref:`a maximum image size in form design <scaling-down-images>`. Also, consider making test submissions to your server with the Internet conditions you expect when gathering data to make sure that you can send files of the size you expect.
 
+.. warning::
+  As of Android 13, Google introduced a new photo picker, backported to devices running Android 11 and 12. This new picker is automatically used when users click the ``Choose Image`` button. The picker enhances privacy by removing sensitive metadata, including GPS location, from selected images. If you rely on metadata like GPS location, please note that this data may no longer be available.
```
lognaturel, line 1685: This reads as a bit more technical than it needs to be! Does it only affect images that are picked? Are GPS coordinates always included in pictures that are taken directly from Collect?

How about "Sensitive metadata, including GPS location, is generally no longer included in images selected from :guilabel:`Choose Image` on Android 11+." Then you won't need the "backported" addition to spelling either.

grzesiek2010: > Does it only affect images that are picked?

Yes.

> Are GPS coordinates always included in pictures that are taken directly from Collect?

It depends on the camera app if it is supported then yes.

> How about "Sensitive metadata, including GPS location, is generally no longer included in images selected from :guilabel:Choose Image on Android 11+." Then you won't need the "backported" addition to spelling either.

Ok we can use such a simple message. I will update the pull request.

PR #1886, opened 2024-11-15
Update docs for select one from map lines and shapes

grzesiek2010:
closes #


lognaturel:
It looks like there's something wrong with how you have gitlfs configured. Do you have it installed (https://git-lfs.com/)?

When I do `git lfs ls-files | grep select` I see several files but not the ones committed here. I think this has been an ongoing issue: https://github.com/getodk/docs/pull/1862, https://github.com/getodk/docs/pull/1831 That second PR has instructions on how to fix them. Could you try that, please?

PR #1894, opened 2024-12-02
Fix LFS pointers

yanokwa:


lognaturel:
This really makes it look like there's something wrong with @grzesiek2010's fork, clone or git lfs. @grzesiek2010 can you please make sure you have the latest git and git lfs? What do you see when you do `git lfs fsck`?

PR #1904, opened 2024-12-16
Offline Entities for Central v2024.3.0

lognaturel:
Progress towards https://github.com/getodk/docs/issues/1855

I did some rearranging that may be controversial. I recommend reviewing commit by commit.


docs/central-entities.rst
```diff
@@ -57,11 +64,9 @@ What's available now:
```
ktuite, line 64: I'm reading these docs top down (after looking at the individual commits) and I feel like the upper list of "what's available now" could also use a bullet point or two about updating entities and offline entities. I'm not sure if "Use Entities in one or more follow-up form" covers it.

docs/central-entities.rst
```diff
@@ -300,39 +307,15 @@ Updating Entities from Forms
 
 You can use Forms to update Entity information. These Forms can be authored to, for example, update previous observations to new values or change the status of an Entity. Just like Entity creation through Forms, you can specify which properties on which Entity instances will be updated when the Form's Submissions are uploaded to Central. The data in Submissions uploaded by Entity-updating Forms are applied to the Entity data saved on the Central server. These updated Entity values are then distributed to data collection clients once they synchronize with Central.
 
-.. _central-entities-update-conflicts:
-
-Parallel Updates in Updating Forms
```
ktuite, line 305: I'm a fan of moving this down and not having it be so prominent in the updating entities section!

docs/central-entities.rst
```diff
@@ -457,15 +442,67 @@ As you type, Central will highlight any fields you have changed in yellow so you
...
+
+If an Entity is created and then updated offline, there is the possibility that the Form Submissions representing updates will be received by Central before the Submission representing the creation of the Entity. In that case, the update Submissions will be held for up to 5 days. If the create Submission is received any time before 5 days, the Entity will be created by Central and any held updates will immediately be processed in order. If the Submission that would have created the Entity is not received within 5 days, Central will process the first update Submission it has as a create and the Entity will be marked as having a conflict.
+
```
ktuite, line 501: Could be worth mentioning the reasoning behind this design decision (to hold submissions but then process them anyway) is about salvaging information and using the data in the Submissions even if previous Submissions are missing.

ktuite (approved):

lognaturel:
Thanks @ktuite!

PR #1912, opened 2025-01-13
backend: allow configuring S3_REGION

alxndrsn:
See: https://github.com/getodk/central/pull/870


lognaturel:
https://github.com/getodk/central/issues/846#issuecomment-2610714967

PR #1913, opened 2025-01-13
Add details about random seeds

lognaturel:
@latin-panda I've added some more detail to the random seed description that will hopefully be helpful. Let me know what you think and if you have ideas for more information that would be helpful for end users.


latin-panda (approved):
Thank your for sending this! It's clearer and provide good advice for values

latin-panda:
@lognaturel, I think you need to assign it to a reviewer with write access to unblock the merge.

yanokwa (approved):

PR #1927, opened 2025-03-11
Revise section about millions of Entities

lognaturel:
Add more details about performance with large Entity Lists.

### DIFF
--- a/docs/entities-intro.rst
+++ b/docs/entities-intro.rst
@@ -124,11 +124,21 @@ Can I have millions of Entities?
 
 There are two current limitations that make millions of Entities impractical: data transfer and form performance.
 
-Currently, all Entities that have not been deleted are sent to every device on every update. Depending on your data connection, this may be a limiting factor for your project. We will eventually add support for archiving Entities to address this limitation.
+Currently, all Entities that have not been deleted are sent to every device on every update. Depending on your data connection, this may be a limiting factor for your project. To address this, we will eventually add functionality such as archiving Entities, synchronizing only updated Entities, or assigning a subset of Entities to a user.
 
-Entities are currently represented in memory for access by forms. Modern devices can easily process multiple tens of thousands of entities in this way, but your form may become slow or crash if you have more than 50,000 Entities.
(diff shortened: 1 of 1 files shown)


docs/entities-intro.rst
```diff
@@ -124,11 +124,21 @@ Can I have millions of Entities?
...
+Currently, all Entities that have not been deleted are sent to every device on every update. Depending on your data connection, this may be a limiting factor for your project. To address this, we will eventually add functionality such as archiving Entities, synchronizing only updated Entities, or assigning a subset of Entities to a user.
 
-Entities are currently represented in memory for access by forms. Modern devices can easily process multiple tens of thousands of entities in this way, but your form may become slow or crash if you have more than 50,000 Entities.
+Starting in Collect v2024.3 and Central v2024.3, Entities are represented in a database. If you use simple expressions using ``=``, ``and`` and ``or`` in choice filters or :ref:`lookup expressions <referencing-values-in-datasets>`, they will be performed quickly. We verify that a form with 100,000 Entities with 6 properties that are each 36 random characters long is usable on a device from 2021 with 3 GB of RAM.
```
seadowg, line 129: Do we need to say "database"? It feels like too much detail. What about:
 
> Starting in Collect v2024.3 and Central v2024.3, Entities are are optimized so that larger numbers can be used in forms. There are limitations to this however.

docs/entities-intro.rst
```diff
@@ -124,11 +124,21 @@ Can I have millions of Entities?
...
+
+The number of properties that each Entity has and the size and uniqueness of values saved will also affect both data transfer and form performance.
+
+.. seealso::
+    :ref:`What form fields should I save to my Entities as properties? <entities-intro-form-fields>`
+
+If you have a form with may Entities that feels slow, we encourage you to post about it on `the forum <https://forum.getodk.org/c/support/6>`_ so that we can recommend approaches that will make it work faster or design improvements to the system.
+
+.. note::
+    We generally do not recommend using `pulldata <https://xlsform.org/en/#how-to-pull-data-from-csv>`_ and `search() <https://xlsform.org/en/#dynamic-selects-from-pre-loaded-data>`_ instead of ``instance`` and ``select_one_from_file`` because they do not work with offline Entities. Starting in Collect v2024.3 and Central v2024.3, there is limited performance benefit to ``pulldata`` and ``search()``.
```
seadowg, line 141: `pulldata` does work with offline entities, but `search` would use the original CSV (so only uses online ones).

seadowg (changes requested):

docs/entities-intro.rst
lognaturel: I had totally forgotten this piece of work! https://github.com/getodk/collect/pull/6451

PR #1956, opened 2025-05-28
Add details on autocomplete appearance

lognaturel:
In response to some questions from @latin-panda around Web Forms behavior for this appearance.


latin-panda (approved):
Thanks! That clarify expectations

PR #1963, opened 2025-06-16
Update docs for Central v2025.2

lognaturel:
@nicoleorlowski I was sure there'd be more to it than that! We could add details about search and how it will look for the provided text as a prefix in the various fields it searches but it's probably not too important.


nicoleorlowski (approved):
LGTM!

PR #1971, opened 2025-07-03
Finalized or sent submissions can be edited

lognaturel:
Closes https://github.com/getodk/docs/issues/1957

I considered adding some details about the user experience in form lists. Things like:
- every edit will appear separately with the edit count in parentheses
- you can't edit an earlier edit, only the latest one and Collect will prompt for that
- if there's an active draft that's an edit, you can only edit or discard that one, not start a new edit

I decided not to include them because I think there's enough information in Collect to explain these things. Let me know if you think otherwise.

I hesitated on where to put the form design docs. We don't have any details or examples about the settings sheet in the XLSForm page so I thought it worked best where it is even though it's the only piece of form design on that page.

### DIFF
--- a/docs/collect-forms.rst
+++ b/docs/collect-forms.rst
@@ -136,6 +136,8 @@ Unlike with Central-based edits, Collect-based edits use the version of the form
 
 Edits to submissions that create or update Entities are allowed but they will only modify the submission, not the corresponding Entity.
 
+If you manually send submissions from "Ready to send," edits must be sent in order. If you attempt to send a later edit before an earlier one, you will get an error.
+
 To allow editing of finalized or sent submissions on device, the form definition must be modified. Add a ``client_editable`` column to the ``settings`` sheet in your XLSForm and set its value to ``true``:
 
 .. rubric:: XLSForm --- enabling edits from Collect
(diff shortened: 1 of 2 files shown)


docs/collect-forms.rst
```diff
@@ -119,6 +119,35 @@ To show sent and unsent forms:
...
+
+Unlike edits to drafts, edits to finalized or sent submissions are tracked in Central and shown in the submission :ref:`activity feed <central-submissions-details>`. A data collector can make a series of edits to a single submission from Collect and each will result in a separate entry in the activity feed. Once an edit to a specific submission is made from Central, however, it is no longer possible to edit that one from Collect.
+
+Unlike with Central-based edits, Collect-based edits use the version of the form that was active when the form was originally filled, NOT the latest one. That means you can turn off edits to new submissions by changing the form definition but existing submissions will continue to be editable.
+
+Edits to submissions that create or update Entities are allowed but they will only modify the submission, not the corresponding Entity.
+
+If you manually send submissions from "Ready to send," edits must be sent in order. If you attempt to send a later edit before an earlier one, you will get an error.
```
yanokwa, line 139: Drop the comma after "Ready to send". Or maybe change it to 
>:guilabel:`Ready to send`

lognaturel:
Thanks! @grzesiek2010 @seadowg, would be great for one or both of you to read through and see if there's any detail I forgot. I can follow up adding those in and address @yanokwa's suggestion.

PR #1975, opened 2025-07-03
Match case

yanokwa:


lognaturel:
I suppose we only capitalize `Form` in the Central context, not when talking about Collect.

PR #1977, opened 2025-07-13
Add tip to disable improve location accuracy when using mock locations

### LINKED ISSUE #1976: Add tip to disable improve location accuracy when using allow-mock-accuracy
In https://forum.getodk.org/t/gps-location-switching-between-tablet-location-and-mock-location/55466/8, @grzesiek2010 said....
> Based on our testing, if you [disable "Improve location accuracy"](https://www.youtube.com/watch?v=6nj9JYw0ric) in the Android settings, the device will rely solely on GPS. In that case, if GPS is being spoofed, the location will consistently appear fake.

I think we should add this as a tip to the allow-mock-accuracy section in the docs. I'd put it where the selected-at() tip is currently and move that one before the XLSForm with optional parameters section.

grzesiek2010:
closes #1976 

Added a tip to clarify that when using mock locations, real location data may still be delivered if the device uses not only GPS, but also Wi-Fi or cellular networks to determine location.

### DIFF
--- a/docs/form-question-types.rst
+++ b/docs/form-question-types.rst
@@ -1495,6 +1495,14 @@ Captures the current geolocation from the device. The location is displayed in d
 
 This question type shows a dialog with the current accuracy and lets the data collector decide when to capture the point. For capturing location without data collector intervention, see :ref:`start-geopoint <metadata-start-geopoint>`. For a geopoint with a user-selected location, see :ref:`placement-map <placement-map-widget>`.
 
+.. tip::
+
+  You can use :func:`selected-at()` to require geopoints meet a particular threshold. For example, if you need points with an accuracy better than 10 meters, use this constraint:
+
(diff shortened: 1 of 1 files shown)


lognaturel (approved):

lognaturel (changes requested):
https://github.com/getodk/docs/pull/1977/files#diff-21f092e50d37df3a102b60ffb5796cf82096ab4fe257d54f14ed1bb3da6668bfL1524

This anchor (.. _accuracy-constraint: ) needs to go with the corresponding section. 

https://github.com/getodk/docs/pull/1977/files#diff-21f092e50d37df3a102b60ffb5796cf82096ab4fe257d54f14ed1bb3da6668bfL1511

This section needs to mention the issue this PR addressed somehow. Maybe like the other tip, there could be an anchor for this tip and a brief sentence here with a link to the tip?

PR #1997, opened 2025-10-15
Remove Kapa until frequent captcha challenges are resolved

seadowg:
As discussed on Slack, multiple core team members are seeing repeated captcha challenges when trying to use "Ask AI" or "Search" at [docs.getodk.org](http://docs.getodk.org). We've confirmed that this is the case in multiple countries, across multiple ISPs and when using Chrome rather than other less popular browsers.

I've removed `search.html` so the default search is provided again. I couldn't find a way of providing a fallback based on captchas as the Kapa widget does not seem to offer an event around that (https://docs.kapa.ai/integrations/website-widget/javascript-api/events). Discussing this with Kapa support is probably the only way to resolve this and make it an option again.

Potentially an issue to add Kapa back in based on discussion with their support?

### DIFF
--- a/docs/_templates/sidebar/search.html
+++ b/docs/_templates/sidebar/search.html
@@ -1,3 +1,0 @@
-<div class="sidebar-search-container">
-<button class="sidebar-search" id="custom-ask-ai-button">Ask AI</button>
-</div>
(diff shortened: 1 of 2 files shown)


seadowg:
@yanokwa agreed!

PR #2018, opened 2025-12-15
Upgrade to Central v2025.4

lognaturel:
Instructions for upgrading Central to v2025.4

I went back and forth on whether or not to recommend the reindexing. From [PostgreSQL 14.11 release notes](https://www.postgresql.org/docs/release/14.11/), pointed out by @alxndrsn:

> one bug was fixed that could have resulted in corruption of GIN indexes during concurrent updates. If you suspect such corruption, reindex affected indexes after installing this update.

That doesn't provide much information, presumably because it's hard/impossible to tell if an install was affected. We have GIN indexes and some servers do have high concurrency. So it's best for us to recommend the maintenance.

I tried the commands on test. The `audits` one took 24 seconds so I think it will be very fast for most users. I ordered the commands from fastest to most likely to take time.

### DIFF
--- a/docs/central-upgrade.rst
+++ b/docs/central-upgrade.rst
@@ -15,7 +15,7 @@ To upgrade, start by reviewing upgrade notes for all versions between your curre
 Upgrade notes
 -------------
 
-* Central v2025.4: no upgrade notes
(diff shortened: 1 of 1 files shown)


sadiqkhoja (approved):

matthew-white:
Could we write a database migration to do this instead of adding it to docs?

PR #2022, opened 2026-01-12
Improve DB upgrade language, add Sentry warning

yanokwa:
I decided to put the Sentry IP change in the Sentry section because it's not bound by the version...


lognaturel (changes requested):
Suggestions inline

PR #2028, opened 2026-02-05
Add documentation for multiple Entities per form

lognaturel:
The goal of this PR is to update all existing documentation to lift the limitation of one Entity being created or updated per form. I'll follow up once the `pyxform` work is complete with one or two tutorials focused on multiple Entities being created or updated per form.

### DIFF
--- a/docs/entities-intro.rst
+++ b/docs/entities-intro.rst
@@ -115,7 +115,7 @@ If you usually have Internet connectivity, this is unlikely to be very important
 I need to assign specific Entities to specific data collectors, how can I represent this?
 -----------------------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


docs/entities-intro.rst
```diff
@@ -206,14 +202,18 @@ How do I use forms to create or update Entities?
 
 Add an ``entities`` sheet to your form. This sheet is used to define how data from this form's submissions should be applied to Entity Lists.
 
-Currently, a single submission can only affect a single Entity in a fixed Entity List. To specify which list to create or update an Entity in, use the ``list_name`` column.
+To specify which list to create or update an Entity in, use the ``list_name`` column.
```
ktuite: For some reason, I felt like this should specify Entity List and not just list. It worked before because of the previous sentence.

```suggestion
To specify which Entity List to create or update an Entity in, use the ``list_name`` column.
```

PR #2029, opened 2026-02-10
Add intersects function and incremental=true 

lognaturel:
Closes https://github.com/getodk/docs/issues/2021

I think it would be nice to have a dedicated tutorial as well, we can add it to the list! The goal here is to document the function and update the screenshots. I hesitated on where to document `incremental=true`. I ended up only putting it with `intersects` so it wouldn't be duplicated between trace and shape. I don't think it's very likely to be used with other constraints expressions but I tried to explain clearly how it could be used in case there are cases I'm not thinking of.


alyblenkin (approved):
So nice to finally update these!!

PR #2046, opened 2026-04-03
Central v2026.1.0

lognaturel:
Closes https://github.com/getodk/docs/issues/2034 along with https://github.com/getodk/docs/pull/2038

I tried to add a little bit of info on how to unlink forms and clear properties. @ktuite please review and push any updates straight to the branch!

### DIFF
--- a/docs/central-customization.rst
+++ b/docs/central-customization.rst
@@ -1,0 +1,16 @@
+Customizing Central's interface
+===============================
+
+As a Central system administrator, you can customize the appearance of your Central login page to match your organization. This will help users recognize they're on the correct server and feel confident that they're signing into the right place. You can also use the login page to give short updates or inspirational messages to users of your ODK Central server.
+
+The customizations you make are immediately live for your users.
+
+.. image:: /img/central-customization/central-login-customization.jpg
+
+You can clear your customizations at any time to return to the ODK defaults.
+
+The images you select will be downloaded by your users as-is. We recommend sizing them down as much as you can while maintaining high quality.
+
+For the logo, we recommend an image that has a long edge of 500 - 800 pixels. Aim for a maximum download size of 500 KB.
+
+The login image is displayed as large as your users' screens. We recommend aiming for a maximum of 2 MB so the page doesn't take too long to download.
--- a/docs/central-entities.rst
+++ b/docs/central-entities.rst
@@ -387,6 +387,20 @@ To browse all Entity Lists in a Project, go to the :guilabel:`Entity Lists` tab.
 
 .. image:: /img/central-entities/entity-lists.png
 
+.. _central-entities-settings:
+
+Entity List Settings
+~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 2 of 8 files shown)


docs/central-customization.rst
```diff
@@ -0,0 +1,16 @@
...
+
+The customizations you make are immediately live for your users.
+
+.. image:: /img/central-customization/central-login-customization.jpg
+
+You can clear your customizations at any time to return to the ODK defaults.
+
+The images you select will be downloaded by your users as-is. We recommend sizing them down as much as you can while maintaining high quality.
+
+For the logo, we recommend an image that has a long edge of 500 - 800 pixels. Aim for a maximum download size of 500 KB.
```
lognaturel, line 14: @matthew-white I'm surprised the image is capped at 130px in both dimensions. Did you consider just capping it vertically? A lot of logos are wide.

I did confirm that indeed as-expressed 130 is dp so users would ideally use images that are 520px on the long edge to get true 4x sharpness on high-density displays.

matthew-white: @nicoleorlowski and I landed on 130x130 during a quick design pairing session, but we could definitely adjust those numbers. Feel free to file an issue about that, it'd be easy to make that change.

lognaturel: https://github.com/getodk/central/issues/1761

docs/central-entities.rst
```diff
@@ -380,12 +398,50 @@ See a sample form `here <https://docs.google.com/spreadsheets/d/1R3Ja6hOjjntE42H
...
+.. _deleting-entity-lists:
+
+Deleting Entity Lists
+~~~~~~~~~~~~~~~~~~~~~~
+
+You can delete an Entity List that is no longer needed if it is not linked to any Forms. If the list is currently in use, unlink it from all Forms before proceeding.
+
+To unlink a Form from an Entity List, you must :ref:`edit the Form's definition <central-forms-updates>`.
+
+If a form uses Entities from the list you want to delete, you can remove the question(s) that reference the list or replace the list with a CSV. If a form creates or updates Entities in the list you want to delete, you must delete the row on the ``entities`` sheet of your form definition that references that Entity List.
```
ktuite, line 432: I would like to improve this section in the future! 

What I wrote in the API docs:
A Dataset cannot be deleted if any Forms are still linked to it, either as a source (a Form that creates/updates Entities in it) or as a consumer (a Form that uses it as an attachment). All such Forms must be updated first:
- To update a Form that writes to a Dataset, remove that dataset from the `entities` tab of the XLSForm and remove associated `save_to`s.
- To update a Form that reads from a Dataset, replace the attachment with a static CSV or update the Form to not use that attachment.
A Form can also be deleted as a way to unlink it from a Dataset, but then that Form cannot be restored.

docs/central-entities.rst
ktuite, line 497: Not seeing this image in the local preview?

ktuite (approved):
let's get it in there!

docs/central-entities.rst
lognaturel: Interesting! Looks like it's at https://docs.getodk.org/central-entities/#editing-entity-data and it was an existing image so maybe just a temporary quirk.

PR #2049, opened 2026-04-08
Add household survey tutorial

lognaturel:
Starts the household survey tutorial illustrating multiple Entity declarations in a single form definition.

[Rendered](https://output.circle-artifacts.com/output/job/226201db-e062-4748-a9a7-ed37f2520477/artifacts/0/docs/_build/dirhtml/tutorial-household-survey/index.html)

The big thing I need to do is write a narrative around the form itself. I'll also include a short follow-up form. 

Some feedback that might be helpful at this point:
* Does the form strike the right balance between feeling realistic and not having too much detail?
* Are the diagrams helpful?
* Is the level of explanation ok?

### DIFF
--- a/docs/tutorial-household-survey.rst
+++ b/docs/tutorial-household-survey.rst
@@ -1,0 +1,130 @@
+Entities tutorial: household survey
+===================================
+
+Household-based surveys are one of the most common patterns in field data collection, spanning population studies, public health, agriculture, education, and humanitarian response. Across these domains, the same core structure is found: there's a household or other parent unit, the people or items linked to it, and the operational forms and supporting data needed to carry out the work.
+
+This tutorial shows how to leverage Entities to support a household survey at scale.
+
+In this workflow, households have already been identified, sampled, and assigned to enumerators. We focus on enumerators visiting each of their assigned households to capture baseline data and build member rosters. They will then need to follow up both with the households and selected individuals over time:
+
+.. mermaid::
+    :align: center
+
+    flowchart LR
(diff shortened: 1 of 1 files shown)


alyblenkin:
Overall great! I'm still working on following every step. Need to figure out the linking! 

### Intro
> This tutorial shows how to leverage Entities to support a household survey at scale.

This is helpful because it makes clear to the user why they would want to use entities instead of what they have currently. Maybe we also include scenarios of when it might be a good fit. 

> In this workflow, households have already been identified, sampled, and assigned to enumerators. We focus on enumerators visiting each of their assigned households to capture baseline data and build member rosters. They will then need to follow up both with the households and selected individuals over time

Diagram is great! We could make it aligned with our existing style? 

Could we also make the objectives easier to scan? Maybe something like this...

What You Will Learn
By the end of this tutorial, you will be able to:

- Link Entity lists : Link individual members and operational forms to a parent household
- Handle pre-assigned workloads: Use already sampled and assigned households to specific enumerators
- Register households: Capture baseline data and register the household as an Entity.
- Enable follow-up workflows: Access and update these records during subsequent visits

### Designing the Entity model

This section is very helpful! It would be great to get Andrew's thoughts on this too. I'm pretty sure he wrote a form post about it. I don't think we need a checklist but maybe we could pull out key things to consider in a bulleted list?

> while Entities store the current state of real-world things that need to be accessed and updated over time.

I like this distinction!

### Ending - Building the baseline form
The end kind of feels abrupt. Maybe we can say the intended outcome again? Do we recommend them testing it out with real people or what would be the best next step?

PR #2059, opened 2026-05-12
Getting started improvements

yanokwa:
- Split out Cloud and self-hosting to make it easier for folks to know there is a choice
- Added links to tutorial and template because those are more directly useful than the all widgets form.
- Reframed the docs as being for a project manager (e.g., explaining that QR codes need to be distributed to data collectors)
- Offered the two ways of collecting data (Collect, Web Forms) and put it in tabs for glanceability
- Added a shoutout to Ask AI because it's helpful!


lognaturel (approved):
Add link to forum in the help section

PR #2074, opened 2026-07-03
Add v2026.2 upgrade instructions

lognaturel:
I don't think there's anything that a typical user would need to look into, just two special notices. I tried to make it clear that most people are not affected.

### DIFF
--- a/docs/central-upgrade.rst
+++ b/docs/central-upgrade.rst
@@ -15,7 +15,7 @@ To upgrade, start by reviewing upgrade notes for all versions between your curre
 Upgrade notes
 -------------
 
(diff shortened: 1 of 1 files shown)


docs/central-upgrade.rst
```diff
@@ -15,6 +15,7 @@ To upgrade, start by reviewing upgrade notes for all versions between your curre
 Upgrade notes
 -------------
 
+* :ref:`Central v2026.2 <central-upgrade-2026.1>`: special instructions if you use direct backups via API or build frontend from source
```
sadiqkhoja:
```suggestion
* :ref:`Central v2026.2 <central-upgrade-2026.2>`: special instructions if you use direct backups via API or build frontend from source
```

sadiqkhoja (approved):

PR #2076, opened 2026-07-03
Add more v2026.2 docs

lognaturel:
A few things from https://github.com/getodk/docs/issues/2054


sadiqkhoja (approved):

sadiqkhoja:
should we update install instruction as well to not initialize/update `client` submodule?

lognaturel:
I added a commit, is that what you had in mind, @sadiqkhoja, or is there more?

sadiqkhoja:
yes. And maybe just `git submodule update -i server`

lognaturel:
I figured we should try to keep everything as familiar as possible for now but you're right we could simplify some of this!

PR #2081, opened 2026-07-28
feat(#2080): add section for default parameters

### LINKED ISSUE #2080: Document how to prefill forms fields using query parameters
This has been a feature in enketo for a while (and soon to be supported in web-forms) but I can't find any official documentation about how to do it. Unofficial documentation exists on the forum...

garethbowen:
closes: #2080 

A section about using default parameters with public link URLs. I'm not convinced this is exactly the right place for this to live but it is the best option I could find right now.

None.

Nothing - but the web-forms feature should be merged and released before the documentation is pushed.

None.

### DIFF
--- a/docs/central-submissions.rst
+++ b/docs/central-submissions.rst
@@ -75,10 +75,12 @@ Once a Link is created, it will appear in the table, along with a web address yo
 
 You cannot yet edit any of the details of a Public Link.
 
-.. _central-submissions-link-default-parameters:
+.. _central-submissions-link-field-values-url:
 
-Default Parameters
-^^^^^^^^^^^^^^^^^^
+Set field values in the URL
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Speed up data entry by including known field values as parameters in the form's URL. ODK will automatically fill matching values when the form opens.
 
 Parameters use the format :code:`d[{field}]={value}`, where ``field`` identifies the form field you want to set and ``value`` is the text to put in it. For example, :code:`d[name]=odk` sets the ``name`` field to ``odk``. For a field inside groups or repeats, include the names of the groups or repeats in the field identifier, separated by slashes. For example: :code:`d[person/name]=Maria`.
 
--- a/docs/central-submissions.rst
+++ b/docs/central-submissions.rst
@@ -80,7 +80,7 @@ You cannot yet edit any of the details of a Public Link.
 Default Parameters
 ^^^^^^^^^^^^^^^^^^
 
-You can add parameters to the public link URL to set values when the form is opened. This can be useful to prefill the form with data that's available on the page where the link is clicked. Parameters must be in the format :code:`d[{xpath}]={value}`, where xpath is the ref of the element you want to set, and value is the string value. For example, :code:`d[/root/name]=odk` will set the string "odk" into the name element. You can set values on multiple elements by including more parameters.
+You can add parameters to the public link URL to set values when the form is opened. This can be useful to fill the form with data that's available on the page where the link is clicked. Parameters must be in the format :code:`d[{xpath}]={value}`, where "xpath" is the ref of the element you want to set, and value is the string value. For example, :code:`d[/root/name]=odk` will set the string "odk" into the name element. You can set values on multiple elements by including more parameters.
 
 .. _central-submissions-link-revoke:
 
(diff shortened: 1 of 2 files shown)


docs/central-submissions.rst
```diff
@@ -75,6 +75,13 @@ Once a Link is created, it will appear in the table, along with a web address yo
 
 You cannot yet edit any of the details of a Public Link.
 
+.. _central-submissions-link-default-parameters:
+
+Default Parameters
+^^^^^^^^^^^^^^^^^^
+
+You can add parameters to the public link URL to set values when the form is opened. This can be useful to fill the form with data that's available on the page where the link is clicked. Parameters must be in the format :code:`d[{XPath}]={value}`, where XPath is the ref of the element you want to set, and value is the string value. For example, :code:`d[/root/name]=odk` will set the string "odk" into the name element. You can set values on multiple elements by including more parameters.
```
garethbowen:
```suggestion
You can add parameters to the public link URL to set values when the form is opened. This can be useful to fill the form with data that's available on the page where the link is clicked. Parameters must be in the format :code:`d[{XPath}]={value}`, where XPath is the ref of the element you want to set, and value is the string value. For example, :code:`d[/root/name]=odk` will set the string "odk" into the name element. To reduce the length of the URL omit the primary instance name and leading slash, for example:  :code:`d[name]=odk`. Set values on multiple elements by including more parameters separated by an ampersand, :code:`d[name]=odk&d[engine]=webforms`.

Note that the parameters must be URL encoded to be parsed correctly.
```

docs/central-submissions.rst
```diff
@@ -75,6 +75,15 @@ Once a Link is created, it will appear in the table, along with a web address yo
 
 You cannot yet edit any of the details of a Public Link.
 
+.. _central-submissions-link-default-parameters:
+
+Default Parameters
+^^^^^^^^^^^^^^^^^^
+
+You can add parameters to the public link URL to set values when the form is opened. This can be useful to fill the form with data that's available on the page where the link is clicked. Parameters must be in the format :code:`d[{XPath}]={value}`, where XPath is the ref of the element you want to set, and value is the string value. For example, :code:`d[/root/name]=odk` will set the string "odk" into the name element. To reduce the length of the URL omit the primary instance name and leading slash, for example:  :code:`d[name]=odk`. Set values on multiple elements by including more parameters separated by an ampersand, :code:`d[name]=odk&d[engine]=webforms`.
```
lognaturel: We don't typically use "XPath", "ref", "primary instance" in user-facing docs.

The default root node name out of XLSForm is `data`, not `root`. I don't think we need to document that the root node name can be included, it's too much to explain for most users.

Here's an alternative proposal:

```
Parameters use the format :code:`d[{field}]={value}`, where ``field`` identifies the form field you want to set and ``value`` is the text to put in it. For example, :code:`d[name]=odk` sets the ``name`` field to ``odk``. For a field inside groups or repeats, include the names of the groups or repeats in the field identifier, separated by slashes. For example: :code:`d[person/name]=Maria`.

To set multiple fields, separate parameters with an ampersand (``&``), for example: :code:`d[name]=odk&d[engine]=webforms`.
```

lognaturel (changes requested):
Suggestion inline!

docs/central-submissions.rst
garethbowen: Done.

lognaturel (approved):
Thank you! Can't wait to release this!

docs/central-submissions.rst
```diff
@@ -75,6 +75,17 @@ Once a Link is created, it will appear in the table, along with a web address yo
 
 You cannot yet edit any of the details of a Public Link.
 
+.. _central-submissions-link-default-parameters:
+
+Default Parameters
```
yanokwa: I think "URL defaults" is more human?

garethbowen: I struggled with the naming here - I wanted something that people would understand but would also be a searchable feature. I'm not a fan of "url defaults" because it's ambiguous (is the URL defaulting to something?), but also would be hard to google for. I don't really like Default Parameters either. We could go for something verbose like "setting default values via the URL"? 

Suggestions welcome...

yanokwa: ChatGPT likes "URL pre-fill"

yanokwa: Also liking "Set field values in the URL"

Then the user-facing benefit becomes
> Speed up data entry by including known field values in the form's URL. ODK will automatically fill the matching fields when the form opens, so users can focus on entering what’s new.

garethbowen: @yanokwa I've updated the title and added an intro paragraph borrowing from your user-facing benefit.

PR #2088, opened 2026-08-14
Add warning about S3 outages affecting form downloads

### LINKED ISSUE #2055: Add a warning to S3 storage docs about how users experience S3 down time
An insider reported that..
> I am self hosting and had a problem with the S3 storage provider - it went offline on Sunday 12th April and I was in XXXXXXX running a training course on how to use ODK, without a server that could download forms to new devices. So, as a backup strategy, I signed up for a month's subscription and intended to upload the forms to the ODK Cloud server and use them that way (without entity access of course!).
>
> Fortunately, we managed to find enough devices that already had the form definition downloaded so were able to work using the self-hosted server on the first day of training. Coming back from a day out in the field on Thursday 16th I found that the S3 storage was back online and the Central server was working as expected. I have just got back to the office and I'm trying to untangle all the issues that arose while I was away, including an enquiry to the S3 provider (XXXXXX) asking for an explanation of 4 days downtime on their platform.
>
> So that experience highlights a weak spot for me... S3 storage is great, except when it isn't. In the scenario where a Central Server uses S3 storage, and the S3 link is down, although the server still accepts submissions, it appears that form definitions cannot be downloaded (where they have any stored attachments) nor can submissions that have attachments that have already been transferred to S3 (more obviously!). I guess it is a very rare occurrence, but I just got hit by it. There's not a simple solution here as far as I can see, except perhaps to be able to reconfigure the Central Server on demand to use a different S3 bucket (and to have a spare bucket with another provider that has not gone down, with a copy of the original S3 bucket). Another (complicated) option might be to hold all form definition files on the Central Server so that it operates independently of S3 for distributing data to devices (or loading forms in Enketo) - it already seems to be robust enough to cope with S3 going down. The good news is that I haven't lost any data (which is sort of the point of S3 I think, and I double checked my Central backups were running hard!)

It'd be nice if we warned people in the docs about what happens if S3 is down.

gaurav-goswamii:

Adds a warning to the S3 documentation explaining that an S3 outage can prevent devices from downloading new forms and attachments, even though Central can continue accepting submissions.

Fixes #2055

### DIFF
--- a/docs/central-install-digital-ocean.rst
+++ b/docs/central-install-digital-ocean.rst
@@ -534,8 +534,8 @@ Using S3-compatible Storage
 
 By default, Central stores form and submission attachments in its main database, but it can be configured to move these to an external object store. 
 
-**Important**: If your S3-compatible storage is unavailable, Central can continue accepting submissions, but devices may not be able to download new forms or attachments stored there.
-
+- If S3-compatible storage goes down, Central will still accept submissions, but devices may not receive updates to forms with attachments, and integrations or dashboards that use attachments may fail.
+- If you opt into S3-compatible storage, you must design a backup and restore strategy for that storage.
 If you already have or plan to collect many files, storing them outside the main database can reduce database load and cost. It can also make it more practical to backup and restore the database.
 
 Consider the following to help you decide whether S3-compatible storage is a good fit:
--- a/docs/central-install-digital-ocean.rst
+++ b/docs/central-install-digital-ocean.rst
@@ -532,7 +532,11 @@ Creating and using a custom PostgreSQL database server
 Using S3-compatible Storage
 ---------------------------
 
-By default, Central stores form and submission attachments in its main database, but it can be configured to move these to an external object store. If you already have or plan to collect many files, storing them outside the main database can reduce database load and cost. It can also make it more practical to backup and restore the database.
+By default, Central stores form and submission attachments in its main database, but it can be configured to move these to an external object store. 
+
+**Important**: If your S3-compatible storage is unavailable, Central can continue accepting submissions, but devices may not be able to download new forms or attachments stored there.
+
+If you already have or plan to collect many files, storing them outside the main database can reduce database load and cost. It can also make it more practical to backup and restore the database.
 
 Consider the following to help you decide whether S3-compatible storage is a good fit:
 
--- a/docs/entities-intro.rst
+++ b/docs/entities-intro.rst
@@ -27,7 +27,7 @@ Concepts
 What are Entities?
 ------------------
 
-In the ODK context, an "Entity" can be thought of as a "thing". If your project involves things that need to be shared between forms and may change over time, you can represent them as Entities. You can use Entities to represent real things like trees, people, or cities. You can also represent more abstract things like tree visits, malaria cases, or city ratings as Entities.
+In ODK, an **Entity** can be thought of as a “thing.” If your project involves items that need to be shared across forms and may change over time, you can represent them as Entities. Entities can represent real-world things such as trees, people, or cities. They can also represent more abstract concepts, such as tree visits, malaria cases, or city ratings.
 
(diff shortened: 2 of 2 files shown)


yanokwa (changes requested):
Thanks for the PR. I think it would read better if slotted in as the second to last bullet.

* If S3-compatible storage goes down, Central will still accept submissions, but devices may not receive updates to forms with attachments, and integrations or dashboards that use attachments may fail.
* If you opt into S3-compatible storage, you must design a backup and restore strategy for that storage.

gaurav-goswamii:
@yanokwa Thanks! I've moved the warning to the second-to-last bullet and updated the wording as suggested.

docs/central-install-digital-ocean.rst
```diff
@@ -532,7 +532,11 @@ Creating and using a custom PostgreSQL database server
 Using S3-compatible Storage
 ---------------------------
 
-By default, Central stores form and submission attachments in its main database, but it can be configured to move these to an external object store. If you already have or plan to collect many files, storing them outside the main database can reduce database load and cost. It can also make it more practical to backup and restore the database.
+By default, Central stores form and submission attachments in its main database, but it can be configured to move these to an external object store. 
+
+- If S3-compatible storage goes down, Central will still accept submissions, but devices may not receive updates to forms with attachments, and integrations or dashboards that use attachments may fail.
+- If you opt into S3-compatible storage, you must design a backup and restore strategy for that storage.
+If you already have or plan to collect many files, storing them outside the main database can reduce database load and cost. It can also make it more practical to backup and restore the database.
```
lognaturel, line 539: This explains why someone might want to opt into the functionality, it feels like it should be at the top as it was before.

docs/central-install-digital-ocean.rst
```diff
@@ -532,7 +532,11 @@ Creating and using a custom PostgreSQL database server
 Using S3-compatible Storage
 ---------------------------
 
-By default, Central stores form and submission attachments in its main database, but it can be configured to move these to an external object store. If you already have or plan to collect many files, storing them outside the main database can reduce database load and cost. It can also make it more practical to backup and restore the database.
+By default, Central stores form and submission attachments in its main database, but it can be configured to move these to an external object store. 
+
+- If S3-compatible storage goes down, Central will still accept submissions, but devices may not receive updates to forms with attachments, and integrations or dashboards that use attachments may fail.
+- If you opt into S3-compatible storage, you must design a backup and restore strategy for that storage.
```
lognaturel, line 538: This is already in the consideration list below. I think it works well where it is.

docs/central-install-digital-ocean.rst
```diff
@@ -532,7 +532,11 @@ Creating and using a custom PostgreSQL database server
 Using S3-compatible Storage
 ---------------------------
 
-By default, Central stores form and submission attachments in its main database, but it can be configured to move these to an external object store. If you already have or plan to collect many files, storing them outside the main database can reduce database load and cost. It can also make it more practical to backup and restore the database.
+By default, Central stores form and submission attachments in its main database, but it can be configured to move these to an external object store. 
+
+- If S3-compatible storage goes down, Central will still accept submissions, but devices may not receive updates to forms with attachments, and integrations or dashboards that use attachments may fail.
```
lognaturel, line 537: I would probably tend to add this bullet point to the consideration list below. Another alternative would be to style it here as a warning.

lognaturel (changes requested):

docs/central-install-digital-ocean.rst
gaurav-goswamii: i removed the duplicate.

docs/central-install-digital-ocean.rst
gaurav-goswamii: Added the S3 outage bullet into the existing bullet list. thanks

docs/central-install-digital-ocean.rst
gaurav-goswamii: The introductory paragraph is fully restored right under the header.

PR #2092, opened 2026-08-22
Add more clarity

yanokwa:
I think we should bias to fully-spelled out flags, but it adds clarity. I think this is especially important for commands on the common path.

Related: #2091


lognaturel:
I disagree! I think expanding flags is generally useful but `-d` in this context is so common and so baked into people's fingers that I think it's just more confusing to see the less common `--detach`. I wouldn't make this change but don't feel terribly strongly if you really want to.

PR #2095, opened 2026-08-27
Clarify rules about custom property names

matthew-white:
Creating a PR based on https://github.com/getodk/central/issues/1875#issuecomment-5429805232

The rules in the user docs about custom property names are stricter than what Central enforces. In the issue linked to above, there was confusion because a leading underscore actually is allowed. For property names, two leading underscores are disallowed, but one is accepted.

I'll leave a few line comments as well with thoughts about these rules and whether they could be further clarified.

### DIFF
--- a/docs/central-projects.rst
+++ b/docs/central-projects.rst
@@ -131,7 +131,7 @@ Managing Custom Properties
 
 Custom Properties let you control which Entities each App User or Public Link receives. For example, if your organization works across multiple regions, you can create a ``region`` Custom Property, assign each App User a region, and configure an :ref:`Entity List access filter <entity-list-access-filter>` so each user receives only Entities from their region.
 
-Go to the :guilabel:`Custom Properties` tab for a project to see the available Custom Properties and add new ones. Custom Property names follow the same rules as form field names or Entity property names: they can't have spaces in them, must start with a letter, and can only contain letters, numbers, _ or -.
+Go to the :guilabel:`Custom Properties` tab for a project to see the available Custom Properties and add new ones. Custom Property names follow the same rules as form field names or Entity property names: they can't have spaces in them, must start with a letter or underscore, and can only contain letters, numbers, and a limited set of symbols (``_``, ``-``, and ``.``).
 
 Once a Custom Property is defined, assign values to App Users or Public Links. You can then create :ref:`Entity List access filters <entity-list-access-filter>` that compare Entity properties with those Custom Properties to determine which Entities each user receives.
 


docs/central-projects.rst
```diff
@@ -131,7 +131,7 @@ Managing Custom Properties
 
 Custom Properties let you control which Entities each App User or Public Link receives. For example, if your organization works across multiple regions, you can create a ``region`` Custom Property, assign each App User a region, and configure an :ref:`Entity List access filter <entity-list-access-filter>` so each user receives only Entities from their region.
 
-Go to the :guilabel:`Custom Properties` tab for a project to see the available Custom Properties and add new ones. Custom Property names follow the same rules as form field names or Entity property names: they can't have spaces in them, must start with a letter, and can only contain letters, numbers, _ or -.
+Go to the :guilabel:`Custom Properties` tab for a project to see the available Custom Properties and add new ones. Custom Property names follow the same rules as form field names or Entity property names: they can't have spaces in them, must start with a letter or underscore, and can only contain letters, numbers, and a limited set of symbols (``_``, ``-``, and ``.``).
```
matthew-white, line 134: Here, "letter" isn't restricted to English A-Z. It includes letters with diacritical marks, Chinese characters, etc. So users may still think that the validation is stricter than it actually is.

matthew-white, line 134: Should we mention that two leading underscores aren't allowed? How about reserved property names — `label`, `name`, `displayName`?

matthew-white, line 134: Code references:

- [`validatePropertyName()`](https://github.com/getodk/central-backend/blob/5d88c1ebfe707b013e0a64da3f353e3576824af9/lib/data/dataset.js#L83-L96)
- [`validateActorPropertyName()`](https://github.com/getodk/central-backend/blob/5d88c1ebfe707b013e0a64da3f353e3576824af9/lib/data/actor-properties.js#L4-L8)

matthew-white (commented):
Questions and comments about these user docs

ktuite:
I'm wondering where these rules came from in the first place, because they are stricter than the dataset property names, but I didn't happen to implement them in the backend. Should Central be doing more to check properties like this?

docs/central-projects.rst
lognaturel: I think that's ok. It's another one of those things where accented characters are allowed for overall consistency across identifier names but I would absolutely not recommend them.

lognaturel: What is `displayName`? 😬

I don't feel strongly about it. I think it's unlikely enough that folks will come up with these names themselves that the errors from Central can stand on their own for those who do.

lognaturel (approved):

matthew-white:
@lognaturel, I had been thinking after the meeting last week that I should close this PR. I just hadn't gotten around to it yet. Or is that wrong and I should go ahead and merge? I'm happy for us not to be more explicit than the current docs if that helps users avoid awkward cases. Also happy to merge though.

docs/central-projects.rst
matthew-white: `displayName` does seem pretty unlikely. Same for two leading underscores hopefully. `name` doesn't seem terribly unlikely to me though.

>  the errors from Central can stand on their own

One thing on my mind is that the errors in Central aren't super specific. E.g., Backend just says, "This is not a valid property name." It doesn't say, "`name` is reserved as a system property." I'm sure the user will figure it out, but I don't think it's necessarily obvious why a name like `name` would be a problem.

matthew-white:
> Should Central be doing more to check properties like this?

We discussed in the team meeting last Thursday that the user docs intentionally document stricter rules than what's technically allowed, so that we don't lead users to choosing names that would be inconvenient in certain settings. I'm not seeing a case where Backend needs to be doing more validation than it currently is.

matthew-white:
Just to link issues, I've also filed getodk/central#2182 about making it clear in Central Frontend what makes a property name valid or not. Probably we'll want the rules/explanation in Central to match what we document here. Or maybe we link from Central to the user docs.

lognaturel:
@matthew-white Let's merge for now. We have to revisit this because the Custom Properties page is going away and at that point I can try some wording around reserved keywords to see whether it's worth including.

PR #2099, opened 2026-09-03
replaces more docs that reference enketo to web forms

garethbowen:
closes #

addresses #

Various replacements of enketo -> wf


lognaturel (approved):
Thanks! This can be merged now.

PR #2105, opened 2026-09-11
chore: clarify last-saved feature for encrypted forms

latin-panda:

Last-saved values are not encrypted; adds that info in the docs. 

none

none

none


lognaturel (approved):
Thanks!

PR #2110, opened 2026-09-18
chore: add note on calculate and setValue with upload and annotate questions

latin-panda:
A note on calculate and setValue with upload and annotate questions

none

none

none


latin-panda:
@lognaturel Another small update to the docs. What do you think?

latin-panda:
@lognaturel I've added the additional note

lognaturel (approved):

